PR #157, opened 2025-02-20
fix: restore correct ssn validation

serikjensen:
While centralizing the SSN regex for assign signatory, there was actually a mismatch between how ssn was being validated in profile and signatory. In profile, we were stripping all non digits and then testing against the regex. In signatory we were using a regex to validate the full string.

This caused regressions in the profile form since the ssn regex was now expecting non digit characters.

This PR updates to make SSN validation more consistent
* Updates the SSN regex to assume non digits consistently between profile and signatory
* Moves the strip non digits function used for phone numbers to the format string utility file since it has uses beyond phone
* Updates to use that function in both ssn inputs to remove non digit input prior to validation
* Updates the SSN regex to just validate the digits

Note: 

### Profile Before

https://github.com/user-attachments/assets/da8de004-e952-4c98-b8ef-9dd83020b7ea

### Profile After

https://github.com/user-attachments/assets/a3c777e8-c7dd-4104-8aaf-c1ed4995e326

### No regressions in signatory

https://github.com/user-attachments/assets/c295b544-f650-4596-9db8-59d465a6c35b


jeffredodd (approved):
🚢

dmortal (approved):

PR #159, opened 2025-02-20
feat: add company signature form

serikjensen:
This adds a company signature form as seen in the figma here https://www.figma.com/design/6dxOSiONDiJoa9zY1wOwbs/Frontend-SDK-Partner-Design-File?node-id=5220-17986&m=dev

### Signature form submitting correctly and firing an event with the sign form response

https://github.com/user-attachments/assets/03365ede-764a-4026-9422-b1ca43f72e6a

### Form updated to complete in the document list

Screenshot 2025-02-20 at 4 14 38 PM

### Form validation behaving correctly

https://github.com/user-attachments/assets/1008e9e3-55da-4609-83cd-506ad9db609c

### Download functionality behaving as expected

https://github.com/user-attachments/assets/5b9b557b-a9cb-40e3-8540-03bf9cad29da

### Event firing for back navigation

https://github.com/user-attachments/assets/a5adf74e-2ad3-46c3-9d4b-03a2017cd4ae

### Mobile view

Screenshot 2025-02-20 at 4 20 33 PM


src/components/Company/DocumentSigner/DocumentList/Head.tsx
```diff
@@ -1,7 +1,7 @@
 import { useTranslation } from 'react-i18next'
 
 export function Head() {
-  const { t } = useTranslation('Company.DocumentSigner')
+  const { t } = useTranslation('Company.DocumentList')
```
serikjensen, line 4: No idea what happened with these but there must have been some regression 😅 it has always been named DocumentList. This updates to fix

src/components/Company/DocumentSigner/SignatureForm/SignatureForm.tsx
```diff
@@ -0,0 +1,122 @@
...
+    await baseSubmitHandler(data, async payload => {
+      const signFormResponse = await signForm({
+        form_id: form.uuid,
+        body: {
+          signature_text: payload.signature,
+          agree: payload.confirmSignature.length > 0,
+          /**
+           * TODO[GWS-3365]: This is a temporary value here until we resolve the correct way forward with the IP address
+           */
+          signed_by_ip_address: '127.0.0.1',
```
serikjensen, line 78: This is something we still need to revisit. i think the consensus was that we just need to have them add the ip address via their proxy. Updated the ticket to generalize from employee document signing to company document signing

jeffredodd: This is probably not the solution but looks like there are a lot of tools out there that let you hit their API and they return your IP address. Maybe Gusto needs an endpoint like this? https://api.myip.com/

scally (approved):

PR #160, opened 2025-02-21
fix: restore validation for self onboarding profile form

serikjensen:
We were running into an issue with Profile form where validation was failing because self_onboarding was undefined which was causing validation to fail.

This removes the self_onboarding property from home address validation so that it is properly ignored when we are onboarding as self and the checkbox for self_onboarding is not present.

### Before: submit handler failing with self_onboarding undefined
profile-form-bug

### After: submit handler succeeding for self onboarding

https://github.com/user-attachments/assets/76f50b5a-53fe-435f-b814-dcd4590d09b9

### Address validation for admin only fires when self onboarding is not selected
When it is selected, validation does not fire and form submits correctly

https://github.com/user-attachments/assets/48727c07-b130-4ed8-a5de-336ff4e10649

### DIFF
--- a/src/components/Employee/Profile/HomeAddress.tsx
+++ b/src/components/Employee/Profile/HomeAddress.tsx
@@ -1,5 +1,5 @@
 import { Alert, Checkbox, Grid, Select, TextField } from '@/components/Common'
-import { EmployeeOnboardingStatus, STATES_ABBR } from '@/shared/constants'
+import { STATES_ABBR } from '@/shared/constants'
 import { Link, ListBoxItem } from 'react-aria-components'
 import { useFormContext } from 'react-hook-form'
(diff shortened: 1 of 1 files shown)


src/components/Employee/Profile/Profile.tsx
```diff
@@ -242,40 +242,37 @@ const Root = ({ isAdmin = false, ...props }: ProfileProps) => {
       }
       //create or update home address - only if not intended for self onboarding
       if (!watchedSelfOnboarding || !isAdmin) {
-        //typeguard: in this scenario payload will contain address information
-        if (!payload.self_onboarding) {
```
serikjensen, line 246: This highlighted a lot of code. Only change here is removing this conditional

dmortal (approved):

PR #162, opened 2025-02-24
chore: fix & elide typescript errors; error fail CI build

scally:
For simple typescript errors I fixed them inline.

For more complicated ones I added a `@ts-expect-error` for us to go fix in a future PR so that we can get back to good as quickly as possible.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "./CHANGELOG.md"
   ],
   "scripts": {
-    "build": "npm run build:clean && npm run generate && vite build",
+    "build": "npm run build:clean && npm run generate && npm run tsc && vite build",
     "build:clean": "rm -rf ./dist && mkdir ./dist",
     "commitlint": "commitlint --edit",
     "dev": "node ./build/prompt.js && npm run build && vite build --watch",
--- a/src/components/Company/DocumentSigner/DocumentList/Head.tsx
+++ b/src/components/Company/DocumentSigner/DocumentList/Head.tsx
@@ -1,6 +1,7 @@
 import { useTranslation } from 'react-i18next'
 
 export function Head() {
+  // @ts-expect-error HACK missing translations
   const { t } = useTranslation('Company.DocumentSigner')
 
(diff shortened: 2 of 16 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "./CHANGELOG.md"
   ],
   "scripts": {
-    "build": "npm run build:clean && npm run generate && vite build",
+    "build": "npm run build:clean && npm run generate && npm run tsc && vite build",
```
scally, line 31: This felt like the best place to insert the typecheck. Broken types won't break your local dev build while you're working but will break both CI and publish here. There's some room for efficiency to run tsc and vite in parallel here but didn't feel worth right now.

tsconfig.json
```diff
@@ -12,7 +12,7 @@
     "module": "ESNext",
     "moduleDetection": "force",
     "allowJs": true,
-    "noEmit": false,
+    "noEmit": true,
```
scally, line 15: We don't actually compile with `tsc`, just typecheck. `vite` runs our compilation step via `esbuild`, so we can save some time/duplication by just telling `tsc` not to emit files.

src/components/Company/DocumentSigner/DocumentList/Head.tsx
```diff
@@ -1,6 +1,7 @@
 import { useTranslation } from 'react-i18next'
 
 export function Head() {
+  // @ts-expect-error HACK missing translations
```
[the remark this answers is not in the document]
serikjensen: no idea what happened here if it was a rename that didn't get updated, but for some reason it points to a file that doesn't exist. it should be 

```
const { t } = useTranslation('Company.DocumentList')
```

 I have a fix for it included in this PR https://github.com/Gusto/embedded-react-sdk/pull/159#discussion_r1964499897

i'm good with you just adding the comment here and then rebasing that signature form PR and fixing it there

serikjensen (approved):
Looks good! left comment about translation type error, it's already fixed in my signature form PR so i can just rebase that once this is merged

PR #164, opened 2025-02-25
feat: add company document signer state machine

serikjensen:
This adds a document signer state machine to put together all steps in company document signer.

See workflow in figma here https://www.figma.com/design/6dxOSiONDiJoa9zY1wOwbs/Frontend-SDK-Partner-Design-File?node-id=5220-17923&m=dev

### Assign signatory (as self) -> documents list -> signature form -> documents list
* Initialized on assign signatory form
* If verified as self signatory, continues to document list and is able to sign documents
* Sign documents leads to the signature form
* Signing the form leads back to documents list with the document status correctly updated
* You can return to assign signatory form and update user information

https://github.com/user-attachments/assets/f7cb2305-ad07-46fa-ab31-d047ce7cd312

### Inviting signatory
* Initialized on assign signatory form
* Not verified as self signatory so in next step is unable to sign documents (gets a not signed document status instead)

https://github.com/user-attachments/assets/de8ba642-dede-49ec-a136-3a730c019d90


src/components/Flow/hooks/useFlowParams.ts
```diff
@@ -6,24 +6,30 @@ export interface UseFlowParamsProps<TFlowContext> {
   requiredParams?: Array<keyof TFlowContext>
 }
 
+function hasRequiredParams<TParams extends object, TRequiredParams extends keyof TParams>(
+  params: TParams,
+  requiredParams: TRequiredParams[],
+): params is TParams & Required<Pick<TParams, TRequiredParams>> {
+  return requiredParams.every(param => param in params && typeof params[param] !== 'undefined')
+}
+
```
serikjensen, line 15: This provides a type guard for the existing useFlowParams hook below. We can assert that all the required params are supplied as expected and then communicate that with TS. 

Without this TS determines a prop which is conditional in the flow interface (by design since it's not required for each flow component, only signature form) and throws an error

src/types/Helpers.d.ts
```diff
@@ -11,3 +11,11 @@ export type DeepPartial<T> = {
...
   [K in keyof T]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<keyof T, K>>>
 }[keyof T]
+
+export type MachineEventType<
+  TEventPayloads,
+  TEventType extends keyof TEventPayloads = keyof TEventPayloads,
+> = {
+  type: TEventType
+  payload: TEventPayloads[TEventType]
+}
```
serikjensen, line 21: Needed to share this across state machines, lmk if there's a better file to put it in

src/components/Flow/hooks/useFlowParams.ts
jeffredodd: Nice! Looks good! Great use of types here

src/components/Flow/hooks/useFlowParams.ts
```diff
@@ -6,24 +6,30 @@ export interface UseFlowParamsProps<TFlowContext> {
...
-          param,
-          component,
-          provider: 'FlowProvider',
-        }),
-      )
-    }
-  })
+  if (!hasRequiredParams(params, requiredParams)) {
+    const missingParam = requiredParams.find(param => typeof params[param] === 'undefined')
+    throw new Error(
```
jeffredodd, line 25: What happens when the required params are missing? Do we just throw an error and break the flow? Just trying to think of cases where this could happen.

jeffredodd (approved):
Looks good! I don't feel confident enough in robot3 to be your approver but I think this looks like its on the right track.

src/components/Company/DocumentSigner/DocumentSigner.tsx
```diff
@@ -0,0 +1,28 @@
...
+  AssignSignatory,
+  type DocumentSignerContextInterface,
+} from './documentSignerStateMachine'
+
+export interface DocumentSignerProps extends BaseComponentInterface {
+  companyId: string
+  signatoryId?: string
+}
+
+export const DocumentSigner = ({ companyId, signatoryId, onEvent }: DocumentSignerProps) => {
```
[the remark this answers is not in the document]
serikjensen: So the thought around this is that `Flow` already is kind of a reserved word for the mega user flows like employee onboarding. These are still technically at the level of a component like payment method or compensation so the idea would be that they are usable in the same way.

I guess it would come down to if we felt ok making flow something that could be at the component level which is tough given that we have language around what flows mean atm

i also might be overthinking this lol here's the current docs on flows https://docs.gusto.com/embedded-payroll/docs/workflows-overview

src/components/Company/DocumentSigner/documentSignerStateMachine.tsx
```diff
@@ -0,0 +1,150 @@
...
+import { companyEvents } from '@/shared/constants'
+
+import * as Company from '@/components/Company'
+import { useFlowParams, type UseFlowParamsProps } from '@/components/Flow/hooks/useFlowParams'
+import { Schemas } from '@/types/schema'
+import { FlowContextInterface } from '@/components/Flow'
+import { type MachineEventType } from '@/types/Helpers'
+
+type EventPayloads = {
+  [companyEvents.COMPANY_VIEW_FORM_TO_SIGN]: Schemas['Form']
```
dmortal, line 11: I wonder if we should further scope these events for easier management, like `documentSignerEvents.`

src/components/Company/DocumentSigner/DocumentSigner.tsx
```diff
@@ -0,0 +1,28 @@
...
+  const manageEmployees = createMachine(
```
dmortal, line 17: manageEmployees? Should this be documentSignerMachine?

dmortal:
this is awesome!

src/components/Company/DocumentSigner/DocumentSigner.tsx
serikjensen: Yes caught my copy pasta! 😅

src/components/Company/DocumentSigner/documentSignerStateMachine.tsx
serikjensen: yeah +1 to providing better scope!

PR #165, opened 2025-02-26
feat: speakeasy integration

dmortal:


scally (approved):
Nice approach to slicing this work 👍🏻

PR #166, opened 2025-02-26
chore: remove commented out files to clean up repo

jeffredodd:
This PR aims to remove all commented out files. Thanks to the work of @scally we can confirm our types and build are working as expected when these files are removed.


scally (approved):
🧹 🧹

dmortal (approved):

PR #170, opened 2025-02-28
feat: introduce payschedule components and forms

jeffredodd:
Setting up PaySchedule Draft PR to get early feedback while finishing UX of PayPreviewCard


src/components/Company/PaySchedule/PaySchedule.tsx
```diff
@@ -0,0 +1,300 @@
...
+}
+
+export const PaySchedule = (props: PayScheduleProps & BaseComponentInterface) => {
+  useI18n('Company.PaySchedule')
+  return (
+    <BaseComponent {...props}>
+      <Root {...props}>{props.children}</Root>
+    </BaseComponent>
+  )
+}
```
serikjensen, line 117: Nit: I think we need to destructure or pick the PayScheduleProps here and pass them explicitly to root, and then the rest can go to the BaseComponent as BaseComponent props. Not a huge deal as it is currently, just ends up passing some extra props to BaseComponent that it doesn't use

src/components/Company/PaySchedule/PaySchedule.tsx
```diff
@@ -0,0 +1,300 @@
...
+      day_2: allValues.day_2 || undefined,
+    })
+  }, [
+    allValues.anchor_end_of_pay_period,
+    allValues.anchor_pay_date,
+    allValues.day_1,
+    allValues.day_2,
+    allValues.frequency,
+    setPayScheduleDraft,
+  ])
```
serikjensen, line 183: Nit: this seems pretty performance intensive? and if i'm reading this correct this would set the pay schedule draft which would also retrigger the useGetPaySchedulePreview hook? Maybe not a huge deal since it's a limited number of inputs but wondering if there are any perceivable performance issues

src/components/Company/PaySchedule/_parts/Edit.tsx
```diff
@@ -0,0 +1,129 @@
...
+  }, [custom2xPerMonth, frequency, setValue])
+
+  if (mode !== 'EDIT_PAY_SCHEDULE' && mode !== 'ADD_PAY_SCHEDULE') {
+    return null
+  }
+
+  return (
+    <div style={{ width: '100%' }}>
+      <Grid gap={4} gridTemplateColumns={{ base: '1fr', small: '1fr 1fr' }}>
+        <div style={{ width: '100%' }}>
```
serikjensen, line 47: Nit: we could probs set as className instead of inline styles

serikjensen (commented):
Looking good! It matches up with the existing architectural approaches, and there are no immediately obvious concerns. I left some questions and nits. Probably my biggest question/suggestion would be around the error handling in the useGetPaySchedulePreview  query, more on that in the comments. Excellent work!

src/components/Company/PaySchedule/PaySchedule.tsx
jeffredodd: In terms of perceivable impact it has been minimal. Since we're just changing dates or numbers we're not facing an issue where this needs to be debounced for performance reasons. But this is a good call out i'll triple check we're not facing perf issues.

src/components/Company/PaySchedule/_parts/Edit.tsx
jeffredodd: Yes good call out

src/api/queries/payschedule.ts
```diff
@@ -95,10 +96,27 @@ export function useUpdatePaySchedule(
...
+      } catch (err) {
+        if (err instanceof ApiError) {
+          throw new Error(
+            JSON.stringify({
+              message: err.message,
+              status: err.statusCode,
+              payload: err.errorList,
+            }),
+          )
+        }
```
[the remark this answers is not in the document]
jeffredodd: Update! I tried this! A few things here at play. One, we have to `throw Error` inside the queryFn for this useQuery to populate their `error` data. 

So i tried passing just `err` in the hopes that I can parse it from `PaySchedule.tsx` but it seems it doesn't want to pass an object that far, only a string.... 

Thoughts here?

.vscode/settings.json
```diff
@@ -0,0 +1,22 @@
+{
```
dmortal, line 1: 🙇‍♂️

src/api/queries/payschedule.ts
```diff
@@ -95,10 +96,27 @@ export function useUpdatePaySchedule(
 export function useGetPaySchedulePreview(
   company_id: string,
   params: Operations['get-v1-companies-company_id-pay_schedules-preview']['parameters']['query'],
+  enabled = false,
 ) {
   const { GustoClient: client } = useGustoApi()
-  return useSuspenseQuery({
+  return useQuery({
```
dmortal, line 102: why `useQuery` as opposed to `useSuspense`?

src/api/queries/payschedule.ts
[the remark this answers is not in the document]
dmortal: I would say ignore it for now - Speakeasy does it differently and we are moving away from `ApiError`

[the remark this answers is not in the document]
dmortal: and we will have better type support for the shape of error response - it's tricky

dmortal (commented):
It's a mighty big diff

PR #172, opened 2025-03-01
chore: add tests for create signatory and signature form

serikjensen:
This adds the rest of the tests for company document signer. It provides coverage for the create signatory and signature form, primarily around ensuring events are correctly firing as expected.

### DIFF
--- a/src/components/Company/AssignSignatory/CreateSignatory/CreateSignatory.test.tsx
+++ b/src/components/Company/AssignSignatory/CreateSignatory/CreateSignatory.test.tsx
@@ -1,0 +1,186 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { render, screen, waitFor, within } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { HttpResponse } from 'msw'
+import { CreateSignatory } from './CreateSignatory'
+import { GustoTestApiProvider } from '@/test/GustoTestApiProvider'
+import { setupApiTestMocks } from '@/test/mocks/apiServer'
+import { companyEvents } from '@/shared/constants'
+import {
+  handleGetAllSignatories,
+  handleCreateSignatory,
+  handleUpdateSignatory,
+} from '@/test/mocks/apis/company_signatories'
+import { server } from '@/test/mocks/server'
+
+describe('CreateSignatory', () => {
+  const mockOnEvent = vi.fn()
+
+  beforeEach(() => {
+    setupApiTestMocks()
+    mockOnEvent.mockClear()
+    vi.resetModules()
+  })
+
+  describe('when user is creating a signatory', () => {
+    beforeEach(() => {
+      server.use(
+        handleCreateSignatory(() =>
+          HttpResponse.json({
+            uuid: 'new-signatory-uuid',
+            first_name: 'Michael',
+            last_name: 'Bluth',
+            email: 'michael.bluth@example.com',
+            title: 'President',
+          }),
+        ),
+      )
+    })
+
+    it('fires the correct created events when form is submitted successfully', async () => {
+      const user = userEvent.setup()
+
+      render(
+        <GustoTestApiProvider>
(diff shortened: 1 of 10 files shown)


src/components/Company/DocumentSigner/SignatureForm/SignatureForm.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+  beforeEach(() => {
+    setupApiTestMocks()
+    mockOnEvent.mockClear()
+    vi.resetModules()
+
+    server.use(handleSignCompanyForm(() => HttpResponse.json(testForm)))
+  })
+
+  it('fires the correct events when form is submitted successfully', async () => {
+    const user = userEvent.setup()
```
jeffredodd, line 43: Is this needed? I've legit never used this with RTL before

src/components/Company/DocumentSigner/SignatureForm/SignatureForm.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+    await waitFor(() => {
+      expect(screen.getByText(/Signature required for/)).toBeInTheDocument()
+    })
+
+    await user.type(screen.getByLabelText('Signature'), 'John Doe')
+
+    const checkbox = screen.getAllByLabelText(
+      'I am the signatory and I agree to sign electronically',
+    )[0]
```
jeffredodd, line 59: Are there two labels with the same text? That could be an a11y issue but i'm not sure. Just guessing because getByLabel uses a role/element to do so.

jeffredodd (commented):
Looks good! Just a few questions

src/components/Company/DocumentSigner/SignatureForm/SignatureForm.test.tsx
serikjensen: Yes! see docs here https://testing-library.com/docs/user-event/setup

I believe it could technically be called in a beforeEach, but that would mean sharing state across tests which could result in some unexpected behaviors

src/components/Company/DocumentSigner/SignatureForm/SignatureForm.test.tsx
serikjensen: Yeah this is our lingering single input checkbox group issue that keeps surfacing. both the group and the checkbox share the same label. In a world where we were off react aria, we could have our own single checkbox validation and this would go away. Not sure what the best workaround is in the meantime.

serikjensen: One quick stopgap would be to provide a distinct label for the group from the label on the checkbox input itself. maybe we can chat about it when you get in

src/components/Company/DocumentSigner/SignatureForm/SignatureForm.test.tsx
jeffredodd: Interesting TIL

jeffredodd: Interesting i've never used this before. TIL

jeffredodd (approved):

src/components/Company/AssignSignatory/CreateSignatory/CreateSignatory.test.tsx
```diff
@@ -0,0 +1,186 @@
...
+      })
+      await user.click(titleControl)
+      const ownerOption = screen.getByRole('option', { name: 'Owner' })
+      await user.click(ownerOption)
+
+      const dateOfBirthInput = screen.getByRole('group', { name: 'Date of birth' })
+      await user.type(within(dateOfBirthInput).getByRole('spinbutton', { name: /month/i }), '01')
+      await user.type(within(dateOfBirthInput).getByRole('spinbutton', { name: /day/i }), '01')
+      await user.type(within(dateOfBirthInput).getByRole('spinbutton', { name: /year/i }), '1995')
+
```
serikjensen, line 71: Date of birth input testing

PR #173, opened 2025-03-05
chore: remove outdated generated schema

serikjensen:
This PR removes the outdated schema.d.ts file located in the generated directory in favor of a single schema.d.ts file.

## Context
We had two generated schema.d.ts files. One in `src/types/schema.d.ts`. The other in `src/generated/schema.d.ts`

Only the schema file in `src/types/schema.d.ts` was getting updated when running `npm run openapi`. The `src/generated/schema.d.ts` file was not getting updated.

Doing further investigation, looking at the old repository, we can see the following https://github.com/Gusto/embedded-javascript-sdk/blob/main/packages/embedded-api/build/generateOpenApiFiles.js where the schema is getting output in the generated directory.

This combined with the fact that it is not updating and not included in our gitignore indicate it is a holdover from the previous repository and can be removed.


scally (approved):
🧹 🧹 🧹

jeffredodd (approved):
This looks sane to me (LSTM). and if we have to revert this its an easy fix so go for it. 🚢

PR #174, opened 2025-03-05
test: employee self-onboarding flow

scally:
Proof that it can detect the bug fixed here https://github.com/Gusto/embedded-react-sdk/pull/160 :

Screenshot 2025-03-10 at 2 42 11 PM


scally:
Some discussion before this goes through:

* My initial stance is that I'm not going to merge with all the stubs inlined the way they are, but will tuck them in using either our existing patterns for mocks or some derivation of it. I have left them all here intentionally during initial review to make it super, super clear how many dependencies this test really has. I am open to some conversation here as to how much effort we want to make to try to magically hide all the mocks. 
* The test is written to stub only what is required to make the tests pass with no additional data. This is also intentional to try to keep this test as non-flakey as possible. 
* `fillDate` is a codification of tricks we use in other tests to fill our special date fields. We have room to add a `selectOption` function as well, as lines 115-119 will attest. I will refactor this out to a different location before this goes through to main, and in the future we can refactor some tests to use `fillDate`. It will likely need to change when we move to a non-react-aria component system so how much work we want to do to clean it up is definitely up for debate.

What do we think? Is it worth it? Do we want to commit to keeping some tests like this in the future?

scally:
Runtime stats for one test of this class (JSDOM + "e2e" flow)

Screenshot 2025-03-10 at 3 01 40 PM

serikjensen:
> I think there is value in e2e tests but looking at the singular test you've added it sure looks like a lot of work. I'm curious if others find value in this.
> 
> Has this test been flakey at all? Is it passing consistently?

I kinda had the opposite reaction 😅 i thought this would be longer/more involved. Most overhead comes from the mocks, but as mentioned above we can consider how much we want these to live in the constellation of other mocks or stay defined here

IMO this seems like a good baseline smoke test for the essential flow functionality and would've caught the bug experienced by BQE. I could see a similar one for our key flows like employee onboarding, and eventually company onboarding.

serikjensen:
I think the biggest thing this tests is that the transitions between pages are succeeding and there are no validation errors. I think if we opted to go with more of a unit test approach instead of this we would just make sure that the event for successful submission of forms/continuing to next step was firing on each step (which is basically what i have with company document signer). I'd be fine with focusing on that approach instead if we don't like this, it keeps things tested more at the unit level. a test like that would've also caught the bqe issue

This one really tests that the state machine is correctly capturing those events and managing the transition which is probably safer. So really it's a matter of deciding how much we care about that level of coverage

scally:
> Has this test been flakey at all? Is it passing consistently?
I've manually re-run 20 times, 0 failures

scally:
> Most overhead comes from the mocks, but as mentioned above we can consider how much we want these to live in the constellation of other mocks or stay defined here

We can tuck the mocks away, but that won't change either the number of true dependencies this test has, or its cost to write/maintain/run.

It's also not going to give particularly clear feedback. You will see where it fails but you'll have to do some investigating across a couple of components at least to determine where/why.

I am totally OK either shipping or not shipping this, to be clear. I'd like to have decided as a team beforehand how we feel about tradeoffs 👍🏻

jeffredodd:
I think overall i'm willing to accept the tradeoffs to have a canary in the testing codebase. I also think the normalization of mocks etc will help this to look more sane. But yeah... I think you got my vote :)

scally:
Update:

FWIW, pulling all the MSW mock calls out of the e2e test we have so far is not super satisfying because we have ignored type errors in our typeHelpers it uses to strongly type all the operations/params/responses. 

I know it's something likely to change / be revisited during the SpeakEasy conversion, so I am not super eager to spend a lot of time writing throwaway code to make a proper testing API for creating mock handlers. 

As a result, some of the msw handlers should be accepting alternative responses and currently don't

PR #175, opened 2025-03-05
chore: update open api and form signing queries

serikjensen:
This PR:
* Updates open api to bring in change where `signed_by_ip_address` is now optional https://github.com/Gusto/zenpayroll/pull/263939
* Updates to remove the stubbed out `signed_by_ip_address` value from the queries, since per https://github.com/Gusto/gws-flows/pull/2704 the ip address has already been configured in the header for gws flows which means it can be removed on the SDK side of things
* Small fixes with the pdf url type which introduced null as a value with the open api update

### Company document signer working as before
https://github.com/user-attachments/assets/cd82727c-9a67-43aa-8b13-9857e0147c12

### Employee document signer working as before
https://github.com/user-attachments/assets/861720bc-c7de-434a-af4d-7113d2b35c41


src/components/Employee/Deductions/Deductions.tsx
```diff
@@ -139,7 +139,12 @@ export const Root = ({ employeeId, className }: DeductionsProps) => {
       //Deletion of deduction is simply updating it with active: false
       const updateMutationResponse = await updateDeductionMutation.mutateAsync({
         garnishment_id: payload.uuid,
-        body: { ...payload, active: false, version: payload.version as string },
+        body: {
+          ...payload,
+          total_amount: payload.total_amount ?? undefined,
```
serikjensen, line 144: After updating openapi, total_amount in the body here needed to be string or undefined but it was optionally null from the payload. Update to force it to undefined if null to satisfy the type

dmortal (approved):

PR #176, opened 2025-03-05
docs: add ip address guidance for partners

serikjensen:
This adds more documentation about configuring the backend proxy, especially around configuring it for recent updates to the IP address.

For further context, the IP address must be provided for form signing operations to be completed. We are unable to provide that as a client side lib so this is a partner ask when they are configuring their backend proxy.

We merged an update to ZP here https://github.com/Gusto/zenpayroll/pull/263939  to allow for the IP address to be supplied via a request header in a single place in the proxy. Here is an example of configuring it in our own proxy for GWS flows for the SDK https://github.com/Gusto/gws-flows/pull/2704

### DIFF
--- a/docs/03/getting-started-1.md
+++ b/docs/03/getting-started-1.md
@@ -26,9 +26,28 @@ function App() {
 export default App
 ```
 
-The `baseUrl` property is typically configured with the address of a backend proxy. The recommended approach is to use a backend service that acquires OAuth2 tokens from the Gusto Embedded API for authenticated users and proxies API calls using those tokens. Learn more about configurintg this and setting up authentication in the `Authentication` section.
+The `baseUrl` property is configured with the address of your backend proxy which is detailed further in the following section.
 
-For more configurations available on the GustoApiProvider, see the docs in the `Styles and Theming` and `Translation` sections.
+## Configuring a backend proxy
+
+When building with the React SDK, a backend proxy is required. React SDK components do not make calls to the Gusto Embedded API directly. Instead, the `baseUrl` configuration defines the URL of your proxy server. This proxy layer gives you complete control over requests sent to Gusto, which is essential for:
+
+1. Authentication
+2. Providing the user IP address for form signing operations
+
+The React SDK is designed to mirror the [Gusto Embedded API Reference](https://docs.gusto.com/embedded-payroll/reference/whats-new-in-v2024-04-01) with a 1:1 mapping of endpoints. The SDK maintains consistent naming conventions, parameters, and response structures with the Gusto API.
+
+Your proxy server simply needs to forward any incoming SDK requests to the corresponding Embedded API endpoints. The proxy's main task is adding the necessary authentication headers before forwarding the request onwards. Since the SDK requests are already in the Embedded API format, no extra endpoint mapping or request transformation is required.
+
+### Using the proxy for authentication
+
+The proxy layer allows for authentication. The recommended approach is to use a backend service that acquires OAuth2 tokens from the Gusto Embedded API for authenticated users and proxies API calls using those tokens. Learn more about configuring this and setting up authentication in the `Authentication` section.
+
+### Using the proxy to provide the user IP address
+
+Some UI workflows require users to sign forms, which need the user's IP address for security purposes. To prevent vulnerabilities such as IP address spoofing, this information must be provided by your proxy server rather than collected client-side.
+
+Your proxy server can provide the IP address by adding the `x-gusto-client-ip` header with the user IP address to all forwarded requests on the backend. By setting this header once in your proxy it will be configured for all form signing operations.
 
 ## Including styles
 


docs/03/getting-started-1.md
```diff
@@ -26,9 +26,28 @@ function App() {
...
+
+Your proxy server simply needs to forward incoming SDK requests to the corresponding Gusto API endpoints. No special endpoint mapping or request transformation is required, as the SDK requests already match the Gusto API format. The proxy's main task is adding the necessary authentication headers before forwarding the request.
+
+### Using the proxy for authentication
+
+The proxy layer allows for authentication. The recommended approach is to use a backend service that acquires OAuth2 tokens from the Gusto Embedded API for authenticated users and proxies API calls using those tokens. Learn more about configuring this and setting up authentication in the `Authentication` section.
+
+### Using the proxy to provide the user IP address
+
+Some UI workflows require users to sign forms, which need the user's IP address for security purposes. Since the SDK runs in the browser, it cannot reliably access the IP address.
```
jeffredodd, line 48: I think it might be worth calling out more that the client could spoof the IP instead of the lack of the browsers ability to get the IP. Because I think there are many ways to get the IP but none of them provide the IP in a way which can't be spoofed.

serikjensen: Good callout! updated the wording, lmk if that accounts for the above

jeffredodd: Looks good!

dmortal (approved):
Thanks!

willbetts (commented):
- 'The React SDK is designed to mirror the [Gusto Embedded API Reference](https://docs.gusto.com/embedded-payroll/reference/whats-new-in-v2024-04-01) with a 1:1 mapping of endpoints. The SDK maintains consistent naming conventions, parameters, and response structures with the Gusto API. Your proxy server simply needs to forward incoming SDK requests to the corresponding Gusto API endpoints. No special endpoint mapping or request transformation is required, as the SDK requests already match the Gusto API format. The proxy's main task is adding the necessary authentication headers before forwarding the request. On this point, I think an example would be helpful, IE, here is an example of an API call that the SDK would make to your backend (ie URL, params, body) and here is the corresponding request that would be sent to the Gusto API.
- The second piece of feedback is around the IP. isn't this specific to the signatory/company onboarding? If so, wouldn't this be premature to put in our docs since company onboarding hasn't been released?

serikjensen:
> * 'The React SDK is designed to mirror the [Gusto Embedded API Reference](https://docs.gusto.com/embedded-payroll/reference/whats-new-in-v2024-04-01) with a 1:1 mapping of endpoints. The SDK maintains consistent naming conventions, parameters, and response structures with the Gusto API. Your proxy server simply needs to forward incoming SDK requests to the corresponding Gusto API endpoints. No special endpoint mapping or request transformation is required, as the SDK requests already match the Gusto API format. The proxy's main task is adding the necessary authentication headers before forwarding the request. On this point, I think an example would be helpful, IE, here is an example of an API call that the SDK would make to your backend (ie URL, params, body) and here is the corresponding request that would be sent to the Gusto API.
> * The second piece of feedback is around the IP. isn't this specific to the signatory/company onboarding? If so, wouldn't this be premature to put in our docs since company onboarding hasn't been released?

on the first point, great idea! i'm going to update [this ticket](https://gustohq.atlassian.net/browse/GWS-3883) to be adding an example. I need to think more on the best medium to provide that (ex. an example proxy implementation in a sandbox, or code snippet in the docs, or some combo of the two)

on the second point, they would run into the signatory ip issue for both company onboarding and employee onboarding which is released today. BQE is currently building employee onboarding which has a document signer step so that was the motivation behind getting this out there: to ensure they had a way to properly configure the ip address for that. There are also other form signing ops for things like contractors, i9, etc that will be needed in the future and the hope is that we can address this in a single holistic way and allow them to not worry about every individual case

docs/03/getting-started-1.md
```diff
@@ -26,9 +26,28 @@ function App() {
...
+## Configuring a backend proxy
+
+When building with the React SDK, a backend proxy is required. React SDK components do not make calls to the Gusto Embedded API directly. Instead, the `baseUrl` configuration defines the URL of your proxy server. This proxy layer gives you complete control over requests sent to Gusto, which is essential for:
+
+1. Authentication
+2. Providing the user IP address for form signing operations
+
+The React SDK is designed to mirror the [Gusto Embedded API Reference](https://docs.gusto.com/embedded-payroll/reference/whats-new-in-v2024-04-01) with a 1:1 mapping of endpoints. The SDK maintains consistent naming conventions, parameters, and response structures with the Gusto API.
+
+Your proxy server simply needs to forward incoming SDK requests to the corresponding Gusto API endpoints. No special endpoint mapping or request transformation is required, as the SDK requests already match the Gusto API format. The proxy's main task is adding the necessary authentication headers before forwarding the request.
```
TheHumanJenkins, line 40:
```suggestion
Your proxy server simply needs to forward any incoming SDK requests to the corresponding Embedded API endpoints. The proxy's main task is adding the necessary authentication headers before forwarding the request onwards. Since the SDK requests are already in the Embedded API format, no extra endpoint mapping or request transformation is required. 
```

docs/03/getting-started-1.md
```diff
@@ -26,9 +26,28 @@ function App() {
...
+
+### Using the proxy for authentication
+
+The proxy layer allows for authentication. The recommended approach is to use a backend service that acquires OAuth2 tokens from the Gusto Embedded API for authenticated users and proxies API calls using those tokens. Learn more about configuring this and setting up authentication in the `Authentication` section.
```
TheHumanJenkins, line 44: General question: what's the recommended approach for accessing different company access tokens or refresh tokens via the proxy? 


It's super clear on `/company/{companyID}` endpoints; but how to handle for non company based routes like `/employees/{employee_id}/home_addresses`?  Say I have 10 partner managed companies.

docs/03/getting-started-1.md
```diff
@@ -26,9 +26,28 @@ function App() {
...
+
+### Using the proxy to provide the user IP address
+
+Some UI workflows require users to sign forms, which need the user's IP address for security purposes. To prevent vulnerabilities such as IP address spoofing, this information must be provided by your proxy server rather than collected client-side.
```
TheHumanJenkins, line 48: So the partner's server IP address is fine?

docs/03/getting-started-1.md
serikjensen: Good question! This is meant to be an extremely lightweight explanation that authentication will happen in the proxy layer and not the SDK. The reference there for learning more about configuring authentication is located here https://docs.gusto.com/embedded-payroll/docs/authentication-1 Lmk if that doc does a better job of clarifying, if not we can probably create a separate ticket to update that page with more details

docs/03/getting-started-1.md
serikjensen: They would be retrieving the user IP address from the request and then setting it via the `x-gusto-client-ip` header. Lmk if there's better wording here to describe that! attempted to explain further below

TheHumanJenkins: Sounds good, it is on the partner to retrieve the clients IP address possibly when the SDK is launched or access.

docs/03/getting-started-1.md
TheHumanJenkins: More details to come on this, resolving for now.

TheHumanJenkins (approved):

PR #181, opened 2025-03-06
style: add styling to SDK internal error component

aaronlee777:
Screenshot 2025-03-06 at 3 36 47 PM

### DIFF
--- a/src/components/Common/InternalError/InternalError.module.scss
+++ b/src/components/Common/InternalError/InternalError.module.scss
@@ -1,5 +1,6 @@
 h1 {
   font-size: toRem(21) !important;
+  color: var(--g-colors-gray-1000);
 }
 
 .internalErrorCard {
--- a/src/components/Common/InternalError/InternalError.module.scss
+++ b/src/components/Common/InternalError/InternalError.module.scss
@@ -1,0 +1,27 @@
+h1 {
+  font-size: toRem(21) !important;
+}
+
+.internalErrorCard {
+  background-color: #fff;
+  border-right: 1px solid #eaeaea;
+  border-bottom: 1px solid #eaeaea;
+  border-left: 1px solid #eaeaea;
+  box-shadow:
+    0px 1px 18px rgba(0, 0, 0, 0.08),
+    inset 0 4px 0 #d5351f;
+  padding: toRem(20);
+  display: flex;
+  flex-direction: column;
+  gap: toRem(8);
+  max-width: toRem(600);
+  margin: 0 auto;
+}
+
+.errorMessage {
+  font-size: toRem(14);
+  font-weight: 500;
+  color: #d5351f;
+  line-height: normal;
+  margin-top: toRem(6);
+}
--- a/src/components/Common/InternalError/InternalError.module.scss
+++ b/src/components/Common/InternalError/InternalError.module.scss
@@ -3,13 +3,13 @@ h1 {
 }
 
 .internalErrorCard {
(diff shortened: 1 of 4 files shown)


src/components/Common/InternalError/InternalError.tsx
```diff
@@ -13,14 +14,23 @@ export const InternalError = ({ error, resetErrorBoundary }: FallbackProps) => {
...
-    <section role="alert">
-      <Trans
-        t={t}
-        i18nKey="errors.globalReactError"
-        values={{ error: errorMessage }}
-        shouldUnescape={true}
-      />
-      <Button onPress={resetErrorBoundary}>{t('errors.resetGlobalError')}</Button>
-    </section>
+    <div className={styles.internalErrorCard}>
```
dmortal, line 17: please keep `role='alert'` -> this is needed for accessibility

src/components/Common/InternalError/InternalError.tsx
```diff
@@ -13,14 +14,23 @@ export const InternalError = ({ error, resetErrorBoundary }: FallbackProps) => {
...
+      <div>
+        <h1 className={styles.internalErrorCardTitle}>Error</h1>
```
dmortal, line 19: this should be internationalized

src/components/Common/InternalError/InternalError.module.scss
```diff
@@ -0,0 +1,27 @@
+h1 {
+  font-size: toRem(21) !important;
+}
```
serikjensen, line 3: I think if you change this to

```suggestion
.internalErrorCardTitle {
  font-size: toRem(21);
}
```

it should work (unless this is what you already tried)

You're already applying the `internalErrorCardTitle` class via className so i would hope that the specificity of the css module class would beat out the h1 element styles

src/components/Common/InternalError/InternalError.module.scss
```diff
@@ -0,0 +1,27 @@
...
+  box-shadow:
+    0px 1px 18px rgba(0, 0, 0, 0.08),
+    inset 0 4px 0 #d5351f;
+  padding: toRem(20);
+  display: flex;
+  flex-direction: column;
+  gap: toRem(8);
+  max-width: toRem(600);
+  margin: 0 auto;
+}
```
serikjensen, line 19: Nit: should we consider using colors here for error rather than hardcoding these hex values? otherwise these won't be themeable for partners

src/components/Common/InternalError/InternalError.module.scss
```diff
@@ -0,0 +1,27 @@
...
+
+.errorMessage {
+  font-size: toRem(14);
+  font-weight: 500;
```
serikjensen, line 23: We can use a variable here 

```suggestion
  font-weight: var(--g-typography-fontWeight-medium);
```

src/components/Common/InternalError/InternalError.module.scss
```diff
@@ -0,0 +1,27 @@
...
+  color: #d5351f;
```
serikjensen, line 24: Nit: same as above, are there potentially colors from our theme variables we can reference here instead?

src/components/Common/InternalError/InternalError.module.scss
aaronlee777: I tried this before and still get the same thing :(

serikjensen (approved):
LGTM!

PR #184, opened 2025-03-07
feat: introduce calendardisplay + add to payschedule

jeffredodd:
This PR focuses on introducing both the CalendarDisplay component as well as integrating it into PaySchedule.tsx. 

## Screenshots
Screenshot 2025-03-13 at 1 58 02 PM
Screenshot 2025-03-13 at 1 55 58 PM


src/components/Common/CalendarDisplay/CalendarDisplay.tsx
```diff
@@ -0,0 +1,130 @@
...
+  highlightDates,
+  selectionControl,
+}: CalendarDisplayProps) => {
+  const highlighter = (date: DateValue) => {
+    if (highlightDates) {
+      const highlight = highlightDates.find(h => h.date.toString() === date.toString())
+      if (highlight) {
+        return highlight.highlightColor
+      } else if (!isInRange(date)) {
+        return 'gray'
```
serikjensen, line 44: Nit: continuing to think about theming here, would it be possible to create highlight theme variables for the calendar theme and have this read that instead? I'm wondering if the data-highlight attribute could take a css var instead

ex. something like defining highlight color with color agnostic language

```
highlightColor: 'primary' | 'secondary'
```

then we could have theme vars for these in createTheme.ts

```
  calendarDisplay: {
    primaryHighlight: 'orange', // or some orange hex value
    secondaryHighlight: colors.gray[400],
  }
```

then do something like the following to read the theme variable

```
  return 'var(--g-calendarDisplay-${highlightColor}Highlight)'
```

We would potentially need to get more creative if the data attribute can't take a variable.

Basically, TL;DR, i think it could be valuable to make this themeable

src/components/Company/PaySchedule/_parts/Edit.tsx
```diff
@@ -28,13 +36,20 @@ export const Edit = () => {
     }
   }, [frequency, customTwicePerMonth, setValue])
 
+  // This is a workaround to ensure that the pay period preview range is set when the selected pay period index changes
+  useEffect(() => {
+    if (payPeriodPreviewRange === undefined) {
+      setValue('pay_period_preview_range', selectedPayPeriodIndex)
+    }
+  }, [selectedPayPeriodIndex, setValue, payPeriodPreviewRange])
+
```
serikjensen, line 45: Nit: are we going to end up with object equality check for the payPeriodPreviewRange in the dep array? Will that reliably fire when payPeriodPreviewRange changes? Could we consider running this as part of a callback event for the pay period changing which would only run when needed?

src/components/Company/PaySchedule/_parts/Edit.tsx
```diff
@@ -92,26 +107,50 @@ export const Edit = () => {
...
-            </Flex>
+              rangeSelected={{
+                start: payPeriodPreview[selectedPayPeriodIndex].start_date || '',
+                end: payPeriodPreview[selectedPayPeriodIndex].end_date || '',
+                label: t('payPreview.payPeriod') || 'Pay Period',
+              }}
+              highlightDates={[
+                {
+                  date: payPeriodPreview[selectedPayPeriodIndex].check_date || '',
+                  highlightColor: 'black',
```
serikjensen, line 144: Nit: since the consumer isn't going to have access to these props when they are being used deep inside these PaySchedule components i think the highlightColor should be set at the theming level so that it can be customized.

src/styles/_DatePicker.scss
```diff
@@ -52,7 +52,177 @@
...
+          left: 50%;
+          transform: translate(-50%, -50%);
+        }
+      }
+    }
+  }
+
+  // Calendar and Range Calendar commons tyles
+  .react-aria-Calendar,
+  .react-aria-RangeCalendar {
```
serikjensen, line 225: Nit: i think you could make use of the `toRem` function in this file instead of hardcoding px values

Would def increase the browser default font size and see if the design holds together

serikjensen (commented):
Looking good! only blocker is there are changes in the package-lock.json that seem to have been accidentally committed. The rest are just optional nits

package-lock.json
```diff
@@ -968,74 +968,6 @@
         "url": "https://github.com/sponsors/wooorm"
       }
     },
-    "node_modules/@esbuild/aix-ppc64": {
```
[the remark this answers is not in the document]
jeffredodd: Yeah I can dump it and pull from main. Not sure why mine is resolving this way.

src/components/Common/CalendarDisplay/CalendarDisplay.tsx
```diff
@@ -0,0 +1,130 @@
...
+import { CalendarDisplayLegend } from './CalendarDisplayLegend'
+
+// TODO: Disabled doesn't seem to work and hovering dates looks weird in this case
+
+export type CalendarDisplayProps = {
+  onNext?: () => void
+  onPrevious?: () => void
+  highlightDates?: Array<{
+    date: string
+    highlightColor: 'orange' | 'black'
```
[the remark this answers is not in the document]
jeffredodd: I think i'll start with 'primary' and 'warning' because thats kinda what we're doing.

src/components/Common/CalendarDisplay/CalendarDisplay.tsx
jeffredodd: Thanks for such a detailed comment, your comment above sent me down this same path so thank you!

src/components/Company/PaySchedule/_parts/Edit.tsx
jeffredodd: Long story short, this is needed because we are using RHF to use a dumb select to control what pay period is shown. Once we separate RHF from Select this code can be deleted easily.

src/components/Company/PaySchedule/_parts/Edit.tsx
jeffredodd: Let me know if the theming i've done here covers it

src/styles/_DatePicker.scss
jeffredodd: Done!

src/components/Common/CalendarDisplay/CalendarDisplay.tsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+  selectionControl?: React.ReactNode
+}
+
+export const CalendarDisplay = ({
+  rangeSelected,
+  highlightDates,
+  selectionControl,
+}: CalendarDisplayProps) => {
+  const highlighter = (date: DateValue) => {
```
dmortal, line 38: useMemo?

src/components/Common/CalendarDisplay/CalendarDisplay.tsx
```diff
@@ -0,0 +1,130 @@
...
+      const highlight = highlightDates.find(h => h.date.toString() === date.toString())
+      if (highlight) {
+        return highlight.highlightColor
+      } else if (!isInRange(date)) {
+        return ''
+      }
+    }
+  }
+
+  const isInRange = (date: DateValue) => {
```
dmortal, line 49: same, which of these functions can be memoized?

src/components/Common/CalendarDisplay/CalendarDisplayLegend.tsx
```diff
@@ -0,0 +1,55 @@
...
+    return new Intl.DateTimeFormat(undefined, {
+      weekday: 'long',
+      month: 'long',
+      day: 'numeric',
+      year: 'numeric',
+    }).format(adjustedDate)
+  }
+
+  return (
+    <div style={{ padding: '20px', borderTop: '1px solid var(--g-colors-gray-600)' }}>
```
dmortal, line 27: inline styles?

src/components/Common/CalendarDisplay/CalendarDisplayLegend.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <VisuallyHidden>Legend</VisuallyHidden>
```
dmortal, line 28: i18n

src/components/Common/CalendarDisplay/CalendarDisplayLegend.tsx
```diff
@@ -0,0 +1,55 @@
...
+      <Flex flexDirection="column" gap={24}>
+        {highlightDates?.map((highlight, index) => (
+          <Flex justifyContent="center" alignItems="center" key={index} gap={16}>
+            <div data-highlight={highlight.highlightColor} className="react-aria-CalendarLegend" />
+            <Flex flexDirection="column" gap={0}>
+              <Text
+                style={{
+                  fontSize: '14px',
```
dmortal, line 36: inline styles

src/components/Company/PaySchedule/_parts/Edit.tsx
dmortal: in the same context, should `useWatch` be used instead to isolate re-renders?

src/components/Company/PaySchedule/_parts/Edit.tsx
```diff
@@ -92,26 +108,50 @@ export const Edit = () => {
...
+              selectionControl={
+                !payPreviewLoading && (
+                  <Select
+                    control={control}
+                    name="pay_period_preview_range"
+                    label={t('labels.preview')}
+                    items={payPeriodPreview.map((period, index) => {
+                      return {
+                        id: index,
+                        name: `${period.start_date ? new Date(period.start_date).toLocaleDateString('en-US', { weekday: 'short', month: 'short', day: 'numeric' }) : ''} – ${period.end_date ? new Date(period.end_date).toLocaleDateString('en-US', { weekday: 'short', month: 'short', day: 'numeric' }) : ''}`,
```
dmortal, line 123: is this potentially reusable? Maybe this belongs in the helper method

src/components/Company/PaySchedule/_parts/Edit.tsx
```diff
@@ -92,26 +108,50 @@ export const Edit = () => {
...
-              )
-            })}
-          {payPreviewLoading && (
-            <Flex justifyContent={'center'} alignItems={'center'}>
-              <Spinner title={t('loading')} />
-            </Flex>
+              rangeSelected={{
+                start: payPeriodPreview[selectedPayPeriodIndex].start_date || '',
+                end: payPeriodPreview[selectedPayPeriodIndex].end_date || '',
+                label: t('payPreview.payPeriod') || 'Pay Period',
```
dmortal, line 140: why falling back on hardcoded string?

src/components/Company/PaySchedule/_parts/Edit.tsx
```diff
@@ -92,26 +108,50 @@ export const Edit = () => {
...
+              }}
+              highlightDates={[
+                {
+                  date: payPeriodPreview[selectedPayPeriodIndex].check_date || '',
+                  highlightColor: 'primary',
+                  label: t('payPreview.payday') || 'Payday',
```
dmortal, line 146: same

src/contexts/ThemeProvider/createTheme.ts
```diff
@@ -157,6 +157,13 @@ const createComponentThemes = ({
     },
     shadow: '0px 1px 2px rgba(0, 0, 0, 0.1)',
   },
+  calendarDisplay: {
+    primaryHighlight: colors.gray[1000],
+    warningHighlight: '#CA464A',
```
dmortal, line 162: why is this color outside color palette?

src/styles/_DatePicker.scss
```diff
@@ -52,7 +52,177 @@
...
+      border-spacing: 0;
+      width: 100%;
+      table-layout: fixed;
+      tr {
+        th.react-aria-CalendarHeaderCell {
+          text-transform: uppercase;
+        }
+
+        td:first-child div.react-aria-CalendarCell {
+          padding-left: 0.25rem;
```
dmortal, line 82: we should not use `rems` directly since we don't know what the root font size is in partner application. `toRem` uses detected pixel value of the root font size to calculate correct rem value

src/components/Company/PaySchedule/_parts/Edit.tsx
jeffredodd: Implemented useWatch

src/components/Common/CalendarDisplay/CalendarDisplay.tsx
jeffredodd: Memoized a few of these functions by outputting arrays with color mappings instead.

src/components/Common/CalendarDisplay/CalendarDisplay.tsx
jeffredodd: Memoized a few here

src/components/Common/CalendarDisplay/CalendarDisplayLegend.tsx
jeffredodd: Good catch, I drafted this way and thought I got them all but forgot to check the legend file.

src/components/Common/CalendarDisplay/CalendarDisplayLegend.tsx
jeffredodd: Thanks!

src/components/Common/CalendarDisplay/CalendarDisplayLegend.tsx
jeffredodd: Thanks

src/components/Company/PaySchedule/_parts/Edit.tsx
jeffredodd: Started a date formatting helper to capture these

src/components/Company/PaySchedule/_parts/Edit.tsx
jeffredodd: Looks like we're getting null strings back from the API here. Handled this via casting as it looks like we never get a null back for this value.

src/components/Company/PaySchedule/_parts/Edit.tsx
jeffredodd: Updated

src/contexts/ThemeProvider/createTheme.ts
jeffredodd: Spoke with @aaronlee777 and decided on coding this as an orange color for now as it doesn't align with warning or error yet.

src/styles/_DatePicker.scss
jeffredodd: Made updates, today I learned here. Thanks!

dmortal (approved):

PR #187, opened 2025-03-11
style: add skeleton loading to component loader UI

aaronlee777:
https://github.com/user-attachments/assets/e4581f90-2640-4a73-ab91-a1a16a4347e1


src/components/Common/Loading/Loading.tsx
```diff
@@ -1,10 +1,10 @@
-import { useTranslation } from 'react-i18next'
+import clsx from 'clsx'
+import styles from './Loading.module.scss'
 
 export const Loading = () => {
-  const { t } = useTranslation('common')
   return (
-    <section aria-live="polite" aria-busy>
-      {t('status.loading')}
```
dmortal, line 7: We can't remove text from this. It can be set through `aria-label` or something, or made invisible, but it has to be there for accessibility purposes

src/components/Common/Loading/Loading.tsx
```diff
@@ -1,10 +1,10 @@
...
-    </section>
+    <div className={styles.skeletonContainer} aria-live="polite" aria-busy>
```
dmortal, line 6: I think `section` is more semantic in this case

src/components/Common/Loading/Loading.tsx
```diff
@@ -1,10 +1,10 @@
-import { useTranslation } from 'react-i18next'
+import clsx from 'clsx'
```
dmortal, line 1: what's `clsx`? we already have `classnames` as a dependency

dmortal (changes requested):

src/components/Common/Loading/Loading.module.scss
```diff
@@ -0,0 +1,32 @@
...
+    var(--g-colors-gray-400) 50%,
+    var(--g-colors-gray-300) 75%
+  );
+  background-size: 200% 100%;
+  animation: pulse 1.5s infinite linear;
+  border-radius: 4px;
+}
+
+.skeletonBox {
+  height: 200px;
```
serikjensen, line 21: Nit: use of `toRem` here instead of px?

serikjensen (approved):
Looks good! will let @dmortal sign off since he left the original feedback

dmortal (approved):

PR #188, opened 2025-03-12
feat: add company federal taxes form

serikjensen:
This adds a company FederalTaxes component. See figma here: https://www.figma.com/design/6dxOSiONDiJoa9zY1wOwbs/Frontend-SDK-Partner-Design-File?node-id=5220-17845&m=dev

Notes
* The figma also has industry selection combined on the same page as federal taxes. We are going to create a separate, combined component for that in a dedicated ticket. For now this just takes on the federal taxes in isolation.
* It doesn't appear that there is a create federal taxes endpoint, only update. [See here](https://docs.gusto.com/embedded-payroll/reference/put-v1-companies-company_id-federal_tax_details). Assumption is that federal taxes must always exist for a given company and we are always just updating here. Lmk if that is incorrect!

## Proof of functionality

https://github.com/user-attachments/assets/62e43ff8-f348-40fa-894b-39082252ae94


src/components/Company/FederalTaxes/Form.tsx
```diff
@@ -0,0 +1,103 @@
...
+import { usePlaceholderEin, normalizeEin } from '@/helpers/federalEin'
+
+export function Form() {
+  const { t } = useTranslation('Company.FederalTaxes')
+  const { federalTaxDetails } = useFederalTaxes()
+  const { control, setValue } = useFormContext<FederalTaxFormInputs>()
+
+  const placeholderEin = usePlaceholderEin(federalTaxDetails?.hasEin)
+
+  const taxPayerTypeOptions = Object.values(TaxPayerType).map(value => ({
```
jeffredodd, line 20: Is there optimization in memoizing them so they are not computed every time the component re-renders/diffs?

src/components/Company/FederalTaxes/Form.tsx
```diff
@@ -0,0 +1,103 @@
...
+    id: value,
+    name: t(`taxPayerType.${value}`),
+  }))
+
+  const filingFormOptions = Object.values(FilingForm).map(value => ({
```
jeffredodd, line 25: Same comment as above

src/components/Company/FederalTaxes/FederalTaxes.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const { onEvent, baseSubmitHandler } = useBase()
+
+  const {
+    data: { federalTaxDetails },
+  } = useFederalTaxDetailsGetSuspense({ companyId })
+  const { mutateAsync: updateFederalTaxDetails, isPending } = useFederalTaxDetailsUpdateMutation()
+
+  const formMethods = useForm<FederalTaxFormInputs>({
+    resolver: valibotResolver(FederalTaxFormSchema),
+    defaultValues: {
```
jeffredodd, line 76: Just curious, do we want to allow the consumer to provide defaultValues to override this ever? Looks like we're just putting the initial data in here which could be fine just doesn't sound like `defaultValues` :)

src/components/Company/FederalTaxes/Form.tsx
serikjensen: Updated to memoize 👍

src/components/Company/FederalTaxes/FederalTaxes.tsx
serikjensen: Good callout! relevant thread just posted by @dmortal here in slack https://gustohq.slack.com/archives/C071WFJUNF5/p1741976496063599

I'm going to keep an eye on that and, based on the outcome, can ticket updating this form with default values as needed

serikjensen: Sep ticket here to add default values https://gustohq.atlassian.net/browse/GWS-4214

src/helpers/federalEin.ts
```diff
@@ -0,0 +1,19 @@
+import { useTranslation } from 'react-i18next'
+
+export const normalizeEin = (value: string) =>
```
jeffredodd, line 3: This is just a light nit.... maybe we need like a comment/doc above this function to explain what it is and why/how it does what it does. Just reading the chaining of match,slice,filter,join,substring made my head go 🤯

jeffredodd (approved):

src/helpers/federalEin.ts
serikjensen: Agreed, updated!

jeffredodd: Thank you! Future developers will be thankful

PR #190, opened 2025-03-13
feat: add property to optionally disallow self onboarding

serikjensen:
This adds a `isSelfOnboardingEnabled` prop to both the Profile and the EmployeeOnboardingFlow. This prop is true by default, but when set to false it sets self_onboarding to false and hides the checkbox for self onboarding.

Admins with this prop set to false will go through the onboarding flow as they normally would when self onboarding is not enabled and will onboard the full employee.

### When isSelfOnboardingEnabled is set to false
* Observe self onboarding checkbox is hidden 
* Observe successfully navigating from the profile to the next step and validations pass

https://github.com/user-attachments/assets/b520263d-6a77-4b40-9db7-07171dd291be

### When isSelfOnboardingEnabled is set to true (default)
* Observe self onboarding checkbox is present
* Observe successful navigation from the profile to the next step and validation passes

https://github.com/user-attachments/assets/7721aa97-256e-4360-b7c2-acaca12b8407


jeffredodd (approved):
LGTM

dmortal (approved):

PR #195, opened 2025-03-14
feat: add flsa_status to compensation default values

serikjensen:
This adds flsa_status to compensation default values.

This was requested by BQE here https://gustohq.slack.com/archives/C0859LP91N1/p1741955135453469

## Proof of functionality

https://github.com/user-attachments/assets/282e86d3-a379-4945-82f7-92233a1c7c2e


src/components/Employee/Compensation/Compensation.tsx
```diff
@@ -31,9 +36,13 @@ import {
...
 
-export type CompensationDefaultValues = RequireAtLeastOne<
-  Partial<Pick<Schemas['Job'], 'rate' | 'title' | 'payment_unit'>>
->
+export type CompensationDefaultValues = RequireAtLeastOne<{
+  rate?: Schemas['Job']['rate']
+  title?: Schemas['Job']['title']
+  payment_unit?: (typeof PAY_PERIODS)[keyof typeof PAY_PERIODS]
+  flsa_status?: Schemas['Compensation']['flsa_status']
+}>
```
serikjensen, line 44: Picking compensation values how we were doing previously was actually not correctly limiting the default values. Ex see

Screenshot 2025-03-14 at 4 39 16 PM

Not sure the best approach beyond picking, but this intentionally enumerates what's available instead

src/components/Employee/Compensation/Compensation.tsx
```diff
@@ -31,9 +36,13 @@ import {
 } from '@/api/queries'
 import { RequireAtLeastOne } from '@/types/Helpers'
 
-export type CompensationDefaultValues = RequireAtLeastOne<
-  Partial<Pick<Schemas['Job'], 'rate' | 'title' | 'payment_unit'>>
->
+export type CompensationDefaultValues = RequireAtLeastOne<{
+  rate?: Schemas['Job']['rate']
+  title?: Schemas['Job']['title']
+  payment_unit?: (typeof PAY_PERIODS)[keyof typeof PAY_PERIODS]
```
serikjensen, line 42: Update to limit payment_unit default value to what we actually have available from pre selected options so when a default value gets supplied this doesn't get out of sync with our infrastructure

PR #209, opened 2025-03-24
Refactor DocumentSignerFlow to use Speakeasy

scally:


jeffredodd (approved):

src/components/Company/DocumentSignerFlow/DocumentList/ManageSignatories.tsx
```diff
@@ -22,8 +22,8 @@ function ManageSignatories() {
   } else if (signatory) {
     signatorySubtext = t('otherSignatorySubtext', {
       signatory: firstLastName({
-        first_name: signatory.first_name,
-        last_name: signatory.last_name,
+        first_name: signatory.firstName,
```
dmortal, line 25: can we avoid keeping snake case in this helper? Some of those helper methods use temporary conditionals (getStreet) that will be removed in cleanup after transition

dmortal (changes requested):

src/components/Company/DocumentSignerFlow/DocumentList/ManageSignatories.tsx
scally: I don't know if I understand what you want here yet. Could you please elaborate?

dmortal (approved):

PR #212, opened 2025-03-26
Da/speakeasy docs

dmortal:
Converting employee document signer to use speakeasy
converting it to use state machine for internal state management
converting employee self-onboarding flow to use nested state machine


src/components/Employee/DocumentSigner/SignatureForm/SignatureForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+    },
+  })
+
+  const handleBack = () => {
+    onEvent(componentEvents.CANCEL)
+  }
+
+  const handleContinue = () => {
+    onEvent(componentEvents.EMPLOYEE_FORMS_DONE)
+  }
```
serikjensen, line 74: Looks like this is leftover from the previous implementation and can be removed? I realized this was firing the same event as DocumentList but actually this one doesn't look like it's called

src/components/Flow/Flow.tsx
```diff
@@ -21,14 +21,20 @@ export interface FlowContextInterface {
...
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
-    send({ type: type, payload: data })
+    //When dealing with nested state machine, correct machine needs to recieve an event
+    if (service.child) {
+      //@ts-expect-error: not sure why 'type' type is incorrectly derived here
+      service.child.send({ type: type, payload: data })
+    } else {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      send({ type: type, payload: data })
+    }
```
serikjensen, line 37: Thanks for figuring this out! will this cover us for more than two levels deep?

src/components/Flow/StateMachines/employeeSelfOnboarding.ts
```diff
@@ -54,14 +67,15 @@ export const employeeSelfOnboardingMachine = {
...
         title: SDKI18next.t('flows.employeeSelfOnboarding.documentSignerTitle'),
       })),
     ),
   ),
-  employeeDocumentSigner: state(
+  //Invoking nested state machine
+  employeeDocumentSigner: invoke(
+    documentSigner,
     transition(
-      componentEvents.EMPLOYEE_FORMS_DONE,
```
serikjensen, line 64: Nit: any reason not to use the constant here? Or do we need a distinct constant instead of hard coding a string value here?

src/components/Employee/DocumentSigner/SignatureForm/SignatureForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+      >
+        {children ? (
+          children
+        ) : (
+          <Flex flexDirection="column">
+            <Head />
+            <FormComponent />
+          </Flex>
+        )}
+      </SignatureFormProvider>
```
serikjensen, line 115: Nit: if you'd like this to be structured with

```
Head
Form
Actions
```

with the form and actions separated, you can follow the approach taken over in the company one here https://github.com/Gusto/embedded-react-sdk/blob/ee8e1635870f7f8ef2def3507f1c3deee51c3a1a/src/components/Company/DocumentSignerFlow/SignatureForm/SignatureForm.tsx#L113-L135

Basically you can move the SharedSignatureForm component up here to this component and then it will wrap all the elements in a form at which point you can pull the actions out of the form and place them separately

src/components/Employee/DocumentSigner/SignatureForm/SignatureForm.tsx
```diff
@@ -0,0 +1,121 @@
...
+        }}
+      >
+        {children ? (
+          children
+        ) : (
+          <Flex flexDirection="column">
+            <Head />
+            <FormComponent />
+          </Flex>
+        )}
```
serikjensen, line 114: Nit: should the Flex component be wrapping both the subcomponents and the children? ex

```suggestion
        <Flex flexDirection="column">
        {children ? (
          children
        ) : (
          <>
            <Head />
            <FormComponent />
          </>
        )}
        </Flex>
```

serikjensen (approved):
Looks great! some small nits and questions but nothing blocking

src/components/Flow/Flow.tsx
dmortal: it should, yes

src/components/Flow/StateMachines/employeeSelfOnboarding.ts
dmortal: This is Robot specific event that is thrown when machine transitions to the final state. I can pull it into a constant

src/components/Employee/DocumentSigner/SignatureForm/SignatureForm.tsx
dmortal: good question - I have pros and cons for both approaches. Arguably, if I choose to customize rendering of children components I might want to control their layout differently. But I think for now, I'll adjust to match what you did - if only because we do not provide access to flex component so if they would want to replace 1 child and keep layout, they would have to re-engineer it

PR #213, opened 2025-03-26
fix: update the base error handling to convert to camelcase

jeffredodd:
Initial pass at getting errors working via the fieldErrors list from `Base.tsx`


src/components/Common/Inputs/DatePicker.tsx
```diff
@@ -59,10 +59,11 @@ export function DatePicker<C extends FieldValues, N extends FieldPath<C>>({
   const { t } = useTranslation()
   const {
     field,
-    fieldState: { invalid },
+    fieldState: { invalid, error },
   } = useController({ name, control })
 
   const dateInputContainerRef = useRef<HTMLDivElement | null>(null)
+  const errorText = error?.message ? error.message : errorMessage
```
jeffredodd, line 66: I'm going to clean this up when we do the RHF separation, but noticed that datepicker wasn't displaying the right errors.

src/components/Company/PaySchedule/PaySchedule.tsx
```diff
@@ -151,7 +151,15 @@ const Root = ({ companyId, children, defaultValues }: PayScheduleProps) => {
     resolver: valibotResolver(PayScheduleSchema),
     defaultValues: transformedDefaultValues,
   })
-  const { watch, setValue, reset } = formMethods
+  const { watch, setValue, reset, setError } = formMethods
+
+  useEffect(() => {
```
jeffredodd, line 156: Hooked up the errors to PaySchedule

src/components/Company/PaySchedule/PaySchedule.tsx
```diff
@@ -107,7 +107,7 @@ export const PaySchedule = ({
 
 const Root = ({ companyId, children, defaultValues }: PayScheduleProps) => {
   const queryClient = useQueryClient()
-  const { baseSubmitHandler, onEvent } = useBase()
+  const { baseSubmitHandler, onEvent, fieldErrors } = useBase()
```
dmortal, line 110: just making sure, `fieldErrors` are currently not populated when speakeasy throws an error - is that being handled in separate PR?

dmortal (approved):

src/components/Company/PaySchedule/PaySchedule.tsx
jeffredodd: If you're talking about the issue I called out earlier this week regarding react-query not letting errors through, I was having issues with the error prop directly from react query. So the following doesn't contain errors, but its not blocking me for now. The fieldErrors in this case are showing up. 
`const { data, errors, isLoading } = usePaySchedulesGetPreview(...)` 

Also just confirmed, both valibot and api errors are appearing here.

PR #219, opened 2025-04-01
feat: add infrastructure for decoupling React Hook Form from inputs

serikjensen:
This PR creates the infrastructure for the work to decouple React Hook Form from our form inputs. It does the following:
* Creates a new FieldLayout component which can be reused for arranging label, description, and error text along with an input
* Creates a new TextInput component that is separated from React Hook Form dependencies
* Creates a new useField hook that allows us to easily access React Hook Form context and map React Hook Form values onto common input props
* Implements a TextInputField component that creates a React Hook Form connected version of the TextInput

## Proof of functionality

You can inspect the new components on ladle

https://github.com/user-attachments/assets/74e7872b-b5ae-4460-95d3-c48d5f4769ad

### Integration with profile form
There will be follow up work to integrate this new TextInputField component with our existing forms and that is beyond the scope of this PR and is not included. However, to ensure this behaves as expected I did some quick working ahead to add this to profile for proof of functionality to ensure everything was working as expected in an existing form. That can be seen here:

https://github.com/user-attachments/assets/9dfa4e0a-3ae0-4d52-be24-c8fd77dba863

And here:

https://github.com/user-attachments/assets/97d40057-3faf-4809-a461-23ee96d3cd4f

### DIFF
--- a/src/components/Common/Fields/TextInputField/TextInputField.tsx
+++ b/src/components/Common/Fields/TextInputField/TextInputField.tsx
@@ -1,0 +1,30 @@
+import type { ChangeEvent } from 'react'
+import { useField, type UseFieldProps } from '@/components/Common/Fields/hooks/useField'
+import { TextInput, type TextInputProps } from '@/components/Common/UI/TextInput'
+
+interface TextInputFieldProps
+  extends Omit<TextInputProps, 'name'>,
+    UseFieldProps<ChangeEvent<HTMLInputElement>> {}
+
+export const TextInputField: React.FC<TextInputFieldProps> = ({
+  rules,
+  defaultValue,
+  name,
+  errorMessage,
+  isRequired,
+  onChange,
+  transform,
+  ...textInputProps
+}: TextInputFieldProps) => {
+  const fieldProps = useField({
+    name,
+    rules,
+    defaultValue,
+    errorMessage,
+    isRequired,
+    onChange,
+    transform,
+  })
+
+  return <TextInput {...fieldProps} {...textInputProps} />
+}
--- a/src/components/Common/Fields/hooks/useField.ts
+++ b/src/components/Common/Fields/hooks/useField.ts
@@ -1,0 +1,54 @@
+import type { RegisterOptions } from 'react-hook-form'
+import { useController, useFormContext } from 'react-hook-form'
+
+export type Transform<TEvent, TValue> = (event: TEvent) => TValue
+
+export interface UseFieldProps<TEvent, TValue = string> {
+  name: string
+  rules?: RegisterOptions
+  defaultValue?: TValue
+  errorMessage?: string
+  isRequired?: boolean
+  onChange?: (event: TEvent) => void
+  transform?: Transform<TEvent, TValue>
+}
+
+export function useField<TEvent, TValue = string>({
+  name,
+  rules = {},
+  defaultValue,
+  errorMessage,
+  isRequired = false,
+  onChange,
+  transform,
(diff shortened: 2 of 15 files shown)


src/components/Common/Fields/TextField/TextField.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const TextField: React.FC<TextFieldProps> = ({
+  rules,
+  defaultValue,
+  name,
+  errorMessage,
+  ...props
+}: TextFieldProps) => {
+  const { control } = useFormContext()
+  const { field, fieldState } = useController({
```
jeffredodd, line 23: Can we somehow get the controller code into a reusable component? We can chat more about this too.

src/components/Common/UI/FormField/FormField.tsx
```diff
@@ -0,0 +1,69 @@
...
+  errorMessage,
+  errorMessageId,
+  children,
+  isRequired = false,
+  htmlFor,
+  className,
+}: FormFieldProps) => {
+  const { t } = useTranslation('common')
+
+  return (
```
jeffredodd, line 40: Yeah I think this makes sense as a rendering component the way you set it up so far.

src/components/Common/Fields/TextField/TextField.tsx
jeffredodd: Feels like HookField like I sorta had earlier might be able to handle this? But I like this separation overall across all the fields.

src/components/Common/Fields/TextField/index.ts
```diff
@@ -0,0 +1 @@
+export { TextField } from './TextField'
```
jeffredodd, line 1: This is high level but I think we're going to want to get away from index.ts for things we want to tree shake away. In this case its probably fine but for the `UI` folder we might not want it to encourage direct imports which can be shaken away.

jeffredodd (commented):
Overall I think this is a good direction and i like where we are headed with this.

src/components/Common/Fields/TextField/TextField.tsx
serikjensen: I updated to try and place some of the common mappings in a hook. The idea would be to effectively just abstract away the direct calls to hook form and return values that were readily mapped to props in our namespace

src/components/Common/Fields/hooks/useField.ts
```diff
@@ -0,0 +1,54 @@
...
+      required: isRequired,
+      ...rules,
+    },
+    defaultValue,
+  })
+
+  const { ref, ...restFieldProps } = field
+
+  const handleChange = (event: TEvent) => {
+    const value = transform ? transform(event) : event
```
serikjensen, line 39: This will allow us to do things like formatting the ssn and phone number without needing to manually set the value like we've been doing

src/components/Common/Fields/TextInputField/TextInputField.tsx
```diff
@@ -0,0 +1,30 @@
...
+    name,
+    rules,
+    defaultValue,
+    errorMessage,
+    isRequired,
+    onChange,
+    transform,
+  })
+
+  return <TextInput {...fieldProps} {...textInputProps} />
```
serikjensen, line 29: This is where we would supply the adapter. TextInput is now cleanly separated from all hook form dependencies

jeffredodd (approved):
Sweet! This is such a great start! 🎉 🚢

PR #220, opened 2025-04-03
fix: support nested error list returned from speakeasy client

jeffredodd:
This change does the following
- Enables error list to come from OpenAPI (Meta data needed to be updated)
- Adds error_key prefix to nested error list logic to create valid nested keys
- Introduced camelCaseToSnakeCase which will be used to convert from/to various cases


src/components/Base/Base.tsx
```diff
@@ -74,23 +75,25 @@ const getFieldErrors = (
...
-        key: keyPrefix + error.errorKey,
+        key: snakeCaseToCamelCase(keyPrefix + error.errorKey),
         message: error.message ?? '',
       },
     ]
   }
   if (error.category === 'nested_errors' && error.errors !== undefined) {
     //TODO: clean this up once Metadata type is fixed in openapi spec
-    const keySuffix =
+    let keySuffix = ''
```
jeffredodd, line 85: Broke this out from a complicated nested ternary to a if/else function as its a bit more clear to read. I had to add another logical branch so figured this made sense.

serikjensen (approved):
LGTM!

PR #221, opened 2025-04-03
chore: cleaning export namespaces

dmortal:
Proposed cleanup:
- Changing index file exports to be a bit more deliberate: making sure that partners only see exports that are explicitly meant for them to consume
- Main change is in bottom 3 index files
As a result, partners don't have access to things like `ThemeProvider` or our 'contextual' component wrappers that we use in flows: 
image


scally (approved):
Appreciating the usage of more named exports besides using `*`, it should help with our HMR issues eventually :)

PR #222, opened 2025-04-03
Employee Onboarding E2E test

scally:
Add end to end test for Employee Onboarding flow.

Not comprehensive; follows the simplest happy path between adding an employee and seeing their onboarding summary at the end.


src/test/mocks/apis/employee_home_addresses.ts
```diff
@@ -42,7 +42,7 @@ export const createEmployeeHomeAddress = http.post<
   EmployeeAddress$Outbound
 >(`${API_BASE_URL}/v1/employees/:employee_id/home_addresses`, async () => {
   const responseFixture = await getFixture('get-v1-home_addresses-home_address_uuid')
-  return HttpResponse.json(responseFixture)
+  return HttpResponse.json(responseFixture, { status: 201 })
```
scally, line 45: Many mocks needed changing to match Speakeasy's expectation of a 201 HTTP status for created objects.

src/test/mocks/apis/employees.ts
```diff
@@ -31,20 +31,24 @@ export function handleGetCompanyEmployees(
...
-  HttpResponse.json([
-    {
-      uuid: 'some-unique-id',
-      first_name: 'Maximus',
-      last_name: 'Steel',
-      payment_method: 'Direct Deposit',
-    },
-  ]),
-)
+export const getCompanyEmployees = (companyId?: string) =>
```
scally, line 39: Made this generalizable so a companyId could be provided. This way the knowledge of the "correct" company ID could move out of the mocks and into the tests for simpler reading of the tests.

src/test/mocks/fixtures/get-v1-employees-employee_id-jobs.json
```diff
@@ -7,7 +7,7 @@
     "payment_unit": "Year",
     "primary": true,
     "title": "Client Support Director",
-    "state_wc_covered": "null,",
+    "state_wc_covered": null,
```
scally, line 10: Fixes a Speakeasy validation that expects this field to be null but not a string.

src/components/Flow/EmployeeOnboardingFlow.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+        getEmployeeStateTaxes,
+        updateEmployeeStateTaxes,
+        getEmptyEmployeePaymentMethod,
+        getEmptyEmployeeBankAccounts,
+        updateEmptyEmployeePaymentMethod,
+        getEmployeeGarnishments,
+        updateEmployeeCompensation,
+        getEmptyEmployeeForms,
+        getEmployeeOnboardingStatus,
+        updateEmployeeOnboardingStatus,
```
scally, line 83: sweet fancy moses that's a lot of API call dependencies... 😓

src/components/Flow/EmployeeOnboardingFlow.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      )
+    })
+
+    it('succeeds', { timeout: 10000 }, async () => {
```
scally, line 87: 10s timeout. It runs around 3sec locally but CI is slower :(

src/components/Flow/EmployeeOnboardingFlow.test.tsx
```diff
@@ -0,0 +1,149 @@
...
+      await user.type(await screen.findByLabelText(/last name/i), 'silver')
+      await user.type(await screen.findByLabelText(/email/i), 'someone@definitely-not-gusto.com')
+
+      // Work address
+      await user.click(await screen.findByLabelText(/work address/i))
+      await user.click(await screen.findByText(/123 Main St/i))
+      await fillDate({ date: { month: 1, day: 1, year: 2025 }, name: 'Start date', user })
+      await fillDate({ date: { month: 1, day: 1, year: 2000 }, name: 'Date of birth', user })
+      await user.type(await screen.findByLabelText('Street 1'), '123 Any St')
+      await user.type(await screen.findByLabelText(/city/i), 'Redmond')
```
serikjensen, line 110: I didn't have time to continue debugging, but were we able to try taking the approach of

```suggestion
      // Page - Personal Details
      await user.type(await screen.findByLabelText(/social/i), '456789012')
      await user.type(await screen.getByLabelText(/first name/i), 'john')
      await user.type(await screen.getByLabelText(/last name/i), 'silver')
      await user.type(await screen.getByLabelText(/email/i), 'someone@definitely-not-gusto.com')

      // Work address
      await user.click(await screen.findByLabelText(/work address/i))
      await user.click(await screen.getByText(/123 Main St/i))
      await fillDate({ date: { month: 1, day: 1, year: 2025 }, name: 'Start date', user })
      await fillDate({ date: { month: 1, day: 1, year: 2000 }, name: 'Date of birth', user })
      await user.type(await screen.getByLabelText('Street 1'), '123 Any St')
      await user.type(await screen.getByLabelText(/city/i), 'Redmond')
```

Essentially the first assertion uses an async find and then once the page has loaded we can just use `get` for everything else on this page (until the next transition) since the rest of the page content is loaded? I was worried that use of only async find assertions here was leading to some of the oddities we were observing when debugging the tests

serikjensen (approved):
Looks good! Question around the use of find assertions and if we ended up figuring out why we were seeing the oddities we were seeing when debugging this test last week

src/components/Flow/EmployeeOnboardingFlow.test.tsx
scally: There's not really anything to debug here, as right now we have passing tests. I don't know that there's evidence yet to support these hypotheses and I have yet to find any reason to believe that using `find` is problematic. That said, if you have evidence that we should be using a different approach here I'm willing to take a look into it.

PR #224, opened 2025-04-04
[WIP] feat: add updated number input

serikjensen:
[WIP]


src/components/Common/UI/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+        isInvalid={isInvalid}
+        isDisabled={isDisabled}
+        isRequired={isRequired}
+        validationBehavior="aria"
+        onChange={onChange}
+        // This is a hack to silence an unnecessary react-aria warning. The FieldLayout component
+        // already associates the label and input with htmlFor + ID. If we include a label id here,
+        // the label will get read twice by assistive tech. This evaluates to an empty string which
+        // does not get associated with any elements and does not get read.
+        aria-labelledby=" "
```
dmortal, line 82: would `aria-hidden` be more appropriate?

serikjensen: honestly i wish we could just leave this out altogether 😔 it's solely to satisfy a react aria warning that's impossible to silence otherwise. It fires with every onChange which makes it super noisy. We don't actually need it in our case because we are already setting a label via the FieldLayout component which is properly associated with the input but there's no way to communicate that to react aria

the end result of this is just an empty `aria-labelledby` attribute on the outputted dom element, but per my testing with screen readers it doesn't seem to get read. The alternative i tried was just associating an id for the label using the aria-labelledby redundantly (since it's already associated with htmlFor + id) but voice over was then reading the label twice

Provided a comment there that this is a temp hack. open to other options if folks can figure them out!

src/components/Common/Fields/NumberInputField/NumberInputField.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+  children: React.ReactNode
+  defaultValues?: {
+    test?: number
+    amount?: number
+    quantity?: number
+    percentage?: number
+  }
+}
+
+const FormWrapper = ({ children, defaultValues = {} }: FormWrapperProps) => {
```
jeffredodd, line 16: Should we maybe put this somewhere modular if we're just using it for Ladle testing?

src/components/Common/UI/NumberInput/NumberInput.tsx
```diff
@@ -0,0 +1,101 @@
...
+  NumberField as AriaNumberField,
+} from 'react-aria-components'
+import { FieldLayout, type FieldLayoutProps } from '../FieldLayout'
+import { useLocale } from '@/contexts/LocaleProvider'
+
+export interface NumberInputProps extends FieldLayoutProps {
+  name?: string
+  format?: 'currency' | 'decimal' | 'percent'
+  currencyDisplay?: keyof Intl.NumberFormatOptionsCurrencyDisplayRegistry | undefined
+  minValue?: number
```
jeffredodd, line 16: We don't have to do this right now, but we've chatted before about picking the props from the React HTML Type for Number Inputs as an example, should we be thinking about that?

src/components/Common/UI/NumberInput/NumberInput.tsx
jeffredodd: More fuel for why React-aria isn't fitting well with us.

src/components/Common/Fields/NumberInputField/NumberInputField.tsx
```diff
@@ -0,0 +1,39 @@
+import { useField, type UseFieldProps } from '@/components/Common/Fields/hooks/useField'
+import { NumberInput, type NumberInputProps } from '@/components/Common/UI/NumberInput'
+
+interface NumberInputFieldProps
+  extends Omit<NumberInputProps, 'name' | 'value'>,
+    UseFieldProps<number, number> {}
+
+export const NumberInputField: React.FC<NumberInputFieldProps> = ({
```
jeffredodd, line 8: I love how simple this file is looking with the abstractions we've made.

jeffredodd (commented):
Small feedback

src/components/Common/Fields/NumberInputField/NumberInputField.stories.tsx
serikjensen: Yeah maybe we can consolidate in one of the follow up tickets for hook form separation? I can look to update it to a single place when doing checkbox group

src/components/Common/UI/NumberInput/NumberInput.tsx
serikjensen: Yeah good point, maybe the best way forward is to pick from the html prop types. let me do some experimentation with it

jeffredodd (approved):
LGTM

PR #227, opened 2025-04-07
feat: add updated checkbox component

serikjensen:
This PR
* Creates a dedicated Checkbox component separated from React Hook Form
* Creates a CheckboxField component that is hooked up to React Hook Form
* Creates a HorizontalFieldLayout component to handle layout concerns for checkbox, switch, and radio
* Breaks out the common description and error message ui pieces and centralizes them so they are shared between field layout components
* Centralizes the id logic for associating label, error message, and description so it is consistent across components

## Proof of functionality
Examples of all components in ladle

https://github.com/user-attachments/assets/aa7fc293-5177-4fbf-8c03-1dd0f0f01984

https://github.com/user-attachments/assets/4aa44da1-1122-4179-9459-b01f04938d8b

### DIFF
--- a/src/components/Common/Fields/CheckboxField/CheckboxField.stories.tsx
+++ b/src/components/Common/Fields/CheckboxField/CheckboxField.stories.tsx
@@ -1,0 +1,66 @@
+import type { Story } from '@ladle/react'
+import { FormProvider, useForm } from 'react-hook-form'
+import { CheckboxField } from './CheckboxField'
+
+interface FormWrapperProps {
+  children: React.ReactNode
+  defaultValues?: {
+    acceptTerms?: boolean
+    subscribeNewsletter?: boolean
+    enableNotifications?: boolean
+  }
+}
+
+const FormWrapper = ({ children, defaultValues = {} }: FormWrapperProps) => {
+  const methods = useForm({
+    defaultValues: {
+      acceptTerms: defaultValues.acceptTerms ?? false,
+      subscribeNewsletter: defaultValues.subscribeNewsletter ?? false,
+      enableNotifications: defaultValues.enableNotifications ?? false,
+    },
+    mode: 'onTouched',
+  })
+
+  return (
+    <FormProvider {...methods}>
+      <form onSubmit={methods.handleSubmit(() => {})}>{children}</form>
+    </FormProvider>
+  )
+}
+
+export const Default: Story = () => (
+  <FormWrapper>
+    <CheckboxField label="Accept Terms and Conditions" name="acceptTerms" />
+  </FormWrapper>
+)
+
+export const Required: Story = () => {
+  return (
+    <FormWrapper>
+      <CheckboxField
+        label="Accept Terms and Conditions"
+        name="acceptTerms"
+        isRequired
+        errorMessage="You must accept the terms"
+      />
+      <br />
+      <button type="submit">Submit</button>
+    </FormWrapper>
+  )
+}
+
+export const WithDefaultValues: Story = () => {
+  return (
+    <FormWrapper
+      defaultValues={{
+        acceptTerms: true,
+        subscribeNewsletter: false,
+        enableNotifications: true,
+      }}
+    >
+      <CheckboxField label="Accept Terms and Conditions" name="acceptTerms" />
+      <CheckboxField label="Subscribe to Newsletter" name="subscribeNewsletter" />
+      <CheckboxField label="Enable Push Notifications" name="enableNotifications" />
+    </FormWrapper>
+  )
+}
--- a/src/components/Common/UI/Checkbox/Checkbox.module.scss
+++ b/src/components/Common/UI/Checkbox/Checkbox.module.scss
@@ -1,0 +1,69 @@
+.checkboxWrapper {
+  width: toRem(20);
+  height: toRem(20);
+  position: relative;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+.checkbox {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
+  height: 100%;
+  border: var(--g-checkbox-borderWidth) solid var(--g-checkbox-borderColor);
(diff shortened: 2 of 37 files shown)


src/components/Common/UI/hooks/useFieldIds.ts
```diff
@@ -0,0 +1,22 @@
...
+  const generatedErrorMessageId = useId()
+
+  const inputId = providedInputId || generatedInputId
+  const errorMessageId = providedErrorMessageId || generatedErrorMessageId
+
+  return {
+    inputId,
+    errorMessageId,
+  }
+}
```
serikjensen, line 22: Hook for centralizing some common logic around id generation

src/components/Common/UI/Checkbox/Checkbox.module.scss
```diff
@@ -0,0 +1,61 @@
...
+
+.checkbox {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  width: 100%;
+  height: 100%;
+  border: var(--g-checkbox-borderWidth) solid var(--g-checkbox-borderColor);
+  border-radius: var(--g-checkbox-radius);
+  transition: all 200ms;
```
dmortal, line 18: nit: we should probably have transition timings in global var or mixin for consistency

dmortal: possibly in theme?

src/components/Common/UI/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,86 @@
...
+  type SharedHorizontalFieldLayoutProps,
+} from '../HorizontalFieldLayout'
+import styles from './Checkbox.module.scss'
+import IconChecked from '@/assets/icons/checkbox.svg?react'
+
+export interface CheckboxProps
+  extends SharedHorizontalFieldLayoutProps,
+    Pick<
+      InputHTMLAttributes<HTMLInputElement>,
+      'name' | 'id' | 'className' | 'onChange' | 'onBlur' | 'checked' | 'value'
```
dmortal, line 15: looks like we need top level type with common props like `className`?

src/components/Common/UI/hooks/useFieldIds.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function useFieldIds({
+  inputId: providedInputId,
+  errorMessage,
+  errorMessageId: providedErrorMessageId,
+  description,
+  descriptionId: providedDescriptionId,
+}: UseFieldIdsProps) {
+  const generatedInputId = useId()
```
dmortal, line 18: I wonder if all of the ids of a field and it's components should use prefixes instead of separate hashes for easier debugging? https://react.dev/reference/react/useId#generating-ids-for-several-related-elements

dmortal (approved):

src/components/Common/UI/hooks/useFieldIds.ts
serikjensen: Agreed! will update!

src/components/Common/UI/Checkbox/Checkbox.module.scss
serikjensen: yeah +1 i can get a var going

src/components/Common/UI/Checkbox/Checkbox.tsx
serikjensen: yeah my thought here is that we can just implement what each component needs for now and then we can see what can be centralized into a common interface after we roll out the adapter pattern. it'll be easier to see the things in common once we have them all implemented

src/components/Common/Fields/CheckboxField/CheckboxField.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { CheckboxField } from './CheckboxField'
+
+interface FormWrapperProps {
+  children: React.ReactNode
+  defaultValues?: {
+    acceptTerms?: boolean
+    subscribeNewsletter?: boolean
+    enableNotifications?: boolean
+  }
+}
```
jeffredodd, line 12: Can we get some actions in here so we can validate the field is working from within ladle? Take a look how I did SelectField stories. This was a callout from @scally and I think it makes a big difference for usability.

src/components/Common/Fields/CheckboxField/CheckboxField.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <FormWrapper>
+      <CheckboxField
+        label="Accept Terms and Conditions"
+        name="acceptTerms"
+        isRequired
+        errorMessage="You must accept the terms"
+      />
+      <br />
+      <button type="submit">Submit</button>
```
jeffredodd, line 47: Maybe this submit does what I called out above IRT actions in ladle

src/components/Common/UI/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export const Default: Story = () => {
+  const [checked, setChecked] = useState(false)
+  return (
+    <Checkbox
+      label="Accept terms and conditions"
+      name="terms"
+      checked={checked}
+      onChange={event => {
+        setChecked(event.target.checked)
```
jeffredodd, line 13: I think another opportunity to use ladle actions here for a better demo experience.

jeffredodd (commented):
Just some small comments

src/components/Common/UI/Checkbox/Checkbox.stories.tsx
serikjensen: Yeah i agree! thinking since this has already gotten quite substantial in size i can create a dedicated follow up? that would be a ticket involving unifying the form wrapper and actions for ease of implementing field stories (since we're going to need to keep rolling these out)

PR #228, opened 2025-04-08
feat: select field react hook forms decoupling

jeffredodd:
This PR aims to introduce the following...

- Select component using React-aria-components as stand alone UI
- SelectField component which implements useField and FieldLayout to provide a consistent experience 

## Proof of functionality
https://github.com/user-attachments/assets/4cd2e764-956d-45bc-8fe4-52334596b868

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -101,4 +101,10 @@ export default [
       'no-restricted-imports': 'off',
     },
   },
+  {
+    files: ['src/components/Common/UI/**/*.tsx'],
+    rules: {
+      'no-restricted-imports': 'off',
+    },
+  },
 ]
--- a/src/components/Common/Fields/SelectField/SelectField.stories.tsx
+++ b/src/components/Common/Fields/SelectField/SelectField.stories.tsx
@@ -1,0 +1,202 @@
+import type { Story } from '@ladle/react'
+import { FormProvider, useForm } from 'react-hook-form'
+import { useEffect } from 'react'
+import { SelectField } from './SelectField'
+
+interface FormWrapperProps {
+  children: React.ReactNode
+  defaultValues?: {
+    category?: string
+    priority?: string
+    status?: string
+  }
+}
+
+const FormWrapper = ({ children, defaultValues = {} }: FormWrapperProps) => {
+  const methods = useForm({
+    defaultValues: {
+      category: defaultValues.category || '',
+      priority: defaultValues.priority || '',
+      status: defaultValues.status || '',
+    },
+    mode: 'onTouched',
+  })
+
+  // Log form state changes to console
+  useEffect(() => {
+    const subscription = methods.watch((value, { name, type }) => {
+      // eslint-disable-next-line no-console
+      console.log('Form state changed:', {
+        values: value,
+        name,
+        type,
+        errors: methods.formState.errors,
+        touchedFields: methods.formState.touchedFields,
+        dirtyFields: methods.formState.dirtyFields,
+        isDirty: methods.formState.isDirty,
+        isValid: methods.formState.isValid,
+      })
+    })
+
+    return () => {
+      subscription.unsubscribe()
+    }
+  }, [methods])
+
+  return (
+    <FormProvider {...methods}>
+      <div>{children}</div>
+    </FormProvider>
+  )
+}
+
+const categories = [
+  { id: '1', name: 'Electronics' },
+  { id: '2', name: 'Clothing' },
+  { id: '3', name: 'Books' },
+  { id: '4', name: 'Home & Garden' },
+  { id: '5', name: 'Sports' },
+]
+
+const priorities = [
+  { id: 'low', name: 'Low' },
+  { id: 'medium', name: 'Medium' },
+  { id: 'high', name: 'High' },
+  { id: 'urgent', name: 'Urgent' },
+]
+
+const statuses = [
+  { id: 'new', name: 'New' },
+  { id: 'in-progress', name: 'In Progress' },
+  { id: 'review', name: 'Under Review' },
+  { id: 'completed', name: 'Completed' },
+]
+
+export const Default: Story = () => {
+  return (
+    <FormWrapper>
+      <SelectField
+        name="category"
+        label="Category"
+        items={categories}
+        placeholder="Select a category"
+      >
+        {item => <div>{item.name}</div>}
+      </SelectField>
+      <SelectField
+        name="priority"
+        label="Priority"
+        items={priorities}
+        placeholder="Select a priority"
+      >
+        {item => <div>{item.name}</div>}
+      </SelectField>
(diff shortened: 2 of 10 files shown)


src/components/Common/UI/Select/Select.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+export const WithPlaceholder: Story = () => {
+  return (
+    <Select
+      label="Select an option"
+      name="select"
+      options={options}
+      onChange={handleChange}
+      onBlur={() => {}}
+      inputProps={{ title: 'Choose an option' }}
```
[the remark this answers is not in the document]
jeffredodd: Fixed

src/components/Common/UI/Select/Select.tsx
```diff
@@ -0,0 +1,96 @@
...
+            {({ defaultChildren, isPlaceholder }) => {
+              return isPlaceholder && placeholder ? placeholder : defaultChildren
+            }}
+          </SelectValue>
+          <div aria-hidden="true">
+            <CaretDown title={t('icons.selectArrow')} />
+          </div>
+        </Button>
+        <Popover UNSTABLE_portalContainer={container.current}>
+          <ListBox items={items}>{item => <ListBoxItem>{item.name}</ListBoxItem>}</ListBox>
```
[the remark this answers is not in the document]
jeffredodd: Fixed

src/components/Common/Fields/SelectField/SelectField.stories.tsx
```diff
@@ -0,0 +1,202 @@
...
+  const methods = useForm({
+    defaultValues: {
+      category: defaultValues.category || '',
+      priority: defaultValues.priority || '',
+      status: defaultValues.status || '',
+    },
+    mode: 'onTouched',
+  })
+
+  // Log form state changes to console
```
scally, line 25: You might be able to do a similar thing with ladle actions?
https://ladle.dev/docs/actions

src/components/Common/Fields/SelectField/SelectField.tsx
```diff
@@ -0,0 +1,45 @@
+import { useField, type UseFieldProps } from '@/components/Common/Fields/hooks/useField'
+import { Select } from '@/components/Common/UI/Select'
+
+interface SelectItem {
+  id: string
```
dmortal, line 5: should `id` and `name` be required at the same time? as in, can name be used as id? I'm wondering if there's going to be a lot of duplication `<SelectItem id="password" name="password"/>`?

dmortal: I also wonder if we should allow passthrough of props like `data` and `aria`?

src/components/Common/UI/Select/Select.test.tsx
```diff
@@ -0,0 +1,142 @@
...
+  }),
+}))
+
+// Mock the SVG import
+vi.mock('@/assets/icons/caret-down.svg?react', () => ({
+  default: () => <div data-testid="caret-down" />,
+}))
+
+// Mock the ThemeProvider context
+vi.mock('@/contexts/ThemeProvider', async () => {
```
dmortal, line 20: we might want to have this mock as global 🤔

src/components/Common/UI/Select/Select.tsx
```diff
@@ -0,0 +1,98 @@
...
+import type { FocusEvent, SelectHTMLAttributes } from 'react'
+import { useId, useMemo } from 'react'
+import type { FieldLayoutProps } from '../FieldLayout'
+import { FieldLayout } from '../FieldLayout'
+import CaretDown from '@/assets/icons/caret-down.svg?react'
+import { useTheme } from '@/contexts/ThemeProvider'
+
+export interface SelectProps
+  extends FieldLayoutProps,
+    Pick<SelectHTMLAttributes<HTMLSelectElement>, 'className' | 'id' | 'name'> {
```
dmortal, line 20: `className` probably should be part of `FieldLayoutProps` as I imagine it will be picked for every single field>?

src/components/Common/UI/Select/Select.tsx
```diff
@@ -0,0 +1,98 @@
...
+  onBlur,
+  options,
+  placeholder,
+  value,
+  ...props
+}: SelectProps) => {
+  const { t } = useTranslation()
+  const generatedInputId = useId()
+  const inputId = providedId || generatedInputId
+  const generatedErrorMessageId = useId()
```
dmortal, line 52: I wonder if it would be beneficial in any way to use same generated id for error messages and fields with prefixes - easier debugging perhaps? https://react.dev/reference/react/useId#generating-ids-for-several-related-elements

dmortal (approved):
Few comments, but all of them are mostly rambling

eslint.config.mjs
```diff
@@ -101,4 +101,10 @@ export default [
       'no-restricted-imports': 'off',
     },
   },
+  {
+    files: ['src/components/Common/UI/**/*.tsx'],
+    rules: {
+      'no-restricted-imports': 'off',
```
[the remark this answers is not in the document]
jeffredodd: Good callout. Will be fixed in the next revision.

src/components/Common/Fields/SelectField/SelectField.tsx
jeffredodd: Good callouts, yeah i'll add data and aria if I can get them working with React-Aria. 

I'm going to move SelectItem into the UI select as that seems to make the most sense. 

Added aria-* attributes to the parent element for now.

src/components/Common/UI/Select/Select.tsx
jeffredodd: Good call out

src/components/Common/UI/Select/Select.tsx
jeffredodd: @serikjensen is actually working on making this pattern better. Tagging him so he can see this comment.

src/components/Common/UI/Select/Select.test.tsx
jeffredodd: So the only reason we need it is because of how React-aria is mounting their popover and we're using an unsafe prop... since we want to eventually send React-Aria packing I think I want to keep it local to the UI component using react Aria first before we do anything else with it. But this is a good callout.

src/components/Common/UI/FieldLayout/FieldLayout.tsx
```diff
@@ -34,7 +34,7 @@ export const FieldLayout: React.FC<FieldLayoutProps> = ({
 
   const labelContent = (
     <label className={styles.label} htmlFor={htmlFor}>
-      {label}
+      {label}{' '}
```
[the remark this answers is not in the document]
jeffredodd: With and without screenshots
Screenshot 2025-04-09 at 8 59 42 AM
Screenshot 2025-04-09 at 8 59 28 AM

PR #229, opened 2025-04-09
feat: update UI form components to accept data props

serikjensen:
This updates our UI form components to accept data attributes
* Any `data-` prefixed props are accepted
* data props are forwarded to the root of the field layout component

## Proof of functionality

The following JSX

Screenshot 2025-04-09 at 1 49 50 PM

Results in

Screenshot 2025-04-09 at 1 49 37 PM


scally:
General question here, is there any reason why we wouldn't want to forward any props we don't specifically handle?

dmortal (approved):

PR #230, opened 2025-04-09
feat: Decouple ComboBox from React-Hook-Forms

jeffredodd:
This PR aims to accomplish the following...
- Decouple ComboBox from React Hook Forms
- Introduce a stand alone ComboBox
- Fix styling for legacy ComboBox to align with the styling of Select

## Proof of Functionality

https://github.com/user-attachments/assets/cea06101-77d4-42d2-bc97-46d8df4bd67e

### DIFF
--- a/src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
+++ b/src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
@@ -1,0 +1,43 @@
+import { useField, type UseFieldProps } from '@/components/Common/Fields/hooks/useField'
+import { ComboBox } from '@/components/Common/UI/ComboBox/ComboBox'
+
+interface ComboBoxItem {
+  id: string
+  name: string
+}
+
+interface ComboBoxFieldProps
+  extends Omit<React.ComponentProps<typeof ComboBox>, 'name' | 'onChange' | 'onBlur' | 'options'>,
+    UseFieldProps<string> {
+  items: ComboBoxItem[]
+}
+
+export const ComboBoxField: React.FC<ComboBoxFieldProps> = ({
+  rules,
+  defaultValue,
+  name,
+  errorMessage,
+  isRequired,
+  onChange,
+  transform,
+  items,
+  ...comboBoxProps
+}: ComboBoxFieldProps) => {
+  const fieldProps = useField({
+    name,
+    rules,
+    defaultValue,
+    errorMessage,
+    isRequired,
+    onChange,
+    transform,
+  })
+
+  return (
+    <ComboBox
+      {...fieldProps}
+      options={items.map(item => ({ label: item.name, value: item.id }))}
+      {...comboBoxProps}
+    />
+  )
+}
--- a/src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
+++ b/src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
@@ -1,12 +1,9 @@
 import { useField, type UseFieldProps } from '@/components/Common/Fields/hooks/useField'
-import type { ComboBoxItem } from '@/components/Common/UI/ComboBox/ComboBox'
 import { ComboBox } from '@/components/Common/UI/ComboBox/ComboBox'
 
 interface ComboBoxFieldProps
-  extends Omit<React.ComponentProps<typeof ComboBox>, 'name' | 'onChange' | 'onBlur' | 'options'>,
-    UseFieldProps<string> {
-  options: ComboBoxItem[]
-}
+  extends Omit<React.ComponentProps<typeof ComboBox>, 'name' | 'onChange' | 'onBlur'>,
+    UseFieldProps<string> {}
 
 export const ComboBoxField: React.FC<ComboBoxFieldProps> = ({
   rules,
@@ -16,7 +13,6 @@ export const ComboBoxField: React.FC<ComboBoxFieldProps> = ({
   isRequired,
   onChange,
   transform,
-  options,
   ...comboBoxProps
 }: ComboBoxFieldProps) => {
   const fieldProps = useField({
@@ -29,11 +25,5 @@ export const ComboBoxField: React.FC<ComboBoxFieldProps> = ({
     transform,
   })
 
-  return (
-    <ComboBox
-      {...fieldProps}
-      {...comboBoxProps}
-      options={options.map(o => ({ label: o.name, value: o.id }))}
-    />
-  )
+  return <ComboBox {...fieldProps} {...comboBoxProps} />
 }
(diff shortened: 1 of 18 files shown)


src/components/Common/UI/Select/Select.tsx
```diff
@@ -101,7 +101,7 @@ export const Select = ({
             <CaretDown title={t('icons.selectArrow')} />
           </div>
         </Button>
-        <Popover UNSTABLE_portalContainer={container.current}>
+        <Popover UNSTABLE_portalContainer={container.current} maxHeight={320}>
```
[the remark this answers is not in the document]
jeffredodd: This is a good idea actually, @serikjensen any thoughts here?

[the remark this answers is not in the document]
jeffredodd: FWIW I only added this because it really does jet to the bottom of a page with a big list which I noticed when using states in my combobox story.

src/components/Common/UI/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,102 @@
...
+  onBlur,
+  options,
+  placeholder,
+  value,
+  ...props
+}: ComboBoxProps) => {
+  const { t } = useTranslation()
+  const generatedInputId = useId()
+  const inputId = providedId || generatedInputId
+  const generatedErrorMessageId = useId()
```
serikjensen, line 55: If you rebase you'll now be able to update this to use `useFieldIds` util instead! see other form components in the UI directory for reference

src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
```diff
@@ -0,0 +1,43 @@
...
+    isRequired,
+    onChange,
+    transform,
+  })
+
+  return (
+    <ComboBox
+      {...fieldProps}
+      options={items.map(item => ({ label: item.name, value: item.id }))}
+      {...comboBoxProps}
```
serikjensen, line 40: Could we update this so that ComboBoxField also uses `options` instead of mapping to a separate `items` type? I'm +1 to options as implemented in ComboBox for the consistency with Select and think ComboBoxField would also benefit from using that so it's not a different prop using one over the other

serikjensen (commented):
Just a couple of things!

src/components/Common/UI/Select/Select.tsx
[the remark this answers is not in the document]
serikjensen: I'm fine with putting it in a constant if we feel it'll be reused!

src/components/Common/UI/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,102 @@
...
+  Input,
+  ListBox,
+  ListBoxItem,
+  Popover,
+} from 'react-aria-components'
+import { useTranslation } from 'react-i18next'
+import type { FocusEvent, InputHTMLAttributes } from 'react'
+import { useId } from 'react'
+import classNames from 'classnames'
+import type { FieldLayoutProps } from '../FieldLayout'
```
serikjensen, line 14: You'll also want to update this to use `SharedFieldLayoutProps`

```suggestion
import type { SharedFieldLayoutProps } from '../FieldLayout'
```

that'll restrict it to only the props that are meant to be shared across input components

src/components/Common/UI/ComboBox/ComboBox.tsx
jeffredodd: Updated - WIll be in next revision

src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
jeffredodd: Updated - WIll be in next revision

src/components/Common/UI/ComboBox/ComboBox.tsx
jeffredodd: Updated - WIll be in next revision

src/components/Common/UI/ComboBox/ComboBox.tsx
```diff
@@ -0,0 +1,110 @@
...
+  onChange?: (value: string) => void
+  onBlur?: (e: FocusEvent) => void
+  options: ComboBoxOption[]
+  value?: string
+}
+
+export interface ComboBoxItem {
+  id: string
+  name: string
+}
```
serikjensen, line 40: Looks like we just need to delete this

src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
```diff
@@ -0,0 +1,39 @@
+import { useField, type UseFieldProps } from '@/components/Common/Fields/hooks/useField'
+import type { ComboBoxItem } from '@/components/Common/UI/ComboBox/ComboBox'
```
serikjensen, line 2: Then we can delete this

src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
```diff
@@ -0,0 +1,39 @@
...
+import { ComboBox } from '@/components/Common/UI/ComboBox/ComboBox'
+
+interface ComboBoxFieldProps
+  extends Omit<React.ComponentProps<typeof ComboBox>, 'name' | 'onChange' | 'onBlur' | 'options'>,
```
serikjensen, line 6: Then not omit options

```suggestion
  extends Omit<React.ComponentProps<typeof ComboBox>, 'name' | 'onChange' | 'onBlur'>,
```

src/components/Common/Fields/ComboBoxField/ComboBoxField.tsx
```diff
@@ -0,0 +1,39 @@
...
+    transform,
+  })
+
+  return (
+    <ComboBox
+      {...fieldProps}
+      {...comboBoxProps}
+      options={options.map(o => ({ label: o.name, value: o.id }))}
+    />
+  )
```
serikjensen, line 38: Once we do the above, we should be able to forward options along via comboBoxProps without needing to explicitly pull it out and map!

```suggestion
export const ComboBoxField: React.FC<ComboBoxFieldProps> = ({
  rules,
  defaultValue,
  name,
  errorMessage,
  isRequired,
  onChange,
  transform,
  ...comboBoxProps
}: ComboBoxFieldProps) => {
  const fieldProps = useField({
    name,
    rules,
    defaultValue,
    errorMessage,
    isRequired,
    onChange,
    transform,
  })

  return (
    <ComboBox
      {...fieldProps}
      {...comboBoxProps}
    />
  )
```

src/components/Common/UI/ComboBox/ComboBox.tsx
serikjensen, line 40: Nit: looks like this isn't used anymore?

serikjensen (approved):
LGTM! thanks for the updates!

PR #236, opened 2025-04-11
chore: welcome page improvements

jeffredodd:
Just some light improvements and updates to text and styles to utilize more of our internal components. Also fixed broken links.


serikjensen (commented):
Since this welcome page is ladle specific, should we consider moving it into the .ladle directory? Looks like it was already here in the previous PR so must not have been paying attention to the dir structure 😅

jeffredodd:
> Since this welcome page is ladle specific, should we consider moving it into the .ladle directory? Looks like it was already here in the previous PR so must not have been paying attention to the dir structure 😅

Good catch! Fixed

serikjensen (approved):
Looks great!

PR #237, opened 2025-04-11
feat: decouple switchfield from react-hook-forms

jeffredodd:
This change focuses on the following updates
- Decouple Switch from React Hook Forms
- Create Switch and SwitchField components
- Create stories for testing both Switch and SwitchField components
- Update tests for Switch to validate input functionality

## Proof of Functionality
https://github.com/user-attachments/assets/9ee05556-dbcd-47d7-b99a-c6825fc0f3a6


src/components/Common/UI/Switch/Switch.module.scss
```diff
@@ -0,0 +1,98 @@
+:global(.react-aria-Switch) {
```
serikjensen, line 1: Nit: for encapsulation we might consider placing this within a parent `.root` class or similar, otherwise it'll end up being truly global scope (in which case we'd probably want to just keep it at the global level)

src/components/Common/UI/Switch/Switch.module.scss
```diff
@@ -0,0 +1,98 @@
...
+      }
+    }
+
+    &[data-selected] .indicator::after {
+      background: var(--g-colors-gray-400);
+    }
+
+    & + .react-aria-Switch-description {
+      color: var(--g-typography-disabledTextColor);
+    }
```
serikjensen, line 92: Any description/label styles we should be able to remove since we updated it to use HorizontalFieldLayout

src/components/Common/UI/Switch/Switch.module.scss
```diff
@@ -0,0 +1,98 @@
...
+  }
+}
+
+:global(.react-aria-Switch-description) {
+  @include description;
+}
```
serikjensen, line 98: Same here, we can likely delete

src/components/Common/UI/Switch/Switch.tsx
```diff
@@ -0,0 +1,68 @@
...
+import { useFieldIds } from '../hooks/useFieldIds'
+import {
+  HorizontalFieldLayout,
+  type SharedHorizontalFieldLayoutProps,
+} from '../HorizontalFieldLayout'
+import styles from './Switch.module.scss'
+
+export interface SwitchProps
+  extends SharedHorizontalFieldLayoutProps,
+    Omit<AriaSwitchProps, 'children'> {
```
serikjensen, line 11: We probs don't want to extend AriaSwitchProps since we're trying to keep our interfaces completely defined on our end. If we do keep this IMO we should specifically use Pick to pull out what we need. my hope would be we could just reference native html props to supplement where needed instead of theirs tho

src/components/Common/UI/Switch/Switch.tsx
```diff
@@ -0,0 +1,68 @@
...
+  isSelected?: boolean
```
serikjensen, line 12: Nit: thoughts on using `checked` instead as a name here for consistency with Radio and Checkbox? That makes more sense from a native html perspective (since it's internally a checkbox) but not sure if it's actually a very intuitive name for engineers

serikjensen (commented):
Very exciting to see these wrapping up! just a few comments

src/components/Common/UI/Switch/Switch.tsx
```diff
@@ -0,0 +1,68 @@
+import { Switch as _Switch, type SwitchProps as AriaSwitchProps } from 'react-aria-components'
+import { useFieldIds } from '../hooks/useFieldIds'
+import {
+  HorizontalFieldLayout,
+  type SharedHorizontalFieldLayoutProps,
+} from '../HorizontalFieldLayout'
+import styles from './Switch.module.scss'
+
+export interface SwitchProps
```
serikjensen, line 9: Could we also update to add an inputRef prop to SwitchProps and hook it up? we end up getting a dom error from hook form if that prop isn't handled

jeffredodd: Yup will do!

src/components/Common/Fields/AllFieldsDemo.stories.tsx
```diff
@@ -9,11 +9,15 @@ import { SelectField } from './SelectField'
...
 import { RadioGroupField } from './RadioGroupField'
 import { CheckboxGroupField } from './CheckboxGroupField'
+import { DatePickerField } from './DatePickerField/DatePickerField'
+import { SwitchField } from './SwitchField/SwitchField'
 
+// Adding a meta object for title
 export default {
-  title: 'UI/Form/Examples',
+  title: 'UI/Form/Examples', // Updated to be under UI/Form instead of top-level Form
 }
```
serikjensen, line 18: Nit: we can probably remove these extra comments?

src/components/Common/UI/Switch/Switch.module.scss
```diff
@@ -0,0 +1,92 @@
...
+
+    font-size: toRem(15);
+    color: var(--g-typography-textColor);
+    cursor: pointer;
+    forced-color-adjust: none;
+
+    .body {
+      display: flex;
+      align-items: start;
+      gap: 0.571rem;
```
serikjensen, line 16: Nit: i know these are copy/pasted styles but we should probably circle back to fix styles like this that are rendering subpixel (since diff browsers handle them differently). Ex. imo this should be one of the following

```
toRem(8) //.5rem
toRem(9) // .5625rem
0.5rem
0.5625rem
```

src/components/Common/UI/Switch/Switch.module.scss
```diff
@@ -0,0 +1,92 @@
...
+    .switch-details {
+      display: flex;
+      flex-direction: column;
+      position: relative;
+      top: -2px;
+    }
+
+    .indicator {
+      width: 2.286rem;
+      height: 1.143rem;
```
serikjensen, line 28: Nit: Same with these values, we should get them off of subpixel rendering

src/components/Common/UI/Switch/Switch.module.scss
```diff
@@ -0,0 +1,92 @@
...
+      transition: all 200ms;
+      position: relative;
+
+      &::after {
+        content: '';
+        position: absolute;
+        top: 0.143rem;
+        left: 0.143rem;
+        width: 0.857rem;
+        height: 0.857rem;
```
serikjensen, line 40: Nit: and these ones as well

src/components/Common/UI/Switch/Switch.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+      isInvalid
+      errorMessage="You must accept the terms to continue"
+      description="By enabling this, you agree to our terms of service"
+      checked={value}
+      onChange={checked =>
+        handleCheckboxChange({ target: { checked } } as ChangeEvent<HTMLInputElement>)
+      }
+    />
+  )
+}
```
serikjensen, line 115: Nit: on these stories, you could probably just create a dedicated Switch state handler that would accept a boolean instead of an event. Then you wouldn't have to do this workaround with creating an event object manually with target hard coded

src/components/Common/UI/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+    const switchWrapper = screen.getByRole('switch').closest('label')
+    expect(switchWrapper).toHaveAttribute('data-selected', 'true')
+  })
+
+  it('shows as not selected when checked is false', () => {
+    render(<Switch {...defaultProps} checked={false} />)
+
+    const switchWrapper = screen.getByRole('switch').closest('label')
+    expect(switchWrapper).not.toHaveAttribute('data-selected', 'true')
+  })
```
serikjensen, line 69: Nit: on these ones i might recommend verifying that the underlying `input` element has the checked attribute set similar to tests for checkbox. That would make these tests more agnostic to react aria's implementation

src/components/Common/UI/Switch/Switch.tsx
```diff
@@ -0,0 +1,74 @@
...
+  extends SharedHorizontalFieldLayoutProps,
+    Pick<InputHTMLAttributes<HTMLInputElement>, 'name' | 'id' | 'checked'> {
+  onBlur?: (e: FocusEvent) => void
+  onChange?: (checked: boolean) => void
+  inputRef?: RefObject<HTMLInputElement>
```
serikjensen, line 16: In order for this to work with hook form, we'll need this to be able to handle a callback ref

that means this should be loosened to accept

```suggestion
  inputRef?: Ref<HTMLInputElement>
```

Unfortunately some of React Aria's components end up only accepting RefObject 😔 Hopefully that's not the case here. If it _is_ the case, i was doing the following

```
inputRef?: Ref<HTMLInputElement>

export function Switch({
  ...
  inputRef,
  ...
}: SwitchProps) {
  ...
  const internalInputRef = useRef<HTMLInputElement>(null)

  useEffect(() => {
    if (inputRef) {
      if (typeof inputRef === 'function') {
        inputRef(internalInputRef.current)
      } else {
        inputRef.current = internalInputRef.current
      }
    }
  }, [])
  return (
    <HorizontalFieldLayout
      ...
    />
      <_Switch
        ...
        inputRef={internalInputRef}
...
```

i ended up not having to use it because i ejected from the react aria implementations for checkbox and radio inputs but i had written this when i thought i needed to get around the fact that they only allowed a ref object

serikjensen (commented):
Looking at this with fresh eyes this morning, left some nits that are probably out of scope. Just one required change to update the type of inputRef to be `Ref` instead of `RefObject` so that it works with react hook form. Everything else is optional!

src/components/Common/Fields/AllFieldsDemo.stories.tsx
jeffredodd: Will be resolved in next iteration. Cursor cruft.

src/components/Common/UI/Switch/Switch.module.scss
jeffredodd: Will be resolved in next iteration.

src/components/Common/UI/Switch/Switch.module.scss
jeffredodd: Will be resolved in next iteration.

src/components/Common/UI/Switch/Switch.stories.tsx
jeffredodd: Will be resolved in next iteration.

src/components/Common/UI/Switch/Switch.tsx
jeffredodd: Will be resolved in next iteration.

src/components/Common/UI/Switch/Switch.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+    render(<Switch {...defaultProps} isRequired />)
+
+    const labelText = screen.getByText('Test Switch')
+    const parentElement = labelText.parentElement
+
+    expect(parentElement?.textContent).toContain('Test Switch')
+
+    const { container } = render(<Switch {...defaultProps} isRequired />)
+    expect(container.innerHTML).toContain('Test Switch')
+  })
```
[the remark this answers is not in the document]
jeffredodd: Looks like we can't require correctly with React-aria, we have a similar issue with checkbox. RHF validation still works and was checked. So for now i've removed this test as it doesn't provide a valid test of functionality.

serikjensen (approved):
Looks great! thanks for the updates!

PR #238, opened 2025-04-11
feat: add new RadioGroup separate from hook form

serikjensen:
This PR
* adds a new RadioGroup component separate from React Hook Form
* adds a new Radio component which is styling around the native html input
* creates a RadioGroupField component that connects to hook form

## Proof of functionality

Form submission with the correct values updating

https://github.com/user-attachments/assets/a60b43c0-e1e4-4392-8ccf-82c08a4b0d29

Required state working and focus on the first element behaving

https://github.com/user-attachments/assets/d1590ad6-80c4-40a0-9f26-eeaec1e2ebe4

All Radio and RadioGroup examples

https://github.com/user-attachments/assets/41d990b2-9652-4331-ae64-1901357d4c34


jeffredodd:
Can you ask cursor to update the AllFields story with your RadioField? Just want to make sure we get it in there. Its been pretty good at making up fake things to put in too so I hope it works for you too.

src/components/Common/UI/RadioGroup/RadioGroup.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const handleInputRef = useForkRef(inputRefFromProps, inputRef)
+
+  const { inputProps, isSelected, isDisabled } = useRadio(
+    {
+      ...props,
+      value,
+      // Hack to suppress aria-label warning from React Aria. We don't actually
+      // use children for the mapping of our radio to React Aria and we already
+      // configure a label for our radios.
+      children: ' ',
```
jeffredodd, line 64: LOL, ugh, why React Aria why!

jeffredodd (approved):
LGTM - If you add the AllFields story fix this will be even better.

src/components/Common/UI/RadioGroup/RadioGroup.tsx
serikjensen: Right?? 😂😭

PR #239, opened 2025-04-11
chore: combobox api adjustment

serikjensen:
* Updates combobox options to be consistent with select (before they were requiring an id even tho that wasn't exposed on the options type)
* Updates to add an inputRef

## Proof of functionality

https://github.com/user-attachments/assets/ccf88933-39d3-48a2-9d61-1e687c4ae21c


jeffredodd:
Good catch thanks!

jeffredodd (approved):

PR #240, opened 2025-04-11
Detect circular dependencies

scally:
Use `vite-plugin-circular-dependency` to detect import cycles.

We want to detect import cycles to help keep hot module reload, tree-shaking, and code splitting healthy. 

We were using `eslint-plugin-import` previously, but it wasn't detecting cycles during testing, so we swapped to a vite plugin that provides similar functionality. 

It's less pleasant for our developer experience for this to be a vite plugin instead of a eslint plugin, but we are accepting the tradeoff of later detection of cycles for accuracy + speed.

Currently, this PR only warns on cycles because we have a large amount of existing cycles to fix. Once those cycles are fixed, we will make this an error. Since this PR adds the vite build to the lint-staging step, this will fail early pre-commit upon build cycles once we change the value of `circleImportThrowErr`. CI already runs `vite build` so we will catch this in CI as well. 

Screenshot 2025-04-11 at 3 16 55 PM


lint-staged.config.js
```diff
@@ -1,3 +1,3 @@
 export default {
-  '*.ts*': ['npm run format --', 'npm run lint --'],
+  '*.ts*': ['npm run build . --', 'npm run format --', 'npm run lint --'],
```
scally, line 2: Without the `.` here vite build kept failing out with the error `Could not resolve entry module "vite.config.ts/index.html"`

vite.config.ts
```diff
@@ -33,6 +34,9 @@ export default defineConfig({
       },
       include: ['**/*.svg?react', '**/*.svg'],
     }),
+    circularDependencyDetector({
+      circleImportThrowErr: false,
```
scally, line 38: Will make this `true` once existing cycles are fixed

lint-staged.config.js
scally: Adding the vite build step here so that in the future when we set cycle detection to return a nonzero exit code via `circleImportThrowErr`, this will result in adding a new import cycle to break pre-commit linting

jeffredodd (approved):
Here we go! 🚢

serikjensen (approved):
LGTM! Glad this worked!

PR #243, opened 2025-04-14
build: remove deprecated `generate` step from publishing

scally:
It was removed here https://github.com/Gusto/embedded-react-sdk/pull/218/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

It looks like we can just run `build` since the openapi part of `generate` is no longer necessary and `build` runs the `i18n:generate` step


serikjensen (approved):
LGTM!

PR #244, opened 2025-04-14
feat: remove numberfield and replace with numberinputfield 

jeffredodd:
- Swap out NumberField to NumbertnputField
- Delete NumberField from common components
- Validate tests are passing E2E tests

Testing
- E2E tests pass
- Validated in Ladle NumberInputField works as expected
- Validated in dev mode that NumberInputField works in Self onboarding flow
- Updated tests where label querying requires `(optional)` now since we fixed our labeling


src/components/Employee/PaymentMethod/Split.tsx
```diff
@@ -204,24 +198,16 @@ function PercentageField({
...
         label={t('splitAmountLabel', {
           name: DOMPurify.sanitize(split.name ?? ''),
           account_number: DOMPurify.sanitize(split.hiddenAccountNumber ?? ''),
           interpolation: { escapeValue: false },
         })}
-        formatOptions={{
-          style: 'decimal',
-          currency: currency,
-        }}
+        format="percent"
```
serikjensen, line 208: Question: this was decimal before, based on the naming tho this seems like the correct update?

serikjensen (commented):
Looking good! biggest question here is around NumberInput handling of cleared/undefined values. It ends up sending `NaN` in those cases. I'm wondering if there is work that needs to be done to correctly handle `NaN` in those cases or if we need to supply errorMessage in those cases. We could also break some of this out into follow up tickets if desired

src/components/Common/Inputs/NumberField.tsx
```diff
@@ -1,85 +0,0 @@
-import type { RefAttributes } from 'react'
```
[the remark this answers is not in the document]
jeffredodd: GWS-4514 created

src/components/Employee/Deductions/DeductionForm.tsx
```diff
@@ -45,20 +45,19 @@ export const DeductionForm = () => {
...
+          isRequired
+          label={t('deductionAmountLabel')}
+          format={watchedDeductAsPercentage === 'true' ? 'percent' : 'currency'}
+          min={0}
+          errorMessage={t('validations.amount')}
         />
+        <NumberInputField name="limit" label={t('annualMaxLabel')} format="currency" min={0} />
+      </div>
+      {watchedRecurring === 'true' && (
+        <NumberInputField name="maxOccurrences" label="Maximum Occurrences" min={0} />
```
dmortal, line 60: Hardcoded label - i18n?

src/components/Employee/PaymentMethod/Split.tsx
```diff
@@ -6,8 +6,9 @@ import { ErrorMessage } from '@hookform/error-message'
 import { Fragment } from 'react/jsx-runtime'
 import DOMPurify from 'dompurify'
 import { useState } from 'react'
+import { z } from 'zod'
```
dmortal, line 9: Why do we have `zod` here?

dmortal (commented):
Just 2 small issues, lgtm otherwise

serikjensen (approved):
LGTM!

PR #245, opened 2025-04-14
Fix cyclical dependencies in Employee components

scally:
part 1 of ??


jeffredodd (approved):
Thanks for the first part! Curious if it will turn into a trilogy 🚢

PR #246, opened 2025-04-14
feat: remove select and replace with selectfield

jeffredodd:
- Swap out Select to SelectField
- Delete Select from common components
- Validate tests are passing E2E tests

Testing
- E2E tests pass
- Validated in Ladle SelectField works as expected
- Validated in dev mode that SelectField works in Pay Schedule


src/components/Common/Inputs/Select.tsx
```diff
@@ -1,118 +0,0 @@
-import type { RefAttributes } from 'react'
```
serikjensen, line 1: Same comment here as the others, maybe worth creating a follow up ticket to move the Select styles into the updated Select component along with a dedicated stylesheet

src/components/Common/PaginationControl/PaginationControl.tsx
```diff
@@ -39,18 +38,20 @@ export const PaginationControl = ({
...
         <div className={style.paginationControlCount}>
           <section>
-            <Select
-              control={control}
-              aria-label={t('labels.paginationControllCountLabel')}
+            <SelectField
               name="pageSize"
-              onSelectionChange={n => {
+              label={t('labels.paginationControllCountLabel')}
+              aria-label={t('labels.paginationControllCountLabel')}
```
serikjensen, line 44: I think this should be updated as follows yeah?

```suggestion
              shouldVisuallyHideLabel
```

Not sure why it isn't throwing a type error for `aria-label`

src/components/Employee/Compensation/Edit.tsx
```diff
@@ -90,14 +94,11 @@ export const Edit = () => {
           errorMessage={t('validations.exemptThreshold', {
             limit: format(FLSA_OVERTIME_SALARY_LIMIT),
           })}
-          items={classificationOptions}
+          options={classificationOptions.map(o => ({ value: o.id, label: o.name }))}
```
serikjensen, line 97: Looks like we already define `classificationOptions` with a map operation elsewhere in this file. Maybe worth just updating that to use value/label instead so we don't have to map twice?

src/components/Employee/Compensation/Edit.tsx
```diff
@@ -123,37 +124,30 @@ export const Edit = () => {
...
         </>
       )}
-      <Select
-        control={control}
+      <SelectField
         name="paymentUnit"
         label={t('paymentUnitLabel')}
         description={t('paymentUnitDescription')}
-        items={paymentUnitOptions}
+        options={paymentUnitOptions.map(o => ({ value: o.id, label: o.name }))}
```
serikjensen, line 143: Same feedback here, looks like we have control over `paymentUnitOptions` in this file so maybe we consider just updating that to have `value` and `label` instead of having to map?

src/components/Employee/Taxes/FederalForm.tsx
```diff
@@ -32,17 +32,14 @@ export function FederalForm() {
...
   return (
     <>
-      <Select
-        control={control}
+      <SelectField
         name="filingStatus"
         label={t('federalFilingStatus1c')}
         placeholder={t('federalFillingStatusPlaceholder')}
-        items={filingStatusCategories}
+        options={filingStatusCategories.map(cat => ({ value: cat.id, label: cat.name }))}
```
serikjensen, line 39: Same feedback, we can probably just update the `filingStatusCategories` constant to have `value` and `label` instead of needing to map

serikjensen (commented):
Couple of requests on this one

src/components/Employee/Compensation/Edit.tsx
jeffredodd: Good catch!

src/components/Employee/Taxes/FederalForm.tsx
jeffredodd: Updated

src/components/Employee/Compensation/Edit.tsx
jeffredodd: Updated

serikjensen (approved):
LGTM!

PR #247, opened 2025-04-14
feat: remove combobox and replace with comboboxfield

jeffredodd:
- Swap out single instance of ComboBox to ComboBoxField
- Delete ComboBox from common components

Testing
- E2E tests pass
- Validated in Ladle ComboBoxField works as expected


src/components/Company/Industry/Edit.tsx
```diff
@@ -1,22 +1,28 @@
...
   const { t } = useTranslation('Company.Industry')
-  const { control } = useFormContext<IndustryFormFields>()
   const { items } = useIndustryItems()
 
+  const options = useMemo(() => {
+    return items.map(item => ({
+      label: item.name,
+      value: item.id,
+    }))
+  }, [items])
```
serikjensen, line 20: Looking closer at the code, it seems we might be able to improve performance here by just using `label`/`value` naming in the generated items over in IndustrySelect here

https://github.com/Gusto/embedded-react-sdk/blob/f2a9966a3ced9a02d94ff8e17a1694df306e3942/src/components/Company/Industry/IndustrySelect.tsx#L27-L32

Which would eliminate the need for this mapping

jeffredodd: Good catch, that is very logical and also proof we picked the right API that we're deleting code.

src/components/Company/Industry/IndustrySelect.tsx
```diff
@@ -22,11 +22,18 @@ export function IndustrySelect({
...
-  const [items, setItems] = useState<ComboBoxItem[]>([])
+  const [items, setItems] = useState<ComboBoxOption[]>([])
 
   useEffect(() => {
     const loadItems = async () => {
-      setItems((await loadAll()).map(({ title: name, code: id }) => ({ id, name })))
+      setItems(
+        (await loadAll()).map(({ title, code }) => ({
+          id: code,
+          name: title,
```
serikjensen, line 32: Can these just be removed? Isn't it just label/value for the combobox options?

jeffredodd: Lol you're faster than me, should be updated now

serikjensen (approved):
Excellent!

PR #249, opened 2025-04-15
chore: replace CheckboxGroup instances with new field components

serikjensen:
This PR updates all current CheckboxGroup usages from `Common/Inputs` and swaps for the corresponding `CheckboxField` and `CheckboxGroupField` components.
* SignatureForm has been updated to just use `CheckboxField` since it only actually needed a single checkbox (previously this was a limitation of react aria)
* Company locations CheckboxGroup has been swapped with CheckboxGroupField (note, there are some oddities with the behavior of checkbox here due to potential business logic with mailing and filing address, but this is unrelated to the swap here. Read more in [this ticket](https://gustohq.atlassian.net/browse/GWS-4431))

### Signature form working as before

https://github.com/user-attachments/assets/e168fcd3-101f-4b29-86e1-51d17d3abce1

### Company locations functioning as before

https://github.com/user-attachments/assets/1bd0962c-d796-4829-8f79-fdb8d0130673

### Style consistency for CheckboxGroup after removing the legacy styles

Screenshot 2025-04-15 at 2 31 42 PM


src/components/Common/UI/Switch/Switch.test.tsx
```diff
@@ -1,5 +1,5 @@
 import { describe, expect, it, vi } from 'vitest'
-import { render, screen, within } from '@testing-library/react'
```
jeffredodd, line 2: ty

jeffredodd (approved):
🚢

PR #256, opened 2025-04-16
fix: migrate styles too ui folder as modules

jeffredodd:
This PR aims to achieve the following....

- Migrate `_{component}.scss` files related to form work into their own `{component}.styles.scss` module
- Decouple DatePicker from Calendar Viewer
- Remove unused styles and update `sdk.scss` to only include needed styles


serikjensen (approved):
LGTM!

PR #257, opened 2025-04-16
refactor: remove cycles from Common components; cycles are now an error

scally:


vite.config.ts
```diff
@@ -34,9 +34,7 @@ export default defineConfig({
       },
       include: ['**/*.svg?react', '**/*.svg'],
     }),
-    circularDependencyDetector({
-      circleImportThrowErr: false,
```
scally, line 38: Cycles are a build breaking error now; locking this down

serikjensen (approved):
Amazing!!

PR #258, opened 2025-04-16
chore: update checkbox instances and fix pagination

serikjensen:
This PR
* Updates to correctly point Checkbox instances to the CheckboxField instead
* Updates those usages to use the correct updated CheckboxField api
* Updates Pagination controls to have internally managed state rather than needing to be rendered in a form provider
* Msc cleanup items that are outstanding with the migration


jeffredodd (approved):
Love that we can use RHF'less selects now. LGTM 🚢

PR #263, opened 2025-04-17
chore: standardize change handlers and values for all form components

serikjensen:
This PR standardizes change handlers and values for all form components.

All form input components in the UI directory now have a `value` prop which sets the value, and an `onChange` handler that passes `value` of the same type. This allows for consistency and simplification. Ex. Look at reduced complexity in useField which now has a single value type. Also in the ladle helper for managing state which can now have a universal handler. This should be a more straightforward contract to maintain with consumers.

Updates to adopt this approach throughout the codebase including stories and test files.

### DIFF
--- a/.ladle/helpers/LadleState.tsx
+++ b/.ladle/helpers/LadleState.tsx
@@ -9,63 +9,11 @@ import { useState } from 'react'
 export function useLadleState<T>(actionName: string, initialValue?: T) {
   const [value, setValue] = useState<T | undefined>(initialValue)
 
-  const handleAction = (newValue: T) => {
+  const handleChange = (newValue: T) => {
     action(actionName)(newValue)
     setValue(newValue)
     return newValue
   }
-
-  // For input elements that pass event objects
-  const handleInputChange = (event: React.ChangeEvent<HTMLInputElement>) => {
-    const newValue = event.target.value as unknown as T
-    action(actionName)(newValue)
-    setValue(newValue)
-    return newValue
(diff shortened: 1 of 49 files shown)


.ladle/helpers/LadleState.tsx
```diff
@@ -9,78 +9,19 @@ import { useState } from 'react'
 export function useLadleState<T>(actionName: string, initialValue?: T) {
   const [value, setValue] = useState<T | undefined>(initialValue)
 
-  const handleAction = (newValue: T) => {
+  const handleChange = (newValue: T) => {
```
jeffredodd, line 12: This is proof this is the right change 😆

src/components/Common/UI/Radio/Radio.test.tsx
```diff
@@ -54,19 +45,8 @@ describe('Radio', () => {
...
-        target: expect.objectContaining({ checked: true }),
-      }),
-    )
-    expect(onChangeFromInputProps).toHaveBeenCalledWith(
-      expect.objectContaining({
-        target: expect.objectContaining({ checked: true }),
-      }),
-    )
+    expect(onChange).toHaveBeenCalledTimes(1)
+    expect(onChange).toHaveBeenCalledWith(true)
```
jeffredodd, line 49: This is awesome to see some of these changes

jeffredodd (approved):
Beautiful! :shipit: :shipit: :shipit:

PR #264, opened 2025-04-17
feat: introduce component adapter + setup fields to be adapted

jeffredodd:
This PR does a lot, lets review all the changes that have been made so far...

## Changes
- Renamed GustoAPIProvider to GustoProvider (Still exporting as GustoAPIProvider to provide backwards compatibility)
- Implemented a Default UI adapter which plumbs in our existing UI/React Aria components
- [Ladle] - Added toggle to switch between our default adapter and a plain HTML adapter for testing
- Used Component context to replace all Form Field components

## Proof of Concept
https://github.com/user-attachments/assets/2d1ea951-b4bd-426c-9fc5-6beabc2da1c2


src/contexts/ComponentAdapter/adapters/ReactAriaComponentAdapter.tsx
```diff
@@ -0,0 +1,34 @@
+import type { ComponentsContextType } from '../ComponentsProvider'
```
serikjensen, line 1: I'd recommend naming this file something generic like maybe even `DefaultComponentAdapter.tsx`? That way if we moved away from react aria in the future to a diff underlying lib we wouldn't have to rename :D

src/contexts/GustoApiProvider/GustoProviderWithUIAdapter.tsx
```diff
@@ -0,0 +1,39 @@
+import type React from 'react'
```
serikjensen, line 1: My thought here is that the GustoProviderWithUIAdapter would be the more common one so maybe this should just be GustoProvider? my reasoning there would be so that the power user needs to deal with the weirder name 😂 so if you don't care about supplying your own components and are fine rolling with our stuff you just bring in the intuitive `GustoProvider` component. If you want to do some advanced usage you bring in `GustoProviderWithoutUIAdapter` or whatever we want to call it naming is hard lol

src/contexts/GustoApiProvider/GustoProvider.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const { children, ...providerProps } = props
+  const { config, lng, locale, currency, theme, components } = useGustoProvider(providerProps)
+
+  const mergedComponents = {
+    ...defaultComponents,
+    ...components,
+  }
+
+  return (
+    <ComponentsProvider value={mergedComponents}>
```
serikjensen, line 29: Assume we meant to leave out the components from this one?

src/contexts/GustoApiProvider/GustoProviderWithUIAdapter.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * A provider that accepts UI component adapters through the components prop
+ */
+const GustoProviderWithUIAdapter: React.FC<GustoProviderWithUIAdapterProps> = props => {
+  const { children, ...providerProps } = props
+  const { config, lng, locale, currency, theme, components } = useGustoProvider(providerProps)
+
+  return (
+    <ComponentsProvider value={components as ComponentsContextType}>
```
serikjensen, line 25: 🤔 wonder why we end up needing to cast using `as` here? might be worth figuring out

src/contexts/GustoApiProvider/GustoProviderWithUIAdapter.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <ErrorBoundary FallbackComponent={InternalError}>
+        <ThemeProvider theme={theme}>
+          <LocaleProvider locale={locale} currency={currency}>
+            <I18nextProvider i18n={SDKI18next} key={lng}>
+              <ReactSDKProvider url={config.baseUrl}>{children}</ReactSDKProvider>
+            </I18nextProvider>
+          </LocaleProvider>
+        </ThemeProvider>
+      </ErrorBoundary>
```
serikjensen, line 34: You were probably already getting to this, but we could just bring in GustoApiProviderWithoutUIAdapter here and then keep all of that centralized instead of needing to duplicate the setup. worried about those configurations getting out of sync which could be a pain to debug 😛

src/contexts/ComponentAdapter/adapters/ReactAriaComponentAdapter.tsx
jeffredodd: Thats fair, I went back and forth if we should name it or not but i'm game to call it Default

.ladle/adapters/PlainComponentAdapter.tsx
```diff
@@ -0,0 +1,718 @@
...
+    return (
+      <div className="field-layout">
+        {!shouldVisuallyHideLabel && (
+          <label htmlFor={inputId}>
+            {label}
+            {isRequired && <span aria-hidden="true"> *</span>}
+          </label>
+        )}
+        {shouldVisuallyHideLabel && (
+          <label htmlFor={inputId} className="visually-hidden">
```
serikjensen, line 45: Nit: i'm assuming we need a stylesheet that is ladle specific that would include this?

src/components/Company/Industry/IndustrySelect.test.tsx
```diff
@@ -12,9 +14,11 @@ describe('IndustrySelect', () => {
     vi.mocked(loadAll).mockResolvedValue([{ code: 'abcd', title: 'Do Things' }])
 
     render(
-      <ThemeProvider>
-        <IndustrySelect />
-      </ThemeProvider>,
+      <ComponentsProvider value={defaultComponents}>
```
serikjensen, line 17: Nit: should this just be updated to use the gusto test provider instead? With that update this would've gotten the adapter changes/context automatically

src/contexts/GustoApiProvider/useGustoProvider.tsx
```diff
@@ -24,31 +18,32 @@ export type Dictionary = Record<
...
   locale?: string
   currency?: string
   theme?: DeepPartial<GTheme>
-  children?: React.ReactNode
   queryClient?: QueryClient
+  components?: Partial<ComponentsContextType>
 }
 
-const GustoApiProvider: React.FC<GustoApiProps> = ({
+export const useGustoProvider = ({
```
serikjensen, line 32: Nit: wondering if it's worth the indirection to move this to the hook vs. just keeping this in GustoProviderCustomUIAdapter? We only end up with one usage of this now that GustoProvider just brings in the GustoProviderCustomUIAdapter

serikjensen (commented):
Looking good! going to hold off on approval until it gets more eyes since i'm not sure if this is set to automerge 😅

.ladle/adapters/PlainComponentAdapter.tsx
jeffredodd: Oh yeah for sure, I think there is going to be a point where we comb the adapter for bugs, this is more POC in our Playground at the moment.

dmortal (approved):

serikjensen (approved):
Awesome 👏

PR #266, opened 2025-04-21
feat: update field components to support generic values

serikjensen:
This PR updates our field components to support generic values. This was necessary for our field components to work with the state tax forms.

This change takes place in the field components instead of the UI components directly. This is because we want to make the contract between us and consumers as simple as possible. This makes it so that for components with options, they just support a consistent case with a string value and options that will have a string value inside corresponding. We take on all the complexity of managing generic types and ensuring that they map to string values.

This PR centralizes that logic inside of a hook and applies it to each component where generic values make sense including
* RadioGroupField
* CheckboxGroupField
* ComboBoxField
* SelectField

This also restores logic to disable the file new hire report when it has already been set for an employee

### File new hire report disabled
Screenshot 2025-04-21 at 3 52 01 PM

### File new hire report select input working
This is a good test case since this is set as a boolean value

https://github.com/user-attachments/assets/0b206bb2-6328-401d-8afc-ddc7508906fa

### Ladle instances examples with generic typed values

https://github.com/user-attachments/assets/7897dc62-6bf0-4e6d-acb7-465118fafff3

Uploading Screen Recording 2025-04-21 at 3.51.28 PM.mov…

### DIFF
--- a/src/components/Common/TaxInputs/TaxInputs.tsx
+++ b/src/components/Common/TaxInputs/TaxInputs.tsx
@@ -1,6 +1,5 @@
 import DOMPurify from 'dompurify'
 import { Text } from 'react-aria-components'
-import { type Control } from 'react-hook-form'
 import type { EmployeeStateTaxQuestion } from '@gusto/embedded-api/models/components/employeestatetaxquestion'
 import { type TaxRequirement } from '@gusto/embedded-api/models/components/taxrequirement'
 import { SelectField } from '../Fields/SelectField/SelectField'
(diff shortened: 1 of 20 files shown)


src/components/Common/TaxInputs/TaxInputs.tsx
```diff
@@ -66,9 +69,11 @@ export function NumberInput({
   const { key, label, description } = question ? question : requirement
   const value = question ? question.answers[0]?.value : requirement.value
 
+  if (!key) return null
```
serikjensen, line 72: If there's no key, this won't get associated with a hook form input and won't actually submit anything so no sense in rendering an input in that case

src/components/Employee/PaymentMethod/Split.tsx
```diff
@@ -122,10 +122,10 @@ export function Split() {
         label={t('splitByLabel')}
         onChange={value => {
           switch (value) {
```
serikjensen, line 124: This was actually caught/enforced by the typing thanks to these changes 🙌

PR #268, opened 2025-04-22
feat: migrate button to use component context

jeffredodd:
This PR is responsible for migrating our Button component from its isolated state into our reusable component adapter. This incldues....

- Moving Button component with React-aria into /UI folder
- Replacing instances of <Button> with <Components.Button> in our codebase
- Migrating styles from _Button.scss to Button.module.scss


src/components/Common/UI/Button/Button.tsx
```diff
@@ -0,0 +1,41 @@
+import { Button as AriaButton } from 'react-aria-components'
+import { type ButtonProps } from './ButtonTypes'
+import styles from './Button.module.scss'
+
+export function Button({
+  variant = 'primary',
+  isError = false,
+  isLoading = false,
+  isDisabled = false,
+  buttonRef,
```
serikjensen, line 10: Would it be preferable to just expose this as `ref` rather than `buttonRef`? `inputRef` makes sense for the field components since they render the input along with the helper/error text and label so it's less clear what you get the ref to. In this case tho feels like it could just be `ref`?

I could see buttonRef being more straightforward for the adapter case where we want it explicitly noted for the consumer what they need to wire up. But it'll also put us into a naming situation where we'll constantly need to prefix `ref` for our components (which may not be a big deal)

src/components/Common/UI/Button/ButtonTypes.ts
```diff
@@ -0,0 +1,31 @@
...
+    | 'onKeyUp'
+    | 'aria-label'
+    | 'aria-labelledby'
+    | 'aria-describedby'
+    | 'form'
+    | 'title'
+    | 'tabIndex'
+  > {
+  buttonRef?: Ref<HTMLButtonElement>
+  variant?: 'primary' | 'secondary' | 'tertiary' | 'link' | 'icon'
```
serikjensen, line 24: I'm wondering if these should actually be broken out into multiple button components. Ex. feeling `LinkButton` should be a separate component. It seems so different from the rest of the button styles and may have additional accessibility considerations to account for. To a lesser degree i'd still push for icon to be a distinct component, ex. IconButton, as well. That would be so these could take care of their own clear dedicated API concerns rather than having everything needing to be combined with the button. From experience, icon buttons end up with different sizing considerations and labeling considerations (since they strictly use visually hidden text)

jeffredodd: I'm game to do this, i'm thinking I might want to do it in a different story just so we're not changing a ton of files. 

That said I just did research and we have zero uses of LinkButton in our app only our stories. We have 2 uses of Icon button in our app in Hamburger component + Pagination control. 

Given the low usage, would you be ok with waiting to do this in another story, or does it low usage for both components (And only being used inside our component library itself) warrant waiting?

jeffredodd:
Leaving this note here so I come back and fix it, but looks like during the rebase the useGustoApiProvider hook was restored. I'll remove it to simplify our code as discussed in a prior PR #264

src/components/Common/UI/Button/ButtonTypes.ts
serikjensen: Yes i think separate story is preferable! we'll def want to prioritize it before the api for adapters goes live but no need to do everything all at once here

interesting that link button is not used. If we can get away with that not existing and just removing it that'd be much less confusing for consumers :D maybe we could chat with @aaronlee777 about it

src/components/Common/UI/Button/ButtonTypes.ts
```diff
@@ -0,0 +1,37 @@
...
+  onBlur?: ButtonFocusHandler
+  onFocus?: ButtonFocusHandler
+}
+
+// Specific button variant props
+export type ButtonProps = ButtonBaseProps
+export type ButtonSecondaryProps = ButtonBaseProps
+export type ButtonTertiaryProps = ButtonBaseProps
+export type ButtonLinkProps = ButtonBaseProps
+export type ButtonIconProps = ButtonBaseProps
```
serikjensen, line 37: Can we update ButtonIconProps to make `aria-label` a required prop?

src/components/Common/UI/Button/Button.tsx
```diff
@@ -0,0 +1,40 @@
...
+        onFocus={onFocus}
+        isDisabled={isDisabled || isLoading}
+        data-variant="primary"
+        data-loading={isLoading || undefined}
+        data-error={isError || undefined}
+        onPress={handlePress}
+      >
+        {children}
+      </AriaButton>
+    </span>
```
serikjensen, line 38: So we end up duplicating this a lot across our files. Also, i think it's potentially risky to have a single scss module being brought in for multiple components. I think for clarity having one module stylesheet corresponding to one component helps avoid unintended side effects

That said, duplicating this stylesheet for every one of these is maybe not ideal

In prior systems, I've ended up with a `ButtonBase` or `BaseButton` component or similar that centralizes the styles and markup. Then the component abstractions just reference that

I know that basically lands us in a similar place as before, so here are my thoughts around that
* I don't mind breaking out every one of these into a distinct component (primary, tertiary etc)
* I also wouldn't mind if this stayed in a single button with a variant prop insofar as we moved ButtonLink and ButtonIcon into dedicated components

The only real hard stop i had was breaking out the icon and link variants since those usually end up having different api requirements. Ex. icon button should have a required aria-label prop

Let's chat about this more tomorrow if you get some time!

jeffredodd: Sounds good, yeah I think I like the idea of the API having this broken out but i'm not fully sold on it yet. Also after doing Alert it makes me wonder is it AlertSuccess or Alert variant="success"

src/components/Common/UI/Button/ButtonTypes.ts
```diff
@@ -0,0 +1,33 @@
...
+  isError?: boolean
+  isLoading?: boolean
+  isDisabled?: boolean
+  children?: ReactNode
+  onBlur?: ButtonFocusHandler
+  onFocus?: ButtonFocusHandler
+  variant?: 'primary' | 'secondary' | 'tertiary' | 'icon'
+}
+
+export type ButtonIconProps = Omit<ButtonProps, 'variant'>
```
serikjensen, line 33: Should we update to require aria-label here as well?

src/components/Common/UI/Button/ButtonIcon.tsx
```diff
@@ -0,0 +1,40 @@
...
+        onFocus={onFocus}
+        isDisabled={isDisabled || isLoading}
+        data-variant="icon"
+        data-loading={isLoading || undefined}
+        data-error={isError || undefined}
+        onPress={handlePress}
+      >
+        {children}
+      </AriaButton>
+    </span>
```
serikjensen, line 38: Could this just be updated to use `Button.tsx`? then it wouldn't have to import the module styles from Button or duplicate this markup and the react aria usage stays in a single component

serikjensen (approved):
Looks good! i still would recommend that IconButton be updated to just use `Button` internally which would reduce the duplication of markup, but i know this is blocking stuff so we can always circle back later!

PR #270, opened 2025-04-23
refactor: fix react-refresh warnings for `src/contexts`

scally:
Fixes react-refresh lint warnings for `src/contexts/*`.

This is going to be noisy so I choose an early-ish stopping point.

Along the way, I also added some deprecation warnings for `GustoAPIProvider` and moved everything to `GustoProvider` to _provide_ some consistency and to help folks fix their imports. The way this PR changes it, we can just lop off the `GustoAPIProvider` directory when we are ready to transition away.


jeffredodd (approved):
Thanks for cleaning up component adapter as well as other context. LGTM

PR #271, opened 2025-04-23
feat: introduce alert for component adapter

jeffredodd:
This change is focused on refactoring <Alert> into our component adapter. Since this component wasn't driven by react-aria it was a much easier lift and shift. 

## Proof of functionality 
Screenshot 2025-04-23 at 2 03 43 PM


jeffredodd:
Currently set to merge into button adapter PR only to simplify code review. Will update to main upon final merge.

.ladle/adapters/PlainComponentAdapter.tsx
```diff
@@ -18,6 +20,17 @@ import type {
 import type { ComponentsContextType } from '../../src/contexts/ComponentAdapter/ComponentsProvider'
 
 export const PlainComponentAdapter: ComponentsContextType = {
+  Alert: ({ label, children, variant = 'info', icon }: AlertProps) => {
```
serikjensen, line 23: Did we settle on a name for variant? I wonder if something like `status` might be more effective for conveying the `info` `warning` `success` etc that will be repeated across other components that use this hierarchy. Variant ends up being kinda general

src/components/Common/UI/Alert/AlertTypes.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export interface AlertProps {
+  /** The variant of the alert */
+  variant?: 'info' | 'success' | 'warning' | 'error'
+  /** The label text for the alert */
+  label: string
+  /** Optional children to be rendered inside the alert */
+  children?: ReactNode
+  /** Optional custom icon component to override the default icon */
+  icon?: ComponentType<SVGProps<SVGSVGElement> & { title?: string }>
```
serikjensen, line 12: Nit: i'm wondering if transitioning this to a react node might be easier for consumers to wire up in the adapter?

src/components/Common/UI/Alert/AlertTypes.ts
```diff
@@ -0,0 +1,13 @@
+import type { ReactNode, ComponentType } from 'react'
+import type { SVGProps } from 'react'
+
+export interface AlertProps {
+  /** The variant of the alert */
+  variant?: 'info' | 'success' | 'warning' | 'error'
+  /** The label text for the alert */
+  label: string
```
serikjensen, line 8: Nit:  wondering if label has too many associations for form elements? that might be overthinking it tho. alternative i could think of would be title or heading but those are also pretty loaded

jeffredodd: I think i'm ok with label because we could use `aria-label` on the alert role so it does align with ARIA.

src/components/Common/UI/Alert/AlertTypes.ts
jeffredodd: Good call

.ladle/adapters/PlainComponentAdapter.tsx
jeffredodd: Fixed!

src/components/Common/UI/Alert/Alert.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <div className={styles.root}>
+      <div
+        className={styles.alert}
+        role="alert"
+        aria-labelledby={id}
+        data-variant={status}
+        ref={alertRef}
+      >
+        <div className={styles.icon}>{icon || defaultIcon}</div>
+        <h6 id={id}>{label}</h6>
```
serikjensen, line 37: Nit: do we need to allow for setting a heading level as part of this? ex. https://github.com/Gusto/embedded-react-sdk/blob/04a7d21a2e0f03ee711827f84e4023eba55fb9d7/src/components/Common/DocumentViewer/DocumentViewer.tsx#L12

Was wondering from an a11y perspective how we can keep the heading levels following a logical order when this could be anywhere on the page

serikjensen (approved):

src/components/Common/UI/Alert/Alert.tsx
jeffredodd: Hmmm this is a good question, I'm less worried about going from h2 to h4.... but this is a good thing we should think about perhaps in the audit? We might be able to get a feel for what it should be when evaluating the whole page.

PR #272, opened 2025-04-23
fix: allow for non active fallback address in profile for self onboard

serikjensen:
For self onboarding, we were experiencing an "Effective date is required" issue for users under the following circumstances
* They are invited to self onboard
* Their admin has filled out the compensation step
* They have submitted their profile with their address
* They refresh the flow and start from the beginning

Under the above circumstances, users were getting an error for "Effective date is required" when attempting to submit a profile update and were blocked from continuing the flow.

This PR provides a fix by ensuring an address is supplied when present even if it is not necessarily active. This ensures that the address is updated properly.

## Further context

* This issue was occurring because we were filtering all addresses out by active
* Address being active is determined by if the effective date is today or in the past ([See ZP code reference here](https://github.com/Gusto/zenpayroll/blob/ef341da922644fb5cd75b1dfb39ea2965c4d6ae5/packs/technical_services/addresses/addresses/app/public/addresses/employees/home_address.rb#L77-L79))
* You are able to create an address once without explicitly setting an effective date
* When you create an address without an effective date, it defaults to the work start date (this is desired behavior for external payroll reasons and correct taxation)
* If you attempt to create _an additional address_ after this, you are required to manually specify a start date

### What was happening in our case?
* We were filtering out home addresses if they are not active
* Home address is only active when it is today or in the past
* When the user submitted the profile the first time in self onboarding under the above circumstances, the effective date was in the future, which meant that the address was not set to active
* When the user attempted to submit the profile the second time, instead of updating the address (desired behavior) we were creating an address since the component filtered out all addresses since there's no active one
* Since we were creating an address, we were getting the error for effective date, since effective date is required when you have already created an address once without an effective date

### How do we fix it?
One option was to set the effective_date to today. We were advised to not do that though for external payrolls reasons and correct taxation. Instead, we allow for referencing addresses beyond just active. We give preference to the active address, but if it is not present we can fallback to the first entry in the work address array. This will either be the only entry (corresponding to the one they have already provided in onboarding) or it will be an entry with the soonest effective date present (since these are sorted on the backend)

### Before
Observe how this user in self onboarding has no address information present. They previously already filled out address information, but it is getting filtered out because the effective date is in the future so it is not active. This means that it attempts to create an additional new address which fails with the error:

https://github.com/user-attachments/assets/998f510e-ffb5-4501-89e7-7ba40cb68f2f

### After
Observe the self onboarding user submit the profile once, refresh, reenter the flow, and resubmit the profile successfully without issues

https://github.com/user-attachments/assets/3124e9f2-7522-489c-adb3-c3cf314406db

### DIFF
--- a/src/components/Employee/Profile/Profile.tsx
+++ b/src/components/Employee/Profile/Profile.tsx
@@ -30,6 +30,7 @@ import { Actions } from './Actions'
 import { HomeAddress, HomeAddressSchema, type HomeAddressInputs } from './HomeAddress'
 import { WorkAddress } from './WorkAddress'
 import { ProfileProvider } from './useProfile'
+import { getEmployeeAddressForProfile } from './getEmployeeAddressForProfile'
 import {
   useBase,
   BaseComponent,
@@ -157,9 +158,8 @@ const Root = ({
 
   const existingData = { employee, workAddresses, homeAddresses }
 
-  const currentHomeAddress = homeAddresses
-    ? homeAddresses.find(address => address.active)
-    : undefined
+  const currentHomeAddress = getEmployeeAddressForProfile(homeAddresses)
+
   const currentWorkAddress = existingData.workAddresses?.find(address => address.active)
   const mergedData = useRef({
     employee: existingData.employee,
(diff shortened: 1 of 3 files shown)


scally (approved):
LGTM, thanks!

PR #278, opened 2025-04-24
feat: introduce breadcrumb to component adapter

jeffredodd:
This PR aims to componentize the existing usage of React-aria's Breadcrumb component

Includes...
- Stories
- Tests
- Styles
- Types
- Removal of _Breadcrumb.scss

## Proof of functionality
Screenshot 2025-04-24 at 4 42 20 PM


src/components/Common/UI/Breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,16 @@
...
+export function Breadcrumb({ children, className, isCurrent, href, onClick }: BreadcrumbProps) {
+  return (
+    <AriaBreadcrumb className={className} data-current={isCurrent}>
+      {href ? (
+        <a href={href} onClick={onClick}>
+          {children}
+        </a>
+      ) : (
+        <span>{children}</span>
+      )}
```
serikjensen, line 13: So this kinda brings us back to our discussion the other day... should this component actually be referencing the `Link` component from component context here instead? I would say at a minimum this should be using that component directly so that we can keep our link styles centralized

serikjensen (approved):
LGTM!

PR #279, opened 2025-04-25
feat: introduce table to component adapter

jeffredodd:
Introduce Table component for component adapter.

This change includes
- All Table related elements (Table, TBody, THead, TR, TD, TH) 
- Does not include TFoot or Caption as React-Aria isn't exporting those components
- Tests for Table component
- Migration of styles from _Table.scss to Table module
- Creation of Table stories in Ladle
- Migration of DataView and Compensation List to use Table from Component Context

# React Aria Adapter 
Screenshot 2025-04-24 at 8 05 30 PM
Screenshot 2025-04-24 at 8 05 36 PM
Screenshot 2025-04-24 at 8 05 40 PM
Screenshot 2025-04-24 at 8 05 43 PM

# Plain Adapter
Screenshot 2025-04-24 at 8 05 49 PM
Screenshot 2025-04-24 at 8 05 52 PM
Screenshot 2025-04-24 at 8 06 03 PM
Screenshot 2025-04-24 at 8 07 04 PM


src/components/Common/UI/Table/Table.module.scss
```diff
@@ -1,6 +1,5 @@
-@use '@/styles/Helpers' as *;
-.GSDK {
-  .react-aria-Table {
+.root {
```
dmortal, line 1: Why root? all of our global styles have to be wrapped in `.GSDK` to avoid partner styles polution - if they also use react-aria, this will mess with their global styling

src/components/Common/UI/Table/Table.tsx
```diff
@@ -0,0 +1,14 @@
+import { Table as AriaTable } from 'react-aria-components'
+import classnames from 'classnames'
+import type { TableProps } from './TableTypes'
+import styles from './Table.module.scss'
+
+export const Table = ({ children, className, ...props }: TableProps) => {
+  return (
+    <span className={styles.root}>
```
dmortal, line 8: Is `span` appropriate here semantically?

serikjensen (commented):
I'm thinking that we may also want to just expose a very top level `DataTable` adapter. We could still expose these granular Table components, but as a consumer, i may just want to swap everything out holistically and it'd be nice to be able to do that with a single adapter vs. needing to maintain 5. That would also give them some flexibility in how they want to present the information. For example, we do tables and cards. They might want to do only cards

Let's chat more about this when you get back next week!

.ladle/adapters/PlainComponentAdapter.tsx
```diff
@@ -878,4 +889,72 @@ export const PlainComponentAdapter: ComponentsContextType = {
       </div>
     )
   },
+
+  Table: <T,>({
```
serikjensen, line 893: Hmm so i'm a little worried about consumers needing to support generics 🤔 i'm wondering if we could simplify. let's chat about it tomorrow. Not a blocker, just something to think about with other components

My thought here is that ultimately the consumer is only concerned about what to render. So i'm wondering if there's a version of the Table adapter we can expose that will effectively provide them with data with react nodes instead of needing them to call render functions. Some of the work for generics is for our flexibility on the SDK side of things to work with data sets, but the consumer just needs to know about what to render

serikjensen (approved):

.ladle/adapters/PlainComponentAdapter.tsx
jeffredodd: Yeah I think the reason I went with this approach was I wanted the column definitions to be able to be strongly typed. I'm willing to revisit this if you think there is a better approach.

PR #281, opened 2025-04-26
feat: add adapter for menu

serikjensen:
This PR adds an adapter for the Menu component
* It updates to have an isolated Menu component that can be swapped by consumers
* It creates a HamburgerMenu component to centralize the Menu presets in a convenient location for dev use

The HamburgerMenu is an implementation detail of our codebase, so I created a new internal facing set of adapters for abstractions like this that are just compositions of the component adapters with convenient presets for our code.

## Proof of functionality

Updated Menu
Screenshot 2025-04-25 at 6 25 32 PM

Created a chakra UI adapter as well as material to make sure that the menu approach would be compatible

https://github.com/user-attachments/assets/499646c6-93d9-4a7b-a26f-3290bb371ce8

Working swaps in our codebase

https://github.com/user-attachments/assets/039046c6-e48a-4f2f-9113-cc9d8f9aa411

https://github.com/user-attachments/assets/32c6a91c-864d-4650-9722-ddf990684e93

https://github.com/user-attachments/assets/e40b3e03-3870-465a-b24c-ad3b758d31ab

https://github.com/user-attachments/assets/6bbaca5d-7d6e-4546-a964-3b3ebc0ba56a

https://github.com/user-attachments/assets/8f7a3693-d636-4dca-9d19-aaf29272e225


.ladle/adapters/PlainComponentAdapter.tsx
```diff
@@ -16,7 +16,9 @@ import type {
 } from '../../src/components/Common/UI/Button/ButtonTypes'
 import type { LinkProps } from '../../src/components/Common/UI/Link/LinkTypes'
 import type { ComponentsContextType } from '@/contexts/ComponentAdapter/useComponentContext'
+import type { MenuProps } from '@/components/Common/UI/Menu/MenuTypes'
 
+// Instead of using require, define a type-safe adapter for Material UI menu
```
jeffredodd, line 21: AI cruft :-P. I actually ask AI to remove bad comments now as a last step. Seems like it thinks this is material UI.

src/components/Common/UI/HamburgerMenu/HamburgerMenu.tsx
```diff
@@ -0,0 +1,37 @@
+import { useTranslation } from 'react-i18next'
+import { type HamburgerMenuProps } from './HamburgerMenuTypes'
+import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentContext'
+import HamburgerIcon from '@/assets/icons/hamburger.svg?react'
+import { useMenu } from '@/components/Common/UI/Menu/useMenu'
+
+export function HamburgerMenu({
```
jeffredodd, line 7: I'm up for being wrong here, but is HamburgerMenu a menu we think about internally where we're providing the icon button and the menu? I wonder if this belongs in the components/Common folder but not the UI folder.

src/components/Common/UI/Menu/Menu.module.scss
```diff
@@ -0,0 +1,52 @@
+.menu {
```
jeffredodd, line 1: Do we need to do `root` here like we've been doing in the other components?

src/components/Common/UI/HamburgerMenu/HamburgerMenu.tsx
serikjensen: Yeah i think moving it to the root of Common makes sense, that also means it would be subject to the linting we put in place to avoid accidentally importing react aria for stuff like this

jeffredodd (approved):
LGTM

PR #283, opened 2025-04-28
Masking transformer

dmortal:
Introducing global transformer for input masking
Adding input masks to state taxes

https://github.com/user-attachments/assets/ae6d2ce9-4a12-403d-8f71-fcbb9756b5d0


src/components/Company/Locations/LocationForm/Form.tsx
```diff
@@ -18,6 +19,7 @@ export type LocationFormInputs = v.InferInput<typeof LocationFormSchema>
 
 export function Form() {
   const { t } = useTranslation('Company.Locations')
+  const transform = useMaskedTransform('###-###-#####')
```
serikjensen, line 22: Nit: should we also update our existing phone transform function to use this as well here? 

https://github.com/Gusto/embedded-react-sdk/blob/0d2fdabe5a2ff1f46cef51e0a1e0bbd7b5a4ee81/src/helpers/phone.ts#L3-L18

that way it's centralized and consistent?

src/helpers/mask.ts
```diff
@@ -0,0 +1,94 @@
...
+    maskIndex++
+  }
+
+  return result.trim()
+}
+
+export const useMaskedTransform = (mask: string | null): Transform<string> | undefined => {
+  if (!mask) return undefined
+  return (value: string) => formatWithMask(value, mask)
+}
```
serikjensen, line 94: Nit: this doesn't seem to be using another hook, maybe it's sufficient to just use the `formatWithMask` function directly and not provide a hook option since the function will be more flexible?

src/helpers/mask.ts
```diff
@@ -0,0 +1,94 @@
...
+ *   ^ represents an uppercase letter ([A-Z])
+ *   % represents a digit or uppercase letter ([0-9A-Z])
+ *   any other character represents the literal character
+ * @returns The formatted string according to the mask
+ * @example
+ * formatWithMask('123456789', '###-##-####') // returns '123-45-6789'
+ * formatWithMask('ABC123', '@@@-###') // returns 'ABC-123'
+ * formatWithMask('123456', '(###) ###-####') // returns '(123) 456'
+ * formatWithMask('ABC123', '^^^-###') // returns 'ABC-123'
+ * formatWithMask('A1B2C3', '%%%-%%%') // returns 'A1B-2C3'
```
serikjensen, line 18: Very cool!

serikjensen (approved):
Looks good! wondering if we should follow up for existing ssn inputs to migrate them to use this as well

dmortal:
@serikjensen I don't think this should replace ssn because we do more there - we check for specific values in ssn transform

src/helpers/mask.ts
dmortal: I didn't want to break useField api -> so would need either a "createTransform" function or a hook anyways

src/components/Company/Locations/LocationForm/Form.tsx
dmortal: done

serikjensen (approved):

PR #284, opened 2025-05-01
feat: adding support for minimum wage requirements in WA state

dmortal:
Adding worker compensation coverage information for employee compensation component (Washington.state)
https://github.com/user-attachments/assets/7a581c59-1e90-448b-9836-c458004ad687

### DIFF
--- a/src/components/Employee/Compensation/Edit.tsx
+++ b/src/components/Employee/Compensation/Edit.tsx
@@ -38,7 +38,8 @@ export const Edit = () => {
   } = useFormContext<CompensationInputs>()
   const watchedFlsaStatus = useWatch({ control, name: 'flsaStatus' })
   const watchedStateWcCovered = useWatch({ control, name: 'stateWcCovered' })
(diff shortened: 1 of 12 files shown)


src/models/WA_RISK_CODES.ts
dmortal, line 1: This is a problem - we need to talk about this.

src/components/Employee/Compensation/Edit.tsx
```diff
@@ -17,14 +25,21 @@ export const Edit = () => {
   const format = useNumberFormatter('currency')
   const Components = useComponentContext()
 
+  const stateWcRiskOptions = useMemo(
+    () => WA_RISK_CLASS_CODES.map(({ code, description }) => ({ value: code, label: description })),
+    [],
+  )
```
serikjensen, line 31: Wonder if it's worth it to break this into a dedicated component and lazy load it similar to what we are doing in company industry? probably overkill if we plan on getting the api work done soon. But if not, would be nice to avoid any performance hits just for the washington case

serikjensen (approved):
LGTM!

PR #285, opened 2025-05-01
chore: move Menu to UI and HamburgerMenu to common

serikjensen:
Small oversight to my Menu implementation. We had decided to move Menu to UI and HamburgerMenu to the root of Common so that it can get linting rules as necessary in the future.

The UI directory components can potentially introduce things like React Aria that we want to keep from consumers if they are using adapter to avoid bringing them in. HamburgerMenu is an abstraction that puts together components that consumers already have adapters for and shouldn't need to override. By moving it to the components directory we can subject it to the same restrictions as other components re: avoiding importing UI components directly to avoid inadvertently bringing in something like React Aria


src/contexts/ComponentAdapter/adapters/internalComponentAdapter.tsx
```diff
@@ -1,5 +1,5 @@
-import { HamburgerMenu } from '@/components/Common/UI/HamburgerMenu'
-import type { HamburgerMenuProps } from '@/components/Common/UI/HamburgerMenu/HamburgerMenuTypes'
+import { HamburgerMenu } from '@/components/Common/HamburgerMenu'
+import type { HamburgerMenuProps } from '@/components/Common/HamburgerMenu/HamburgerMenuTypes'
 import type { InternalComponentsContextType } from '@/contexts/ComponentAdapter/useComponentContext'
 
 export const internalComponents: InternalComponentsContextType = {
```
jeffredodd, line 5: Just curious, why are we doing this again? I know I approved the initial menu but i'm spacing about why we're using internal context. We can chat about it tomorrow during our sync.

jeffredodd (approved):

src/contexts/ComponentAdapter/adapters/internalComponentAdapter.tsx
serikjensen: Yeah let's chat about it today. TL;DR some component abstractions make sense for use in our codebase that put together multiple components that have adapters. We don't want consumers to have to provide defaults for these abstractions since they are implementation details that are SDK specific and only compose components that they already have registered adapters for.

talking this through, another option would be to just import hamburger menu directly from our codebase instead of getting it through context

PR #288, opened 2025-05-05
chore: surface component adapter interfaces

serikjensen:
This PR updates to surface exports needed to implement the adapter pattern locally. This includes exposing exports for
* ComponentContextType
* All UI component types
* GustoProvider and GustoProviderCustomUIAdapter

This also takes care of updating HamburgerMenu to just get imported from the common directory instead of implementing it as an internal component for component context. This internal component implementation was not necessary since we can just import these components from the common directory instead of getting them from context.


jeffredodd (approved):
LGTM

PR #289, opened 2025-05-06
chore: update input styles and add adornments

serikjensen:
This PR updates input styles as outlined in the following figma https://www.figma.com/design/6dxOSiONDiJoa9zY1wOwbs/Frontend-SDK-Partner-Design-File?node-id=8055-1948&t=tQg2O6DDi0BrrBGM-4

* Updates styles for focus outline
* Places error icon inline with the inputs
* Adds start and end adornment options

This also integrates the start and end adornments with the NumberInput so that it can display currency and percentage more aligned with design requirements.

### Tax forms still working as before

https://github.com/user-attachments/assets/f083eb0b-7cc1-4b47-b933-6c8dbf177dc6

### Deductions forms still working as before

https://github.com/user-attachments/assets/87798408-c973-4553-9a23-542a10a9dcef

### Ladle instances

https://github.com/user-attachments/assets/0a4008ab-2a62-4e26-b02f-1061fe114516


src/components/Common/UI/ComboBox/ComboBox.module.scss
```diff
@@ -6,12 +6,19 @@
   :global(.react-aria-ComboBox-root) {
     color: var(--g-typography-text-color);
     width: 100%;
+    box-shadow: var(--g-shadow-100);
+
+    &[data-invalid] {
```
jeffredodd, line 11: Thanks for cleaning this up!

src/components/Common/UI/NumberInput/NumberInput.module.scss
```diff
@@ -37,28 +28,24 @@
...
-    outline: none;
-    flex: 1;
-
-    &[data-disabled] {
-      border-color: var(--g-input-disabled-border);
-      color: var(--g-input-disabled-color);
-      background-color: var(--g-input-disabled-bg);
-    }
-  }
+  // :global(.react-aria-Input) {
```
jeffredodd, line 34: Do we need to keep this?

serikjensen: nope, was experimenting to make sure i could remove the overrides. will nix this!

src/components/Common/UI/Input/Input.module.scss
```diff
@@ -0,0 +1,122 @@
+@use '@/styles/Helpers' as *;
+
```
dmortal, line 2: all of our css definitions should be wrapped in `:global(.GSDK) {` for scoping purposes

src/components/Common/UI/Input/Input.module.scss
```diff
@@ -0,0 +1,122 @@
...
+    background-color: var(--g-input-disabled-bg);
+    color: var(--g-input-disabled-color);
+
+    .adornmentStart,
+    .adornmentEnd {
+      color: var(--g-input-disabled-color);
+    }
+  }
+
+  :global(.react-aria-Input) {
```
dmortal, line 37: I forgot, but won't this affect other this class styling outside sdk components?

src/components/Common/UI/NumberInput/NumberInput.tsx
```diff
@@ -38,6 +41,15 @@ export function NumberInput({
   const minValue = typeof min === 'string' ? Number(min) : min
   const maxValue = typeof max === 'string' ? Number(max) : max
 
+  const currencySymbol = useMemo(() => {
```
dmortal, line 44: 🥹

src/components/Common/UI/NumberInput/NumberInput.tsx
```diff
@@ -75,13 +90,15 @@ export function NumberInput({
       >
         <Group>
           <Input
+            adornmentStart={adornmentStart || (format === 'currency' ? currencySymbol : null)}
+            adornmentEnd={adornmentEnd || (format === 'percent' ? '%' : null)}
```
dmortal, line 94: how would we achieve the following? 
image

src/components/Common/UI/Input/Input.module.scss
serikjensen: all global styles are scoped within the module `.container` class at the top level so this would only be applied to instances of `.react-aria-Input` that occur within the `.container` module class set on the Input component

I know these end up looking kinda wonky/counterintuitive since it ends up being escaped with the `:global` prefix, but this is actually only applied to module styles

src/components/Common/UI/NumberInput/NumberInput.tsx
serikjensen: So for those cases you'd end up setting the `adornmentEnd` property on `NumberInput`. I added a story for that. But you can do

```
      <NumberInput
        label="Pay rate"
        name="pay-rate"
        value={value}
        onChange={handleChange}
        format="currency"
        adornmentEnd="/hr"
      />
```

which then results in

Screenshot 2025-05-07 at 10 51 49 AM

NumberInput will set a default for `adornmentStart` (currency symbol) if it's currency, and a default for `adornmentEnd` (if it's percentage) but if you have a special case you can still override those defaults with the `adornmentStart` and `adornmentEnd` props as desired

dmortal (approved):

PR #292, opened 2025-05-07
feat: add Heading component and implement adapter

serikjensen:
This PR adds a new `Heading` component. The `Heading` component centralizes our heading styles and allows for decoupling of the semantic heading html from the heading styles. For example, you can now render an h2 styled as an h4 if that meets the demands of the heading level .

This also implements an adapter for the heading components, and updates heading instances throughout the codebase (both React Aria `Heading` component and `h1`-`h6`) to use that adapter.

Note: it didn't always make sense to update to use the `Heading` component. For example, for headings that live within a table, table will have specific styling needs that likely diverge from the system. This would also introduce theming issues where the core Heading component styles would conflict with the component level styles. In these cases, it's sufficient to just use the `h*` component directly and provide component specific theme variables.

### Heading styles

https://github.com/user-attachments/assets/b54625ac-b53f-4568-afd6-b736c799f513

### Smoke test of existing employee flows to use updated heading

https://github.com/user-attachments/assets/3b8d8bb4-371b-4ec8-ad63-14fdf32cb320

### DIFF
--- a/.ladle/adapters/PlainComponentAdapter.tsx
+++ b/.ladle/adapters/PlainComponentAdapter.tsx
@@ -22,6 +22,7 @@ import type { ComponentsContextType } from '@/contexts/ComponentAdapter/useCompo
 import type { MenuProps } from '@/components/Common/UI/Menu/MenuTypes'
 import type { BreadcrumbsProps } from '@/components/Common/UI/Breadcrumb'
 import type { TableProps } from '@/components/Common/UI/Table'
+import type { HeadingProps } from '@/components/Common/UI/Heading/HeadingTypes'
 
 export const PlainComponentAdapter: ComponentsContextType = {
   Alert: ({ label, children, status = 'info', icon }: AlertProps) => {
@@ -981,5 +982,25 @@ export const PlainComponentAdapter: ComponentsContextType = {
         ))}
       </ul>
     )
+  },
+
+  Heading: ({ as: Component, styledAs, textAlign, children }: HeadingProps) => {
+    const levelStyles = styledAs ?? Component
+
+    const fontSizes = {
+      h1: '2rem',
+      h2: '1.5rem',
+      h3: '1.25rem',
+      h4: '1rem',
+      h5: '0.875rem',
+      h6: '0.75rem',
+    }
+
+    const headingStyles = {
+      textAlign: textAlign,
+      fontSize: fontSizes[levelStyles],
+    }
+
+    return <Component style={headingStyles}>{children}</Component>
   },
 }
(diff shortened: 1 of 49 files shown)


src/components/Common/UI/DatePicker/DatePicker.module.scss
```diff
@@ -258,6 +258,7 @@
     color: var(--g-colors-gray-1000);
     font-weight: 400;
     font-size: var(--g-typography-fontSize-small);
+    overflow-x: hidden;
```
serikjensen, line 261: Small datepicker fix for showing the scroll bar that i missed when doing my input update

jeffredodd: Thank you!

src/components/Common/DataView/DataCards/DataCards.tsx
```diff
@@ -37,9 +36,7 @@ export const DataCards = <T,>({
           >
             {columns.map((column, index) => (
               <Flex key={index} flexDirection="column" gap={0}>
-                <Heading className={styles.columnTitle} level={5}>
-                  {column.title}
-                </Heading>
+                <h5 className={styles.columnTitle}>{column.title}</h5>
```
jeffredodd, line 39: Do we want to use the component adapter heading here?

jeffredodd (approved):
Just one small comment but it looks good

.ladle/adapters/PlainComponentAdapter.tsx
```diff
@@ -982,4 +983,24 @@ export const PlainComponentAdapter: ComponentsContextType = {
       </ul>
     )
   },
+
+  Heading: ({ as: Component, styledAs, textAlign, children }: HeadingProps) => {
+    const levelStyles = styledAs ?? Component
+
+    const fontSizes = {
+      h1: '2rem',
```
dmortal, line 991: `rems` -> these will default to root font size, which we don't know. We need to use `toRem` here

dmortal (approved):

.ladle/adapters/PlainComponentAdapter.tsx
serikjensen: This one is just ladle specific so doesn't actually get used by consumers :D

src/components/Common/DataView/DataCards/DataCards.tsx
serikjensen: Hello! yes see note in PR description

> it didn't always make sense to update to use the Heading component. For example, for headings that live within a table, table will have specific styling needs that likely diverge from the system. This would also introduce theming issues where the core Heading component styles would conflict with the component level styles. In these cases, it's sufficient to just use the h* component directly and provide component specific theme variables.

There's a distinction between heading being used in isolation as a component vs heading as an internal part of a system level component. In this case the cards have their own theme variables. We'd need to find a way for Heading to access those theme variables if we wanted to use it here. but honestly, at that point, heading isn't really giving us anything since all of it's shared styles (font size, etc) are getting overwritten by table needs anyway.

I think this is going to be a distinction that will also apply to text elements more generally. They are kind of a different flavor of reusable component.

PR #295, opened 2025-05-08
feat: introduce internal visually hidden component

jeffredodd:
This PR introduces an internal VisuallyHidden component in order to reduce our dependency on react-aria.


serikjensen (approved):
LGTM!

PR #297, opened 2025-05-09
chore: add lint rule to restrict UI imports

serikjensen:
This adds a lint rule to restrict imports from the UI directory. We should not be importing components from this directory directly (outside of the default component adapter which is only used in the GustoProvider) since these components import React Aria which we want to allow to be omitted via tree shaking.

This still allows type imports from the UI directory, since those types are necessary for defining interfaces and can potentially be reused in our codebase. To achieve this, imports of files ending in `Types` are valid.

## Proof of functionality

Example import of `Card` from UI when working in the common directory along with error message
Screenshot 2025-05-09 at 4 01 15 PM

Also working for relative imports
Screenshot 2025-05-09 at 4 02 09 PM

Valid import since it is an import of types
Screenshot 2025-05-09 at 4 02 48 PM


eslint.config.mjs
```diff
@@ -93,12 +93,19 @@ export default [
...
+              message:
+                "Please use the useComponentContext hook instead: import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentContext' and get the component from that hook.",
+            },
+          ],
         },
       ],
     },
   },
   {
-    files: ['src/components/Common/Inputs/*.tsx', 'src/components/Common/UI/**/*.tsx'],
```
serikjensen, line 101: The `Common/Inputs` dir was already deleted with the hook form work so it can be safely removed here as well

src/hooks/useContainerBreakpoints/useContainerBreakpoints.stories.tsx
```diff
@@ -1,5 +1,4 @@
 import React from 'react'
-import { type ButtonProps as _ButtonProps, Button as _Button } from 'react-aria-components'
```
serikjensen, line 2: Unused import caught here by looking ahead to further react aria separation which could be deleted

PR #299, opened 2025-05-12
feat: add component adapter for calendar

serikjensen:
This PR implements an adapter for the CalendarDisplay component

* Moves CalendarDisplay to the UI directory
* Restores calendar styles that had been deleted by mistake
* Updates name to be `Calendar` instead
* Updates the Calendar API to use dates consistent with DatePicker instead of strings
* Moves the `selectionControl` node prop out of the component for ease of consumers wiring it up
* Implements an adapter for the `Calendar`

## Proof of functionality

https://github.com/user-attachments/assets/1f2c7218-24ef-49e1-ae95-cbdb9570826e


src/components/Common/UI/Calendar/Calendar.module.scss
```diff
@@ -0,0 +1,258 @@
+:global(.GSDK) {
```
serikjensen, line 1: I accidentally deleted this stylesheet when cleaning up date picker. This restores it. It implements the same styles found here https://github.com/Gusto/embedded-react-sdk/pull/184

src/components/Common/UI/Calendar/Calendar.stories.tsx
```diff
@@ -0,0 +1,93 @@
+import type { CalendarProps } from './CalendarTypes'
+import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentContext'
```
serikjensen, line 2: This file was just moved from the original location to the UI directory, it's not net new. Compare to https://github.com/Gusto/embedded-react-sdk/blob/main/src/components/Common/CalendarDisplay/CalendarDisplay.stories.tsx

src/components/Common/UI/Calendar/CalendarTypes.ts
```diff
@@ -0,0 +1,12 @@
+export type CalendarProps = {
+  highlightDates?: Array<{
+    date: Date
```
serikjensen, line 3: Updated to use `Date` on these instead of `string` to be more clear with consumers using this API what data they are working with. This also makes this consistent with `DatePicker` which uses `Date`

dmortal (approved):

PR #302, opened 2025-05-14
documentSignerFlow fix

dmortal:
This Pr changes default behavior of deocumentsignerflow if signatory exists - instead of taking user to the assign signatory form it directs them to document list, while still allowing to change signatory if they want.

https://github.com/user-attachments/assets/5945b54f-5bbf-4fba-9645-1be9eec9e27d


serikjensen (approved):
LGTM!

PR #304, opened 2025-05-14
feat: replacing breadcrumbs with progress bar

dmortal:
Replacing breadcrumbs with ProgressBar!

https://github.com/user-attachments/assets/eed57895-169f-49dd-be85-01cf3b1905c8


src/components/Common/UI/ProgressBar/ProgressBar.module.scss
```diff
@@ -0,0 +1,31 @@
...
+      border-radius: 9999px;
+      content: '';
+      display: block;
+      height: 100%;
+      width: var(--g-progress-bar-width);
+      background-color: var(--g-colors-gray-900);
+      transition: width 0.5s ease-out;
+    }
+    //Hiding the progress bar from the screen
+    progress {
```
dmortal, line 18: is this fine from accessibility standpoint?

serikjensen: Yeah so long as you're not using display none this should keep it visually hidden while it can still get read for assistive tech

You might also consider just using the `VisuallyHidden` component in `src/components/Common` and wrapping this `progress` element in that since that already maintains these styles.

src/components/Common/UI/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,15 @@
...
+import styles from './ProgressBar.module.scss'
+
+export function ProgressBar({ className, totalSteps, currentStep, label }: ProgressBarProps) {
+  const clampedStep = Math.max(1, Math.min(currentStep, totalSteps))
+  const progressBarStyle = {
+    '--g-progress-bar-width': `${(clampedStep * 100) / totalSteps}%`,
+  } as React.CSSProperties
+  return (
+    <div className={classnames(styles.root, className)} style={progressBarStyle}>
+      <progress aria-label="progress bar" value={clampedStep} max={totalSteps}></progress>
```
serikjensen, line 12: Whoops! i assume we actually meant to do

```suggestion
      <progress aria-label={label} value={clampedStep} max={totalSteps}></progress>
```

serikjensen (commented):
Looking good! one comment to update the ProgressBar to use the label from props instead of the hard coded placeholder

PR #305, opened 2025-05-14
chore: experiment building with tsc only

serikjensen:
Prototype for building with TSC and CSS modules.

* runs tsc to go from ts -> js
* converts sass to css
* converts css to css modules using the postcss plugin
* then runs a babel transform on the output js in dist to get it to point the style imports to the generated json files from postcss

```
import styles from 'Flex.module.scss'
```

becomes

```
import styles from 'Flex.module.scss.json'
import 'Flex.module.css'
```

The styles json file is a mapping of the component classnames to the hashed values. Ex. 

```
{"flexContainer":"Flex-module__flexContainer___7bwGv","flex":"Flex-module__flex___-UxXT"}
```

We then do the plain css import so that the css file only gets imported if the module gets used.

The babel transform for converting the js imports isn't extremely involved, but it was pretty frustrating to have to reach for that. There are existing babel transforms for postcss modules, but those have not seen support in a long time. They also don't provide the css import which I thought would be beneficial for tree shaking consumers to only bring in the styles that they need. Ultimately we could get around that overhead by just outputting a single css file and having consumers import it similar to what they do today.

Even still, existing plugins for css modules haven't been updated for years. Ex. [this one](https://www.npmjs.com/package/babel-plugin-css-modules-transform) does what we need, but it hasn't been updated for 6 years and conflicts with the latest versions of postcss.

## Conclusions

This falls under a category of technically possible, but was a much longer chain of tooling and configuration than was desirable. Maintaining the babel transform is much lower level than anticipated and it was disappointing that existing plugins for that were outdated and conflicting with later postcss versions.

There are some notable drawbacks here, and if we decide to go purely with TSC as our build tool, I would recommend exploring CSS solutions that do not need JS build tooling. That would lead us to either global CSS (not worrying about naming collisions) or utility classes like tailwind.


package.json
```diff
@@ -28,7 +28,8 @@
     "./CHANGELOG.md"
   ],
   "scripts": {
-    "build": "npm run build:clean && npm run i18n:generate && vite build",
+    "build": "npm run build:clean && npm run i18n:generate && tsc && tsc-alias && npm run build:styles",
+    "build:styles": "postcss \"src/**/*.module.scss\" --base src --dir dist --verbose",
```
scally, line 32: Why not use `sass-embedded`? It's already in the package, so it would remove the need for postcss here

postcss.config.cjs
```diff
@@ -0,0 +1,25 @@
+module.exports = {
+  syntax: 'postcss-scss',
+  plugins: [
+    require('postcss-modules')({
+      generateScopedName: '[name]__[local]___[hash:base64:5]',
```
scally, line 5: why do we need to customize the scoped name here? could we somehow reuse an existing algorithm postcss already has for naming things?

postcss.config.cjs
```diff
@@ -0,0 +1,25 @@
...
+        const jsonFileName = outputFileName.replace(/^src/, 'dist') + '.json'
+
+        // Ensure the directory exists
+        const dir = path.dirname(jsonFileName)
+        if (!fs.existsSync(dir)) {
+          fs.mkdirSync(dir, { recursive: true })
+        }
+
+        // Write the JSON to a file
+        fs.writeFileSync(jsonFileName, JSON.stringify(json))
```
scally, line 21: does this need to correspond to a change to all our imports to be able to import this json file to reference class names?

serikjensen: That's the next step i'll be investigating this morning

serikjensen:
Closing as we are seeing more potential going with a pure vite configuration to take care of this

PR #306, opened 2025-05-16
Da/build

dmortal:
**Build process upgrade summary:**
This PR disables vite bundling of our package, enabling tree shakable code while retaining use of css modules
* Introduces development mode, disabling certain plugins and configs in local watch mode for faster rebuild(vite has to rebuild entire package in lib mode on every change since it is using rollup in that mode)
* Removes unused devDeps, including `vite-plugin-sass-dts` that is not needed since we are bundling all css into a single file
* Adds react-query dependency - followup PR is expected that will change how we instantiate speakeasy
* Declares side-effects - needed to avoid incorrect dependency bundling
* Explicitly externalizes all our deps


vite.config.ts
```diff
@@ -3,75 +3,94 @@ import react from '@vitejs/plugin-react-swc'
...
 import stylelint from 'vite-plugin-stylelint'
 import svgr from 'vite-plugin-svgr'
 import circularDependencyDetector from 'vite-plugin-circular-dependency'
 import checker from 'vite-plugin-checker'
+import { externalizeDeps } from 'vite-plugin-externalize-deps'
 
-export default defineConfig({
-  plugins: [
-    react(),
-    sassDts({
```
dmortal, line 15: we don't have any scss files in `./dist` so this plugin is useless

vite.config.ts
```diff
@@ -3,75 +3,94 @@ import react from '@vitejs/plugin-react-swc'
...
+    plugins: [
+      react(),
+      externalizeDeps(), // Externalizes all dependencies
+      !isDev &&
+        dts({
+          include: ['src'],
+          outDir: './dist',
+          tsconfigPath: './tsconfig.json',
+          insertTypesEntry: true,
+          rollupTypes: false,
```
dmortal, line 24: Disabling `rollupTypes` - > there are path resolution issues in rolled up file - with this disabled, each module has it's own `.d.ts` file

vite.config.ts
```diff
@@ -3,75 +3,94 @@ import react from '@vitejs/plugin-react-swc'
...
+          exclude: [
+            '**/node_modules/**',
+            '**/.ladle/**',
+            '**/*.stories.tsx',
+            '**/*.test.tsx',
+            '**/test/**',
+          ],
+          strictOutput: true,
+        }),
+      !isDev && stylelint({ fix: true }),
```
dmortal, line 35: no need in dev mode

vite.config.ts
```diff
@@ -3,75 +3,94 @@ import react from '@vitejs/plugin-react-swc'
...
-    rollupOptions: {
-      input: {
-        main: resolve(__dirname, 'src/index.ts'),
+    build: {
+      lib: {
+        fileName: 'index',
+        entry: resolve(__dirname, 'src/index.ts'),
+        formats: ['es'],
+      },
+      minify: !isDev,
```
dmortal, line 69: disabling in dev for faster compile

vite.config.ts
```diff
@@ -3,75 +3,94 @@ import react from '@vitejs/plugin-react-swc'
...
-    svgr({
-      svgrOptions: {
-        exportType: 'default',
-        titleProp: true,
+export default defineConfig(({ mode }) => {
+  const isDev = mode === 'development'
+  return {
+    plugins: [
+      react(),
+      externalizeDeps(), // Externalizes all dependencies
```
scally, line 17: This is cleaner than what we had before, it's a good win

scally: (how we externalize dependencies, I mean)

scally:
Couple of questions I'd like to know more about, but seems good overall 👍🏻

vite.config.ts
```diff
@@ -3,75 +3,94 @@ import react from '@vitejs/plugin-react-swc'
...
+      minify: !isDev,
+      sourcemap: !isDev,
+      cssCodeSplit: false, //Force vite to generate single css file
+      rollupOptions: {
+        input: resolve(__dirname, 'src/index.ts'),
+        output: {
+          preserveModules: true, //Maintains the original file structure
+          preserveModulesRoot: 'src', //Removes the root folder from the output
+          dir: 'dist',
+          entryFileNames: '[name].js', //Retains the original file name
```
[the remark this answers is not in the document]
dmortal: yes, 
image

PR #310, opened 2025-05-19
docs: component adapter documentation

jeffredodd:
Initial working draft of Component Adapter documentation.


docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### How can I ensure my custom components maintain accessibility features?
+
+When implementing custom components, pay attention to:
+
+- Proper labeling of form controls
+- Appropriate ARIA attributes
+- Keyboard navigation
+- Focus management
+- Color contrast
```
serikjensen, line 47: I'm wondering if this actually needs to be more of a disclaimer that if they are using the adapter, they will own the accessibility for custom components and pointing them to some generally a11y resources

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+
+The prop interfaces include properties like `aria-describedby`, `isInvalid`, and others that support accessibility. Make sure your custom implementations use these props correctly.
```
serikjensen, line 49: Hmm this is a good point, when we supply those props they'll need to know how they are wired up.

I'm wondering if for each of our adapter components, we need to export a code sample demonstrating how to wire it up with native html. We already have essentially implemented these for plain component adapter for demo purposes. But if we provided a basic example with native html it could be more clear. There's a lot of nuances with some of the aria attributes that might be easier to communicate with a code example

Let me think on it with some of the auto generated stuff

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+
+The SDK components follow accessibility best practices by default. When creating custom implementations, study the default components to understand how they handle accessibility concerns. The SDK uses [React Aria](https://react-spectrum.adobe.com/react-aria/) for accessible components, which you can reference for implementation details.
```
serikjensen, line 51: Pointing them to the default examples here may not be as helpful for a11y insights since most are just contained in react aria/and/or our own custom workarounds for dealing with react aria

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+For complex components, you have a few options:
+
+1. Use a third-party library that provides similar functionality
+2. Implement a simplified version that meets your specific needs
+3. Use the SDK's default implementation for complex components while customizing simpler ones
+
+For example, to implement a DatePicker with react-datepicker:
+
+```tsx
+import DatePicker from 'react-datepicker'
```
serikjensen, line 66: I'm wondering in this case if it's sufficient to just point them to our default adapter for DatePicker given we already are demonstrating how to wire it up with react aria and if we link to the actual file it will be ensured to be up to date

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+      {isInvalid && errorMessage && <div className="error">{errorMessage}</div>}
+    </div>
+  )
+}
+```
+
+To understand the expected behavior of complex components, refer to the SDK's implementations:
+
+- [ComboBox on GitHub](https://github.com/Gusto/embedded-react-sdk/tree/main/src/components/Common/UI/ComboBox)
+- [DatePicker on GitHub](https://github.com/Gusto/embedded-react-sdk/tree/main/src/components/Common/UI/DatePicker)
```
serikjensen, line 110: Yeah these might be sufficient vs. the code snippet above

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+
+#### Can I use a different UI framework for my custom components?
+
+Yes, you can use any UI framework or library for your custom components, as long as they correctly implement the required props and behaviors. For example, you could use Material UI, Chakra UI, or any other React-compatible UI library.
```
serikjensen, line 114:
```suggestion
Yes, you can use any React UI framework or library for your custom components, as long as they correctly implement the required props and behaviors. For example, you could use Material UI, Chakra UI, or any other React-compatible UI library.
```

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+  )
+}
+```
+
+To understand the expected behavior of complex components, refer to the SDK's implementations:
+
+- [ComboBox on GitHub](https://github.com/Gusto/embedded-react-sdk/tree/main/src/components/Common/UI/ComboBox)
+- [DatePicker on GitHub](https://github.com/Gusto/embedded-react-sdk/tree/main/src/components/Common/UI/DatePicker)
+
+#### Can I use a different UI framework for my custom components?
```
serikjensen, line 112: Might be worth bumping this one to the top?

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+5. **Debug with React DevTools**: Use React DevTools to inspect the props being passed to your components and compare with what you're expecting.
+
+### Troubleshooting
+
+#### I'm getting errors about missing components. What's wrong?
+
+If you're seeing errors about missing components, it's likely that you haven't provided an implementation for all the components the SDK is trying to use. Make sure your component adapter includes all the components used by the SDK, or use the `defaultComponents` to provide fallbacks.
+
+```tsx
+import { defaultComponents } from '@gusto/embedded-react-sdk'
```
serikjensen, line 137: Similar to above, wondering if we should be recommending them to use GustoProvider if they want to use our defaultComponents since the GustoProvider provides the fallback components from defaultComponents automatically

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+  Button: props => <MyCustomButton {...props} />,
+  TextInput: props => <MyCustomTextInput {...props} />,
+}
+```
+
+### Design and Styling
+
+#### What's the difference between using the Component Adapter and just overriding the CSS?
+
+The Component Adapter gives you full control over the implementation of UI components, including their structure, behavior, and styling. Overriding CSS only allows you to modify the appearance of the default components.
```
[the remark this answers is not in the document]
jeffredodd: I think I'm going to remove this, I'm not sure we need to direct customers in themeing in this doc. If they are coming for the component adapter it should be because they are implementing a design system and theming would relate to that instead.

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+const myAdapter = {
+  ...defaultComponents,
+  Button: props => <MyCustomButton {...props} />,
+  TextInput: props => <MyCustomTextInput {...props} />,
+}
+```
+
+### Design and Styling
+
+#### What's the difference between using the Component Adapter and just overriding the CSS?
```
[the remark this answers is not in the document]
jeffredodd: Agreed

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+
+```tsx
+import { defaultComponents } from '@gusto/embedded-react-sdk'
+
+const myAdapter = {
+  ...defaultComponents,
+  Button: props => <MyCustomButton {...props} />,
+  TextInput: props => <MyCustomTextInput {...props} />,
+}
+```
```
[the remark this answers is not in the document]
jeffredodd: Agreed

docs/06/01/faq.md
jeffredodd: Let me know if you like this update

docs/06/01/faq.md
jeffredodd: Done

docs/06/01/faq.md
jeffredodd: Updated

docs/06/01/faq.md
```diff
@@ -0,0 +1,177 @@
...
+
+You can check which components are being used by examining the SDK's source code or by adding console logs to your adapter implementation.
+
+#### My form values aren't being captured correctly. What might be wrong?
+
+This often happens when the `onChange` handler in your custom form components isn't being called with the correct parameters. The SDK expects specific value formats from each component's onChange handler:
+
+- Checkbox: `onChange(boolean)`
+- DatePicker: `onChange(Date | null)`
+- NumberInput: `onChange(number | string)`
```
serikjensen, line 154: This one can actually just be

```suggestion
- NumberInput: `onChange(number)`
```

docs/06/01/how-it-works.md
```diff
@@ -0,0 +1,73 @@
+## How the Component Adapter Works
+
+The Component Adapter uses React's Context API to provide UI components to the SDK. At its core, the system consists of:
+
+1. **ComponentsContext** - A React context that holds references to all UI components ([View on GitHub](https://github.com/Gusto/embedded-react-sdk/blob/main/src/contexts/ComponentAdapter/useComponentContext.ts))
+2. **ComponentsProvider** - A provider component that makes custom UI components available throughout the component tree ([View on GitHub](https://github.com/Gusto/embedded-react-sdk/blob/main/src/contexts/ComponentAdapter/ComponentsProvider.tsx))
+3. **useComponentContext** - A hook to access the components within SDK components ([View on GitHub](https://github.com/Gusto/embedded-react-sdk/blob/main/src/contexts/ComponentAdapter/useComponentContext.ts))
+
```
serikjensen, line 8: Wondering if we should actually just omit this? worried about trying to keep these implementation details up to date if we ever update our approach

docs/06/01/how-it-works.md
```diff
@@ -0,0 +1,73 @@
...
+│  │ └─────────────┘ │    │
+│  │                 │    │
+│  │ SDK Components  │    │
+│  └─────────────────┘    │
+│                         │
+└─────────────────────────┘
+```
+
+1. You create custom components that implement the required interfaces
+2. You provide these components through the `GustoProviderCustomUIAdapter` ([View on GitHub](https://github.com/Gusto/embedded-react-sdk/blob/main/src/contexts/GustoProvider/GustoProviderCustomUIAdapter.tsx))
```
serikjensen, line 37: I'm wondering if we need to document the distinction between `GustoProviderCustomUIAdapter` and `GustoProvider`? components can be passed to either one. `GustoProvider` will likely be more common unless they have a real dedication to tree shaking?

docs/06/01/setup-guide.md
```diff
@@ -0,0 +1,255 @@
...
+  Table: props => <MyCustomTable {...props} />,
+  Text: props => <MyCustomText {...props} />,
+  TextInput: props => <MyCustomTextInput {...props} />,
+  UnorderedList: props => <MyCustomUnorderedList {...props} />,
+}
+```
+
+### 3. Use GustoProviderCustomUIAdapter
+
+Instead of using the standard `GustoApiProvider`, use the `GustoProviderCustomUIAdapter` ([View on GitHub](https://github.com/Gusto/embedded-react-sdk/blob/main/src/contexts/GustoProvider/GustoProviderCustomUIAdapter.tsx)) to provide your custom components:
```
serikjensen, line 93:
```suggestion
Instead of using the standard `GustoProvider`, use the `GustoProviderCustomUIAdapter` ([View on GitHub](https://github.com/Gusto/embedded-react-sdk/blob/main/src/contexts/GustoProvider/GustoProviderCustomUIAdapter.tsx)) to provide your custom components:
```

docs/06/01/setup-guide.md
```diff
@@ -0,0 +1,255 @@
...
+  Select: props => <MyCustomSelect {...props} />,
+  Switch: props => <MyCustomSwitch {...props} />,
+  Table: props => <MyCustomTable {...props} />,
+  Text: props => <MyCustomText {...props} />,
+  TextInput: props => <MyCustomTextInput {...props} />,
+  UnorderedList: props => <MyCustomUnorderedList {...props} />,
+}
+```
+
+### 3. Use GustoProviderCustomUIAdapter
```
serikjensen, line 91: I think we need to do a better clarification between GustoProvider and GustoProviderCustomUIAdapter. They can supply custom components with each. With GustoProvider we supply the fallbacks vs. if they want to completely eject they can use GustoProviderCustomUIAdapter

docs/06/01/setup-guide.md
```diff
@@ -0,0 +1,255 @@
...
+      <EmployeeOnboardingFlow companyId="company_123" />
+    </GustoProviderCustomUIAdapter>
+  )
+}
+```
+
+### 4. Implement Only the Components You Need
+
+You don't need to implement every component in the `ComponentsContextType` interface. You can choose to implement only the ones you want to customize and let the default components handle the rest by importing the `defaultComponents` ([View on GitHub](https://github.com/Gusto/embedded-react-sdk/blob/main/src/contexts/ComponentAdapter/adapters/defaultComponentAdapter.tsx)):
+
```
serikjensen, line 114: So for this case we should emphasize that they can just use `GustoProvider` and supply the `components` prop. That's already importing defaultComponents under the hood

docs/06/01/setup-guide.md
```diff
@@ -0,0 +1,255 @@
...
+function App() {
+  return (
+    <GustoProviderCustomUIAdapter config={{ baseUrl: '/api/gusto/' }} components={customComponents}>
+      {/* Your application components */}
+    </GustoProviderCustomUIAdapter>
+  )
+}
+```
+
+> **Important Note**: When using the partial implementation approach (spreading `defaultComponents`), be aware that the default components include React Aria design system dependencies. This can potentially increase your bundle size and prevent React Aria from being properly tree-shaken away, which goes against our optimization goals. For optimal bundle size, consider implementing all required components with your own design system instead of relying on the default implementations.
```
serikjensen, line 134: Same here, this is the disclaimer that should potentially be made for `GustoProvider` usage and why they might consider using `GustoProviderCustomUIAdapter` instead

docs/06/01/setup-guide.md
```diff
@@ -0,0 +1,255 @@
...
+4. Test across different browsers and devices
+
+For examples of testing, you can look at the SDK's test files ([View test examples on GitHub](https://github.com/Gusto/embedded-react-sdk/tree/main/test)).
+
+### Complete Example
+
+Here's a more complete example showing a custom implementation with a UI library (in this case, Material UI):
+
+```tsx
+import { GustoProviderCustomUIAdapter, defaultComponents } from '@gusto/embedded-react-sdk'
```
serikjensen, line 152: This example should probs be updated to use GustoProvider instead

docs/06/01/types.md
```diff
@@ -0,0 +1,159 @@
...
+  Radio: (props: RadioProps) => JSX.Element | null
+  RadioGroup: (props: RadioGroupProps) => JSX.Element | null
+  Select: (props: SelectProps) => JSX.Element | null
+  Switch: (props: SwitchProps) => JSX.Element | null
+  Table: <T>(props: TableProps<T>) => JSX.Element | null
+  Text: (props: TextProps) => JSX.Element | null
+  TextInput: (props: TextInputProps) => JSX.Element | null
+  UnorderedList: (props: UnorderedListProps) => JSX.Element | null
+}
+```
```
serikjensen, line 37: I think where possible we should just provide the links to the docs (like you already have) since this stuff will get out of sync with working code fast. Maybe when we do want to provide a snippet we can just show one or two and then do an abbreviation for the rest

docs/06/01/types.md
```diff
@@ -0,0 +1,159 @@
...
+
+Each component type (e.g., `ButtonProps`, `TextInputProps`) is defined in its respective file and exported from the SDK. These interfaces detail all the props that must be supported by your custom implementations.
+
+### GustoProviderCustomUIAdapterProps
```
serikjensen, line 41: This potentially seems out of place in this doc? I'm not opposed to having it but it may belong in a getting started section where the GustoProvider is already explained?

Feeling also like we should favor just linking to the interface so that it can stay up to date rather than enumerating the values here since the interface is likely to get out of date quickly

docs/06/01/types.md
```diff
@@ -0,0 +1,159 @@
...
+- `config`: Configuration for the API connection
+- `components`: Your custom component adapter implementing `ComponentsContextType`
+- `theme`: (Optional) Custom theming options
+- `dictionary`: (Optional) Translation overrides
+- `lng`: (Optional) Language setting
+- `locale` and `currency`: (Optional) Localization settings
+
+### Component Prop Types
+
+Each UI component has its own prop interface. Here are some of the most commonly used ones:
```
serikjensen, line 73: I think this section could be omitted in favor of just referencing the component inventory which will be autogenerated after the work that i'm doing is implemented

docs/06/01/types.md
```diff
@@ -0,0 +1,159 @@
...
+
+By following the type definitions, you can ensure that your custom components will integrate seamlessly with the SDK's business logic.
+
+### Type Hierarchy
+
+The component types often extend basic HTML element props to provide both the SDK-specific props and standard HTML attributes:
+
+```typescript
+// Simplified example
+export interface TextInputProps extends HTMLInputAttributes {
```
serikjensen, line 149: I think we actually got away from this in favor of only picking specific props? we didn't want there to be a lot of noise and uncertainty around what props the consumer needed to implement so we felt it might be better to just pick the ones we needed. If there are interfaces that extend the HTML attributes we might consider updating those

docs/06/component-adapter.md
```diff
@@ -0,0 +1,13 @@
+## Component Adapter
+
+The Component Adapter system provides a powerful way to customize the UI components used throughout the Gusto Embedded React SDK. This feature allows you to replace the default React Aria-based components with your own custom implementations while maintaining compatibility with the SDK's functionality.
```
serikjensen, line 3:
```suggestion
The Component Adapter system provides a powerful way to customize the UI components used throughout the Gusto Embedded React SDK. This feature allows you to replace the default SDK components with your own custom implementations while maintaining compatibility with the SDK's functionality.
```

docs/06/01/setup-guide.md
jeffredodd: Agreed, see latest update

docs/06/01/setup-guide.md
jeffredodd: Kk, I removed this, I think i've covered it enough above

docs/06/01/types.md
jeffredodd: Updated

docs/06/01/types.md
jeffredodd: Updated, also rewrote the types file to push more stuff into the setup-guide which i agree makes more sense.

docs/06/01/types.md
jeffredodd: Updated

docs/06/component-adapter.md
jeffredodd: Added

PR #312, opened 2025-05-20
chore: release 0.8.0

jeffredodd:
This PR is for our 0.8.0 release. See CHANGELOG for detailed changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,44 @@
 # Changelog
 
+## 0.8.0
+
+- Company Onboarding flow improvements and fixes:
+  - Added comprehensive Company.OnboardingFlow component that guides users through the entire onboarding process
+  - Introduced Company.OnboardingOverview component for tracking onboarding progress
+  - Improved state management and context handling for onboarding components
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,44 @@
...
+
+// After
+<GustoProvider config={{ baseUrl: 'https://api.example.com' }}>
+  {children}
+</GustoProvider>
+```
+
+#### Breadcrumbs replaced with Progress Bar
+
+The Breadcrumbs navigation component has been replaced with a Progress Bar for improved user experience. Update any code that relies on the Breadcrumbs component to use the new Progress Bar instead.
```
serikjensen, line 40: Should we remove this since breadcrumb isn't technically a breaking change? I guess it's breaking if you are depending on that in flow, but the breadcrumb was also not working before and i know bqe was just using block components

serikjensen (approved):
LGTM!

CHANGELOG.md
jeffredodd: Good call out, fixed

PR #314, opened 2025-05-21
[WIP] Autogenerate UI adapter props with typedoc (no markdown plugin)

serikjensen:
[WIP]

Autogenerate docs for adapters using [typedoc](https://typedoc.org/) without markdown plugin. Markdown plugin is really opinionated about the output, this leverages typedoc for the doc parsing functionality but then we take care of the linking completely.

See generated example docs live on the branch here:
https://github.com/Gusto/embedded-react-sdk/blob/chore/GWS-5000-custom-markdown/docs/06/01/component-inventory.md

Pros
* Generates a single file and we are able to configure the linking how we want
* We have control over the styling and formatting of the markdown

Cons
* Moderate level of configuration
* Does not document picked props in the props tables, you see this instead

Screenshot 2025-05-21 at 5 46 21 PM


serikjensen:
Opted to go with this direction instead https://github.com/Gusto/embedded-react-sdk/pull/315

PR #320, opened 2025-05-27
chore: expose all types from project root

jeffredodd:
I'm not 💯 sure this works, but in my investigation of types and types issues this seems like it might not be a bad thing to do. I know we want to avoid `export *` but maybe with types its ok?


serikjensen (approved):
Looks good! looks like the PR description could be updated

PR #327, opened 2025-05-28
[GWS-4966] Setting custom headers doesn't work

scally:
fixes https://gustohq.atlassian.net/browse/GWS-4966


dmortal (approved):
Thank you!

PR #328, opened 2025-05-30
feat: adding component level dictionary override

dmortal:
Enabling override of component dictionary on component level
Fixing types for dictionaries
GustoProvider level:
image
component level:
image


vite.config.ts
dmortal, line 1: This is important change

vite.config.ts
```diff
@@ -35,6 +35,13 @@ export default defineConfig(({ mode }) => {
             '**/test/**',
           ],
           strictOutput: true,
+          beforeWriteFile: (filePath, content) => {
+            // This is a hack to ensure that the i18next.d.ts file is imported in the index.d.ts file, otherwise i18next type augmentation is not available
+            if (filePath.includes('index.d.ts')) {
+              return { content: 'import "./types/i18next.d.ts"\n' + content, filePath }
+            }
```
serikjensen, line 42: is index.d.ts generated automatically? is that why we need to append to it here vs. editing it directly?

src/components/Company/DocumentSignerFlow/DocumentSignerFlow.tsx
```diff
@@ -16,7 +17,9 @@ export const DocumentSignerFlow = ({
   companyId,
   signatoryId,
   onEvent,
+  dictionary,
 }: DocumentSignerFlowProps) => {
+  useComponentDictionary('Company.DocumentList', dictionary)
```
serikjensen, line 22: On this one, shouldn't this be moved to the company DocumentSigner component itself? We're also missing this override for the company SignatureForm component with `Company.SignatureForm`

Thinking we probably need to figure out a way to set these at the flow level 🤔 for the purposes of this PR wondering if it's sufficient to just set them on the block level components and we can revisit setting them on flows separately?

Alternatively we may need to refactor to consolidate translations into a single file if that's the direction we want to go

src/components/Company/BankAccount/BankAccountFlow.tsx
```diff
@@ -8,12 +8,14 @@ import { bankAccountStateMachine } from './stateMachine'
 import { BankAccountListContextual } from './BankAccountComponents'
 import { Flow } from '@/components/Flow/Flow'
 import type { BaseComponentInterface } from '@/components/Base'
+import { useComponentDictionary } from '@/i18n/I18n'
 
-export interface LocationsProps extends BaseComponentInterface {
+export interface LocationsProps extends BaseComponentInterface<'Company.BankAccount'> {
```
serikjensen, line 13: Same question/observation here. In this case we have a single translations file for all BankAccountFlow components. Is that the preferable way moving forward? For collectives use case we'd still need to be able to set it on the block component. Maybe we can discuss further in parking lot

dmortal: actually, bankAccount has single dictionary

PR #331, opened 2025-06-03
fix: update comission zod schema

jeffredodd:
This PR is responsible for updating the Zod schema of Compensation to support the variability of rate.

### DIFF
--- a/src/components/Employee/Compensation/useCompensation.ts
+++ b/src/components/Employee/Compensation/useCompensation.ts
@@ -8,78 +8,104 @@ import { yearlyRate } from '@/helpers/payRateCalculator'
 export const CompensationSchema = z
   .object({
     jobTitle: z.string().min(1),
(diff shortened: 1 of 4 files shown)


src/components/Employee/Compensation/useCompensation.ts
```diff
@@ -8,78 +8,104 @@ import { yearlyRate } from '@/helpers/payRateCalculator'
 export const CompensationSchema = z
   .object({
     jobTitle: z.string().min(1),
-    // adjustForMinimumWage can be true or false
     adjustForMinimumWage: z.boolean(),
-    // If adjustForMinimumWage is true, minimumWageId is required
```
scally, line 13: thank you for removing this, the comment doesn't match the schema now or else this would have to be a variant

scally (approved):

PR #343, opened 2025-06-11
fix: update rate to not be labeled optional when it is required

serikjensen:
This PR updates compensation amount (rate) to no longer be labeled as optional since it is required by most employee types. It also updates the validation messages to correctly reflect the underlying validation logic in the schema (ex. provides an explanation if it is required, if the rate amount does not satisfy the min threshold etc)

### Before

Compensation amount is labeled as "optional" when it is actually required. Error message also simply says that the amount is required, it does not provide helpful context. Ex. in the video you can observe that the entered amount does not meet the min salary threshold. The validation message is just around the required field which is frustrating UX.

https://github.com/user-attachments/assets/eb02a468-c8cd-4618-8352-c8fdce03d958

### After

Observe that
* Compensation amount is no longer labeled as optional
* Validation messages are descriptive about the issue the field is actually experiencing

https://github.com/user-attachments/assets/7b7e0fc3-c17f-4be5-ada6-a5df086433a5

### DIFF
--- a/src/components/Employee/Compensation/Edit.tsx
+++ b/src/components/Employee/Compensation/Edit.tsx
@@ -1,7 +1,12 @@
 import { useEffect, useMemo } from 'react'
 import { useFormContext, useWatch } from 'react-hook-form'
 import { Trans, useTranslation } from 'react-i18next'
-import { type CompensationInputs, useCompensation } from './useCompensation'
+import {
+  type CompensationInputs,
+  useCompensation,
+  rateMinimumError,
+  rateExemptThresholdError,
+} from './useCompensation'
 import { FLSA_OVERTIME_SALARY_LIMIT, FlsaStatus, PAY_PERIODS } from '@/shared/constants'
 import useNumberFormatter from '@/components/Common/hooks/useNumberFormatter'
 import {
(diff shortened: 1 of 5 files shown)


src/components/Employee/Compensation/Edit.tsx
```diff
@@ -122,8 +137,8 @@ export const Edit = () => {
         label={t('amount')}
         format="currency"
         min={0}
-        errorMessage={t('validations.rate')}
-        isRequired={false}
+        errorMessage={rateErrorMessage}
+        isRequired
```
jeffredodd, line 141: With this enabled, can the user still submit rates being optional without the browsers default validation throwing an error here since a required field is empty?

jeffredodd (approved):
LGTM, just a simple comment but aside from that 🚢

src/components/Employee/Compensation/Edit.tsx
serikjensen: Yes! since we don't supply the required attribute natively. 

https://github.com/user-attachments/assets/55296dde-4c03-479c-9b53-3d7dc2eac43a

PR #351, opened 2025-06-17
fix: fixing split validation

dmortal:
https://gustohq.atlassian.net/browse/GWS-5140
Fixing paymentMethod split validatioln


serikjensen (approved):
👏

PR #353, opened 2025-06-17
fix: skipping state taxes for state that has only questions for admins

dmortal:
Skipping states in state taxes where user is not admin and state only contains admin specific questions


serikjensen (approved):
Verified locally it's hiding the state taxes heading per the acceptance criteria 👍 however, we may want to investigate this further prior to merge to ensure it's not related https://gustohq.slack.com/archives/C071WFJUNF5/p1750192002143109

PR #369, opened 2025-06-30
refactor: extract `useInvalidateOnSuccess` pattern

scally:
Trying to help prevent a common mistake, which is mutating without invalidating the correct API responses from the cache.

This approach lifts up mutation another level into a hook that uses `useInvalidateOnSuccess`. `useInvalidateOnSuccess` accepts a list of invalidation functions to call once your mutation succeeds and then gives access to the mutator function and pending state to be used as usual in our forms.

### DIFF
--- a/src/components/Employee/Profile/hooks.ts
+++ b/src/components/Employee/Profile/hooks.ts
@@ -1,0 +1,18 @@
+import { useEmployeeAddressesCreateWorkAddressMutation } from '@gusto/embedded-api/react-query/employeeAddressesCreateWorkAddress'
+import { invalidateAllEmployeeAddressesGetWorkAddresses } from '@gusto/embedded-api/react-query/employeeAddressesGetWorkAddresses'
+import { useEmployeeAddressesUpdateWorkAddressMutation } from '@gusto/embedded-api/react-query/employeeAddressesUpdateWorkAddress'
+import { useInvalidateOnSuccess } from '@/hooks/useInvalidateOnSuccess'
+
+export const useCreateEmployeeWorkAddress = () => {
+  return useInvalidateOnSuccess({
+    invalidators: [invalidateAllEmployeeAddressesGetWorkAddresses],
+    mutator: useEmployeeAddressesCreateWorkAddressMutation,
+  })
+}
+
+export const useUpdateEmployeeWorkAddress = () => {
+  return useInvalidateOnSuccess({
+    invalidators: [invalidateAllEmployeeAddressesGetWorkAddresses],
+    mutator: useEmployeeAddressesUpdateWorkAddressMutation,
+  })
+}
--- a/src/hooks/useInvalidateOnSuccess/index.ts
+++ b/src/hooks/useInvalidateOnSuccess/index.ts
@@ -1,0 +1,26 @@
+import type { MutationHookOptions } from '@gusto/embedded-api/react-query/_types'
+import { useQueryClient, type QueryClient, type UseMutationResult } from '@tanstack/react-query'
+
+interface UseInvalidateOnSuccessArgs<TData, TVariables> {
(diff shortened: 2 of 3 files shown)


src/hooks/useInvalidateOnSuccess/index.ts
```diff
@@ -0,0 +1,26 @@
+import type { MutationHookOptions } from '@gusto/embedded-api/react-query/_types'
+import { useQueryClient, type QueryClient, type UseMutationResult } from '@tanstack/react-query'
+
+interface UseInvalidateOnSuccessArgs<TData, TVariables> {
+  invalidators: ((queryClient: QueryClient) => Promise<unknown>)[]
+  mutator: ({
+    onSuccess,
+  }: Pick<MutationHookOptions, 'onSuccess'>) => UseMutationResult<TData, Error, TVariables>
```
scally, line 8: Getting this signature right was about 50% of this PR :D

src/components/Employee/Profile/hooks.ts
```diff
@@ -0,0 +1,18 @@
+import { useEmployeeAddressesCreateWorkAddressMutation } from '@gusto/embedded-api/react-query/employeeAddressesCreateWorkAddress'
+import { invalidateAllEmployeeAddressesGetWorkAddresses } from '@gusto/embedded-api/react-query/employeeAddressesGetWorkAddresses'
+import { useEmployeeAddressesUpdateWorkAddressMutation } from '@gusto/embedded-api/react-query/employeeAddressesUpdateWorkAddress'
+import { useInvalidateOnSuccess } from '@/hooks/useInvalidateOnSuccess'
+
+export const useCreateEmployeeWorkAddress = () => {
+  return useInvalidateOnSuccess({
+    invalidators: [invalidateAllEmployeeAddressesGetWorkAddresses],
```
scally, line 8: In these cases we only invalidate one thing per mutation, but I thought we might need to eventually support multiple invalidations for some objects, so making this take an array is a safeguard around not needing to change call sites later

scally:
references [GWS-5171]

serikjensen (commented):
Looks good!

* Left a nit/question around `mutator` naming, wondering if react has any optimizations around having `use` for hooks?
* Question about best practice moving forward: would the rec be that we move all query hooks with a mutation into the `hooks` directory? or only some of them? just worried it might be confusing to enforce as a recommendation. Ideally we could place all custom hooks that need invalidators into the hooks directory with this pattern and then have some sort of automated linting in place to help guide people to use the right hook

scally:
@dmortal 

> this might create some confusion around using speakeasy mutators directly vs this abstraction approach
To clarify, I would be proposing we don't use the mutators directly and instead use our own custom hooks in all cases. We could use something like our existing linting rules to enforce.

> it also potentially shifts the problem of knowing what needs to be invalidated to another place
This is a typical tradeoff of creating an abstraction -- moving the "knowledge" of how an operation works elsewhere. I'd argue that this knowledge shouldn't live in a render function in a component, but should be elevated higher up, and that it shouldn't be "automagical" since it is a function you have to call and not automatic behavior.

src/hooks/useInvalidateOnSuccess/index.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const useInvalidateOnSuccess = <TData, TVariables>({
+  invalidators,
+  mutator,
+}: UseInvalidateOnSuccessArgs<TData, TVariables>) => {
+  const queryClient = useQueryClient()
+  const { mutateAsync, isPending } = mutator({
```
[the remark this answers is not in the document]
scally: It does seem to work fine in manual testing... perhaps it's not the violation I thought it was.

PR #373, opened 2025-07-02
chore: Introduce frontmatter generator to sync docs data for Github Action Uplaod

jeffredodd:
This PR is responsible for taking our `docs-lock.yml` file and syncing Front matter to it in order to prepare for our upload through the Github Action

This change includes
- Bug fix where new files had an order of 999
- Creation of front matter for existing docs
- Creation of front matter syncing script
- Tests to validate front matter works as expected


docs/workflows-overview/contractor-onboarding.md
```diff
@@ -1,3 +1,10 @@
+---
+title: Contractor Profile Component
+category: 6849ddd92905ee0053320687
+slug: contractor-profile-component
+hidden: false
+parentDoc: 6849ddd92905ee005332084d
```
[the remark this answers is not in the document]
jeffredodd: This is expected and by design per readme docs.

docs/integration-guide/event-types.md
```diff
@@ -1,3 +1,10 @@
+---
+title: 'title: Event Types'
```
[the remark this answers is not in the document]
jeffredodd: Good catch, fixed in script

jeffredodd:
> I'm a little confused about why adding doc generation adds 635 lines of new test code to the system @scally 

Fair point, mostly have created tests here to make sure we don't run up against any edge cases as a defense. The Readme action/api is incredibly brittle (So much so they are abandoning it and we will be migrated in the next year or so), but because of this brittleness i've been leaning on many test cases to make sure we're getting the expected output and to also alert on any changes we might accidentally introduce. 

TLDR; defensive coding

.docs/docs-lock.yml
```diff
@@ -154,30 +178,18 @@ categories:
...
             slug: null
-            order: 999
+            order: null
             hidden: false
             localPath: docs/integration-guide/error-handling.md
             isNew: true
             children: []
           - id: null
-            title: Event Types
+            title: 'title: Event Types'
```
serikjensen, line 190: Whoops! is this the result of a bug?

docs/component-adapter/component-adapter-faq.md
```diff
@@ -1,3 +1,12 @@
+---
+title: Component Adapter FAQ
+category: 6849ddd92905ee0053320687
+slug: component-adapter-faq
+hidden: false
+parentDoc: 6852ed81d85ebd00247da5a0
```
serikjensen, line 6: Is there a way we can do this without storing the category and parent doc in the frontmatter? Worried this is going to be hard to add new documents. Maybe we could chat more about this tomorrow

the hope would be that we could use a unique identifier (file path + filename) and then store identifiers like this in the lockfile instead. I may be missing something tho

jeffredodd: Same concern. FWIW see docs https://docs.readme.com/main/docs/rdme (Link looks generic is not, its for the CLI)

jeffredodd: Just pushed a change that creates a temporary folder to mirror our docs. We add the parentDoc ID and category ID there, and that folder becomes the source for our GitHub Action to publish from.

It makes me wonder if we should have front matter at all. However we have options for `order` `description` `hidden` which seem like a smart place for them.

serikjensen (approved):
Thanks for updating!

PR #391, opened 2025-07-14
[GWS-5167] Add contractor submit block

scally:
Screenshot 2025-07-16 at 5 18 50 PM

Screenshot 2025-07-16 at 5 18 43 PM


src/components/Contractor/Submit/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { Flex } from '@/components/Common'
+import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentContext'
+import WarningIcon from '@/assets/icons/warning.svg?react'
+import { useI18n } from '@/i18n'
+
+interface ContractorSubmitProps {
+  onSubmit: () => void
+}
+
+export function ContractorSubmit({ onSubmit }: ContractorSubmitProps) {
```
dmortal, line 11: Isn't there an API submit action this component must perform?

src/components/Contractor/Submit/index.tsx
```diff
@@ -0,0 +1,33 @@
...
+  useI18n('Contractor.ContractorSubmit')
+  const { Button, Heading, UnorderedList } = useComponentContext()
+  const { t } = useTranslation('Contractor.ContractorSubmit')
+  const items = Object.values(t('warningItems', { returnObjects: true }))
+
+  return (
+    <Flex flexDirection="column">
```
dmortal, line 18: based on design, this looks like an `Alert` component that we have - that component also has correct aria roles on it

src/components/Contractor/Submit/index.tsx
scally: @dmortal did we write down anywhere which API this is meant to call? Otherwise I can go take a look at API specs and take a guess. At the moment it's not super clear where this goes or what data it's supposed to have to hand ready to submit

scally: Talked elsewhere and it looks like this is the API intended to be called https://docs.gusto.com/embedded-payroll/docs/onboard-a-contractor#3-mark-contractor-as-onboarded

src/components/Contractor/Submit/index.tsx
scally: Changed to use Alert and some styles

serikjensen (approved):
Looks good!

small nit, typically i've seen us do filenames that match the name of the React component inside. Ex. instead of `submitDone.tsx` doing `SubmitDone.tsx`. Or instead of `withApi.tsx` doing `ContractorSubmitWithApi.tsx`. Def not a showstopper!

PR #392, opened 2025-07-15
fix: document signer machine recreated on signatory change

jeffredodd:
### Problem
When users went through company onboarding and asserted themselves as the signatory, they couldn't sign documents on the next step. The system showed "Only the signatory can sign documents" even though they had just identified themselves as the signatory.

### Root Cause
The DocumentSigner state machine was being recreated multiple times during signatory creation due to a race condition:

1. **Delete existing signatory** → `doesSignatoryExist` changes from `true` to `false`
2. **API refetch triggers** due to query invalidation 
3. **State machine recreates** due to dependency change → loses `signatoryId` context
4. **Create new signatory** → `doesSignatoryExist` changes from `false` to `true` 
5. **State machine recreates again** → context lost again

The `useMemo` dependency array `[companyId, doesSignatoryExist, signatoryId]` was causing the machine to recreate every time these values changed, wiping out the signatory context that had been set by the state transitions.

### Solution
Changed the `useMemo` dependency array to only include `[companyId]`, preventing unnecessary recreations when signatory data changes. This preserves the `signatoryId` context throughout the signatory creation flow.


serikjensen (approved):
Looks good!

PR #393, opened 2025-07-15
fix: upgrade embedded api and fix state tax input issues

serikjensen:
This PR
* Updates the embedded api to the latest version to get a schema fix for state taxes
* Updates state tax inputs with fixes around `tax_rate` inputs which require a select enum

### DIFF
--- a/src/components/Common/TaxInputs/TaxInputs.tsx
+++ b/src/components/Common/TaxInputs/TaxInputs.tsx
@@ -49,7 +49,7 @@ export function QuestionInput({
       )
     case 'percent':
     case 'tax_rate':
-      return <NumberInput {...props} isPercent />
+      return <TaxRateInput {...props} />
     case 'currency':
       return <NumberInput {...props} isCurrency />
     default:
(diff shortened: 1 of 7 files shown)


src/components/Common/TaxInputs/TaxInputs.tsx
```diff
@@ -203,3 +204,30 @@ export function DateField({
     />
   )
 }
+
+export function TaxRateInput({ requirement, question, ...props }: EmpQ | CompR) {
+  if (requirement) {
+    // Covers case for tax rate where the rate is a one_of option and must be submitted as a string
+    // of enumerated rate values provided
+    const { key, metadata, label, description } = requirement
```
scally, line 212: Could this be destructured in the function declaration?

src/components/Common/TaxInputs/TaxInputs.tsx
```diff
@@ -203,3 +204,30 @@ export function DateField({
...
+    const { validation } = metadata || {}
+    return validation?.type === 'one_of' ? (
+      <SelectField
+        isRequired
+        name={key || ''}
+        label={label as string}
```
scally, line 218: Can we coerce to string instead of `as`'ing it? `String(label)` ?

PR #397, opened 2025-07-16
fix: state tax fixes for tax_rate fields not submitting

serikjensen:
This PR makes the following updates for company state taxes:
* Updates tax_rate inputs to be of type TextInput instead of NumberInput. This is because the api can accept strings universally for these numeric values for tax rates but it can only accept numbers occasionally. There was no way to make the distinction with the api output so using strings was safer. We add `type="number"` to the text input component which prevents non numeric characters. Verified that the validation was working for a sampling of states and properly outputting an error message
* Updates validation messages to consistently output a more human readable required message (before they were saying things like "expected null but got boolean" etc
* Updates default values to correctly not set as strings for radios

## Proof of functionality

https://github.com/user-attachments/assets/01c6b222-c974-4ce8-a4f4-a60e57a03182

### DIFF
--- a/src/components/Common/TaxInputs/TaxInputs.tsx
+++ b/src/components/Common/TaxInputs/TaxInputs.tsx
@@ -25,6 +25,8 @@ interface CompR {
 
 type NumberFieldProps = { isCurrency?: boolean; isPercent?: boolean }
 
+type TextInputProps = { type?: string; isPercent?: boolean }
+
 export function QuestionInput({
   questionType,
   ...props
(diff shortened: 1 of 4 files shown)


src/components/Common/TaxInputs/TaxInputs.tsx
```diff
@@ -85,7 +87,13 @@ export function SelectInput({ question, requirement, isDisabled = false }: EmpQ
   )
 }
 
-export function TextInput({ question, requirement, isDisabled = false }: EmpQ | CompR) {
+export function TextInput({
```
jeffredodd, line 90: Doesen't matter, but if you have a text input, and it allows you to set it types to text.... is it then just an input? 🤔

jeffredodd (approved):

src/components/Common/TaxInputs/TaxInputs.tsx
serikjensen: Good question! I would've said no before this PR but now i need that functionality 😂 

we may have to rethink it if it throws a curveball for adapter folks. but for our purposes here it works well to restrict the input to numeric characters only while providing string output

if we do have to get away from this we'll need to find a way to use the number input and then convert the value to a string for hook form

jeffredodd: Yeah its fine for now especially since these are Tax Inputs. Just had the thought that we were pulling an `ATM Machine` faux pas

PR #401, opened 2025-07-17
chore: loosen peer deps for react to include 18 and 19

serikjensen:
This updates to allow for React 18 or 19 as a peer dependency.

In order to make the SDK React 19 compatible, we had to move away from passing `ref` as a function argument. We opted to avoid a replacement with forwardRef for this case because we wanted it to be more straightforward for consumers to wire up adapters. The only instance where we passed ref like this was Button which we updated to have a `buttonRef` prop which can be wired up.

## Proof of functionality

I updated the SDK workshop app to point to the RC version of the react sdk with the adjusted peer deps and tested it both with React 18 and 19 to verify no regressions.

### App working with React 18

https://github.com/user-attachments/assets/f7cf0cfd-b5f2-4c68-b0b5-3765049c80e0

### App working with React 19

https://github.com/user-attachments/assets/115b6ed4-687f-4a58-bf9c-304e201b366e

### DIFF
--- a/.ladle/adapters/PlainComponentAdapter.tsx
+++ b/.ladle/adapters/PlainComponentAdapter.tsx
@@ -63,7 +63,7 @@ export const PlainComponentAdapter: ComponentsContextType = {
     isError = false,
     isLoading = false,
     isDisabled = false,
-    ref,
+    buttonRef,
     onClick,
     children,
     ...props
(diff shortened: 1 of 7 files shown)


jeffredodd (approved):
Looks good, good catch with the button ref stuff

PR #406, opened 2025-07-21
chore: add new theming infrastructure

serikjensen:
This updates to add infrastructure for new theme variables.

The theme variables represent a more simplified partner theme that carries over the core entries of the previous theme but in a way that will be simpler for partners to interface with.

We currently support both the legacy theme (previous approach) and the updated approach. This will allow us to incrementally update components to use the new theming infrastructure while we make the migration. It will also allow us to make changes to the new theme as we implement it with more of our components.

This PR also updates the button to use the new theme variables as a first component to get the migration.

## Button using new theme variables

Screenshot 2025-07-21 at 4 21 42 PM

## Both old and new variables getting injected into the document head

Screenshot 2025-07-21 at 4 23 38 PM

### DIFF
--- a/src/contexts/ThemeProvider/theme.ts
+++ b/src/contexts/ThemeProvider/theme.ts
@@ -1,0 +1,93 @@
+import '@/styles/sdk.scss'
+import { getRootFontSize, toRem } from '@/helpers/rem'
+
+// Colors are for internal use in our theme currently
+// We don't export them for partner use or overrides
+const colors = {
+  neutral: {
+    100: '#FFFFFF',
+    200: '#FBFAFA',
+    300: '#F4F4F3',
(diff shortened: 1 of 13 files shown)


src/contexts/ThemeProvider/theme.ts
```diff
@@ -0,0 +1,93 @@
...
+    400: '#2BABAD',
+    500: '#0A8080',
+    800: '#005961',
+  },
+  orange: {
+    800: '#CA464A',
+  },
+}
+
+export const gustoSDKTheme = {
```
serikjensen, line 47: This would be the new theme object

jeffredodd (approved):
LGTM

PR #412, opened 2025-07-22
chore: update Badge to use new theme variables

serikjensen:
This updates the Badge component to use the new theme variables
* Removes the old _Badge.scss file because it is no longer used since Badge was updated to use modules
* Updates to have info colors which weren't set before

## Before

https://github.com/user-attachments/assets/0306b1f9-d10b-4e44-bd22-4fb5ed6fd8ee

## After

https://github.com/user-attachments/assets/fb2dfec6-0a21-4922-8c5e-0dfa3826e04a


src/components/Common/UI/Badge/Badge.module.scss
```diff
@@ -1,41 +1,41 @@
 .badge {
   display: inline-block;
-  padding: var(--g-badge-paddingY) var(--g-badge-paddingX);
-  font-size: var(--g-badge-fontSize);
-  font-weight: var(--g-badge-fontWeight);
+  padding: toRem(2) toRem(10);
```
dmortal, line 3: question: are we sure that we don't want to maintain global spacing map, if only for internal use only, for consistency and to make future refactors easier?

src/components/Common/UI/Badge/Badge.module.scss
```diff
@@ -1,41 +1,41 @@
...
+  font-size: var(--g-fontSizeSmall);
+  font-weight: var(--g-fontWeightMedium);
   text-align: center;
   white-space: nowrap;
   vertical-align: baseline;
   border-style: solid;
-  border-width: var(--g-badge-borderWidth);
-  border-radius: var(--g-badge-borderRadius);
+  border-width: 1px;
```
dmortal, line 10: similar question, should we have a global variable for border widths that are reused, even if not exposed to theme engine?

serikjensen: yeah it's a matter of thinking about if things should change together. 🤔 i think if we do it we should probably expose it similar to the radius ones where we would do like a `badgeBorderWidth`, `buttonBorderWidth` etc. I don't love adding the component level variables, but they could make sense for some minimal amount of properties. I basically did it for radius b/c i know border radius is a really common differentiator between buttons, badges, and inputs and i didn't think there was a way around it

dmortal (approved):

PR #416, opened 2025-07-23
feat: update Button styling and variants

serikjensen:
This PR reorganizes Button per design feedback as we were updating to use the new tokens

* Adds a tertiary button
* "error" is now one of the button variants along with primary, secondary, tertiary
* ButtonIcon now supports all button variants above (including tertiary which is the default) which allows icon button to take on the other button states. This one is also from design request

## All button variants

Screenshot 2025-07-23 at 2 00 23 PM

## All icon button variants

Screenshot 2025-07-23 at 2 00 54 PM

## Hover/active states

https://github.com/user-attachments/assets/b9f00236-c581-4d02-80c9-98c78ebc5508

### DIFF
--- a/src/components/Common/UI/Button/Button.stories.tsx
+++ b/src/components/Common/UI/Button/Button.stories.tsx
@@ -16,11 +16,6 @@ export const Loading = () => {
   return <Components.Button isLoading>Loading Button</Components.Button>
 }
 
-export const Error = () => {
-  const Components = useComponentContext()
-  return <Components.Button isError>Error Button</Components.Button>
-}
-
 export const Disabled = () => {
   const Components = useComponentContext()
   return <Components.Button isDisabled>Disabled Button</Components.Button>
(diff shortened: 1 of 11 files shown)


src/components/Common/UI/Button/Button.stories.tsx
```diff
@@ -31,13 +26,8 @@ export const ButtonGrid = () => {
   const Components = useComponentContext()
   const states = [
     { label: 'Default', props: {} },
-    { label: 'Hover', props: { 'data-hovered': true } },
-    { label: 'Pressed', props: { 'data-pressed': true } },
-    { label: 'Focus', props: { 'data-focus-visible': true } },
```
serikjensen, line 36: These css states weren't being triggered anyway in the preview so removed them

scally (approved):
Nothing to add here, make it happen cap'n

PR #422, opened 2025-07-25
feat: introduce Run Payroll state machines

scally:


src/components/Payroll/RunPayrollFlow/RunPayrollFlow.stories.tsx
```diff
@@ -0,0 +1,352 @@
...
+    data: [{ payrollId: 'abcd' }],
+  }
+}
+export const DemoRunPayrollFlow = ({ companyId, onEvent }) => {
+  const [currentPayrollId, setCurrentPayrollId] = useState(undefined)
+  const [isCalculated, setIsCalculated] = useState(false)
+  const { data: payrolls } = useListCompanyPayrollsApi({ companyId })
+  const { mutate } = useSubmitPayrollApi({ payrollId: currentPayrollId })
+
+  const onEdit = () => {
```
scally, line 318: We can convert this to `useReducer` here instead of `useState`, I wasn't convinced initially that this was too much state wrangling but upon further reflection it might make comparison easier. Thoughts?

src/components/Payroll/RunPayrollFlow/RunPayrollFlow.stories.tsx
```diff
@@ -0,0 +1,406 @@
...
+
+// **** PRESENTATIONAL ****
+// **** PRESENTATIONAL ****
+// **** PRESENTATIONAL ****
+
+interface PayrollListProps {
+  onRunPayroll: ({ payrollId }: { payrollId: string }) => void
+  payrolls: any[]
+}
+const PayrollList = ({ onRunPayroll, payrolls }: PayrollListProps) => {
```
serikjensen, line 31: I'm assuming this one would have a parent PayrollListBlock as well that is handling onRunPayroll and fetching the payrolls?

src/components/Payroll/RunPayrollFlow/RunPayrollFlow.stories.tsx
```diff
@@ -0,0 +1,406 @@
...
+// **** BLOCKS ****
+// **** BLOCKS ****
+// **** BLOCKS ****
+
+// TODO: Replace this hook with call to Speakeasy instead
+const useEditEmployeeApi = ({ employeeId }) => {
+  const mutate = async () => {}
+  return { mutate }
+}
+const PayrollEditEmployeeBlock = ({ employeeId, onEvent, onSaved }) => {
```
serikjensen, line 245: Do we actually need `onSaved` exposed here? or could we just use the `onEvent` prop?

serikjensen (commented):
This is looking great excited to see it coming together!

Left a couple of questions around block components. It might be helpful to see the corresponding PayrollListBlock, PayrollOverviewBlock similar to what you have for PayrollEditEmployeeBlock and PayrollConfigurationBlock. I'm assuming you would want a similar composition for each? And I'm assuming each one would have a dedicated hook that handles all the speakeasy calls/logic/etc?

Left additional comments around the component APIs for the block components. Typically it seems we've just kept those simple and gotten navigation from `onEvent` Wondering if that's something we should continue?

src/components/Payroll/RunPayrollFlow/RunPayrollFlow.stories.tsx
scally: @serikjensen  This is part of the question I was asking re: are the blocks in the right place. Assume for now that the things named Block are the blocks we are shipping and everything else is only reachable via the Flow. What do you think about these choices? Do we have sufficient evidence that everything should be a block even if their available actions don't make as much sense without the rest of the blocks?

src/components/Payroll/RunPayrollFlow/RunPayrollFlow.stories.tsx
```diff
@@ -0,0 +1,406 @@
...
+          <NumberInputField defaultValue={40} isRequired label="Hours" name="hours" />
+        </Form>
+      </FormProvider>
+
+      <Button onClick={onDone} title="Done">
+        Done
+      </Button>
+    </Flex>
+  )
+}
```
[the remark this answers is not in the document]
scally: Yes, I do believe we can separate out form submission from presentation concerns. I think that can be an orthogonal decision from what is shipped as a block or not, though.

src/components/Payroll/RunPayrollFlow/RunPayrollFlow.stories.tsx
serikjensen: yeah my thought in still keeping them as blocks would be
* Consistency with the development approach
* Ability to integrate with an application router
* Flexibility for the partner (not a super strong argument in this case, unlikely they would be moving things around)

I'd say i care about things in that order. Keeping our development approach consistent across the domains for ease of understanding and knowing where things are as well as unlocking the ability to manage the transitions with a router instead of a state machine as outlined here https://docs.gusto.com/embedded-payroll/docs/routing

src/components/Payroll/RunPayrollFlow/RunPayrollFlow.stories.tsx
serikjensen: yeah i think it's an interesting discussion. I'd maybe favor just uniformly using onEvent, but i'd be open to arguments to have better named callbacks. In the case of callbacks that also would provide better type safety vs. onEvent which just kinda gets any data passed along

scally: I'm OK with using onEvent as well here, but suspect it would lead to our next discussion being "why do we use onEvent everywhere instead of discrete event handlers" and I'm maybe jumping ahead a step here

PR #425, opened 2025-07-30
chore: introduce cursor rule files to help those who use its ai

jeffredodd:
Hoping to kick off some cursor rules for us...

I had cursor help write these so lets hope it knows best practices for its own `.mdc` specs

Rules
- Explain to cursor that `npm run test` will put it into watch mode and that it should use `-- --run` so it exits properly
- Explain to cursor that as a team we value readable code thats clear and concise over comment spam. Comments indicate that code could be better formatted or reorganized

### DIFF
--- a/.cursor/rules/code-style.mdc
+++ b/.cursor/rules/code-style.mdc
@@ -1,0 +1,117 @@
+# Code Style: Prefer Readable Code Over Comments
+
+## Principle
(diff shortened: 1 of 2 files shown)


serikjensen (approved):
Looks like a great start! thanks for spinning this up!

PR #442, opened 2025-08-07
draft: test improvements

jeffredodd:
Creating a PR for us to discuss test improvements that might help our e2e test\s run better or just speed up CI in general.

### DIFF
--- a/src/test/mocks/setupMswForTest.ts
+++ b/src/test/mocks/setupMswForTest.ts
@@ -1,0 +1,36 @@
+import { beforeAll, afterEach, afterAll } from 'vitest'
+import { server } from './server'
+
+let serverStarted = false
+let testsUsingMsw = 0
+
+/**
+ * Sets up MSW server for individual test files that need API mocking.
+ * Call this function at the top level of test files that use setupApiTestMocks() or server.use().
+ *
+ * This replaces the global MSW setup for better performance - only tests that actually
+ * need API mocking will pay the MSW startup cost.
+ */
+export const setupMswForTest = () => {
+  beforeAll(() => {
+    if (!serverStarted) {
+      server.listen()
+      serverStarted = true
+    }
+    testsUsingMsw++
+  })
+
+  afterEach(() => {
+    // Remove any handlers added in individual tests (runtime handlers)
+    server.resetHandlers()
+  })
+
+  afterAll(() => {
+    testsUsingMsw--
+    // Only close server when no more tests are using it
+    if (testsUsingMsw === 0 && serverStarted) {
+      server.close()
+      serverStarted = false
+    }
+  })
+}
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -79,7 +79,7 @@ export default defineConfig(({ mode }) => {
         cssFileName: 'style',
       },
       minify: !isDev,
-      sourcemap: !isDev,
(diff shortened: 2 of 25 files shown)


vite.config.ts
```diff
@@ -103,6 +103,9 @@ export default defineConfig(({ mode }) => {
       environment: 'jsdom',
       globals: true,
       setupFiles: ['./src/test/setup.ts'],
+      // Performance optimizations
+      css: false, // Skip CSS processing in tests
```
scally, line 107: What specific speedup do we get from these particular lines?

jeffredodd: We don't really need CSS in JSDOM because its not gonna render so was just a speed boost

The other is source maps, we're generating source maps for tests. 

Saved about 3seconds locally

src/test/mocks/setupMswForTest.ts
```diff
@@ -0,0 +1,36 @@
...
+  })
+
+  afterEach(() => {
+    // Remove any handlers added in individual tests (runtime handlers)
+    server.resetHandlers()
+  })
+
+  afterAll(() => {
+    testsUsingMsw--
+    // Only close server when no more tests are using it
```
scally, line 30: This is new, do we have evidence that supports that we need it?

src/test-utils/loadingHelpers.ts
```diff
@@ -0,0 +1,66 @@
+import { screen, waitFor } from '@testing-library/react'
+
+export const waitForPageLoad = async (
+  options: {
+    timeout?: number
+    waitForElement?: () => Promise<HTMLElement>
+    loadingIndicator?: string
+  } = {},
+) => {
+  const { timeout = 15000, waitForElement, loadingIndicator = 'Loading component...' } = options
```
scally, line 10: This seems to duplicate functionality that already exists in testing-library, as well as introducing a longer timeout that sets a higher global ceiling on test run times. What problem is this solving?

vite.config.ts
scally: What % speedup is 3sec for us? What run time did you see with/without?

jeffredodd:
> @jeffredodd one thing that might help here is that optimizations can be hard to reason about time-savings-wise in total like this, and shipping them one at a time with the savings information (old run time, new run time, %, etc) might help us better reason about whether or not the tradeoff of making devex worse in other ways is worth the speedup

100% agree, i'm going to try to break out what i've found here that works into exactly that. Do you see anything you'd like from this PR?

Basically what i've found is
- there really are not a lot of time savings, we're pretty effient
- the biggest savings were not setting up MSW for every test when an API isn't called
- I also sped up the e2e tests fast enough that i think I found their race condition, tldr skeleton was present and the test was failing because it was moving too fast. Things like altering typing speed of this test provided speed improvements
- everything else i think is probably not going to help us
- Moving vite to es2023 build actually sped up some things, minor but nice

What didn't work well
- loading only MSW fixtures you need, didn't gain us much speed
- batching events a user does was thought to speed us up but didn't show anything meaningful

What i think we'd like
- Maybe its me but i can't stand noisey test output, its a smell thing and I think we could do better here, so I might ship these fixes separately for our sanity

Any other thoughts @scally before I start to chunk this out? @serikjensen @dmortal anything you want to add?

PR #443, opened 2025-08-07
feat: update switch to use new theme vars

serikjensen:
This updates Switch to use the new theme variables

## Before

Screenshot 2025-08-07 at 3 43 28 PM

## After

Screenshot 2025-08-07 at 4 02 20 PM


src/components/Common/UI/Switch/Switch.module.scss
```diff
@@ -24,68 +20,65 @@
...
-      color: var(--g-typography-disabledTextColor);
       cursor: not-allowed;
 
       .indicator {
-        background: var(--g-colors-gray-100);
-
-        &::after {
-          background: var(--g-colors-gray-300);
-        }
+        opacity: 0.2;
```
jeffredodd, line 77: we talked about this, but are you doing the same opacity for all disabled things? If so maybe its a variable we should expose for now? That way a consumer can get their color contrast passing and use that as a lever to change. Then it won't matter what their primary color is too 😄

jeffredodd (approved):
LGTM - Just a minor nit / idea

PR #459, opened 2025-08-14
fix: update deductions copy and export components

serikjensen:
* Updates deductions copy around amount instead of leftover tax copy
* Removes unused deductions translations entry
* Exports deduction sub components

Screenshot 2025-08-14 at 4 04 45 PM


jeffredodd (approved):
🚢

PR #484, opened 2025-08-22
build: version bump to v0.12.1

scally:


serikjensen (commented):
Could you run npm install locally so it picks up the change in the lockfile as well?

PR #498, opened 2025-08-27
fix: resolve issue with business contractor new hire report

serikjensen:
Fixes some issues around business contractors and filing a new hire report. The type and self onboarding fields were not being set in the API call for updating the contractor. This ensures that we apply the contractor type and if we are self onboarding correctly along with the new hire report details.

This also updates the Submit component to properly set the invite contractor to self onboarding pending invite now that the above was resolved.

https://github.com/user-attachments/assets/04448c91-8635-4e8b-81e0-0a734d958266


src/components/Contractor/Submit/Submit.tsx
```diff
@@ -76,7 +76,7 @@ export const Root = ({ contractorId, selfOnboarding }: ContractorSubmitProps) =>
     return <SubmitDone onDone={handleSubmitDone} />
   }
   if (
-    onboardingStatus === ContractorOnboardingStatus.ADMIN_ONBOARDING_INCOMPLETE &&
+    onboardingStatus === ContractorOnboardingStatus.SELF_ONBOARDING_NOT_INVITED &&
```
shermango, line 79: for my understanding, why are we changing the enum check here? and is there ever a case where multiple enums would qualify as truthy for this check?

shermango (commented):
left a question, also should a unit test be included for asserting the request payload for when `selfOnboarding` is/is not provided?

src/components/Contractor/Submit/Submit.tsx
serikjensen: I think this was messed up previously due to the contractor update endpoint being called incorrectly in the file new hire report. This was a needed change after i corrected that

For the other statuses we wouldn't want to show the invite screen... unless they were changing the recipient email, but that's a case we would probably need some UX around anyway

This is probably still something that will need work on refining moving forward separately as contractor onboarding gets traction with partners, but this at least makes the happy path functional for now.

jeffredodd (approved):

PR #509, opened 2025-08-29
GWS-5415 Implement PayrollList connection to API

scally:


src/components/Payroll/PayrollList/PayrollListPresentation.tsx
```diff
@@ -1,46 +1,53 @@
...
+              </Text>
+              <Text>
+                {paySchedules.find(schedule => schedule.uuid === payPeriod?.payScheduleUuid)
+                  ?.name ||
+                  paySchedules.find(schedule => schedule.uuid === payPeriod?.payScheduleUuid)
+                    ?.customName}
+              </Text>
             </Flex>
           ),
           title: 'Pay period',
```
jeffredodd, line 34: i18n? Totally fine if we're waiting until later to do it all

serikjensen (approved):
Looks good! we should probs ticket i18n as jeff pointed out or take care of it here

src/components/Payroll/PayrollList/PayrollListPresentation.tsx
scally: Probably faster to do it now than to ticket it 🤷🏻

src/components/Payroll/PayrollList/PayrollListPresentation.tsx
```diff
@@ -1,46 +1,53 @@
...
     <DataView
       columns={[
         {
-          render: () => (
+          render: ({ payPeriod }) => (
             <Flex flexDirection="column">
-              <Text>Jul 5 - Jul 18, 2025</Text>
-              <Text>Regular Payroll</Text>
+              <Text>
+                {payPeriod?.startDate} - {payPeriod?.endDate}
```
dmortal, line 24: nit: I think this is a good candidate for format utility in `dateFormatting.ts` to ensure consistency in how we render dates

src/components/Payroll/PayrollList/PayrollListPresentation.tsx
```diff
@@ -1,46 +1,53 @@
...
         },
         {
           title: 'Run by',
-          render: () => <Text>Wed Jul 23, 2025</Text>,
+          render: ({ payrollDeadline }) => <Text>{payrollDeadline?.toLocaleDateString()}</Text>,
         },
         {
           title: 'Status',
-          render: () => <Badge>Ready to submit</Badge>,
+          render: ({ processed }) => <Badge>{processed ? 'Processed' : 'Unprocessed'}</Badge>,
```
dmortal, line 42: i18n?

src/components/Payroll/PayrollList/PayrollListPresentation.tsx
```diff
@@ -1,46 +1,53 @@
...
+                    ?.customName}
+              </Text>
             </Flex>
           ),
           title: 'Pay period',
         },
         {
           title: 'Run by',
-          render: () => <Text>Wed Jul 23, 2025</Text>,
+          render: ({ payrollDeadline }) => <Text>{payrollDeadline?.toLocaleDateString()}</Text>,
```
dmortal, line 38: We should probably start pulling current locale from our  locale provider (`useLocale.ts`) since it at some point we should expect partners to change that - don't know how close we are to that though. It's a nit comment

dmortal (approved):

src/components/Payroll/PayrollList/PayrollListPresentation.tsx
scally: @dmortal  Wondering also if even the `X - Y` format would hold up to scrutiny for non-US dates

src/components/Payroll/PayrollList/PayrollListPresentation.tsx
scally: Agreed; I think we should look at that holistically as opposed to here specifically

PR #519, opened 2025-09-03
feat: replacing suspence query for employee list pagination purposes

dmortal:
Removing useSuspense from EmployeeList
Adding inline spinner for now to the pagination control - pending design feedback

https://github.com/user-attachments/assets/8bbd660e-62ce-41c9-bbc9-e1a38e4b31a8


src/components/Common/PaginationControl/PaginationControl.tsx
```diff
@@ -47,6 +49,7 @@ const DefaultPaginationControl = ({
           </section>
         </div>
         <div className={style.paginationControlButtons}>
+          {isFetching && <InlineSpinner ariaLabel={t('labels.paginationFetchingLabel')} />}
```
serikjensen, line 52: Nit: we'll potentially need to adapter this since it's not using the core loading indicator. No need to do that now, but we'll probably need to provide that for adapter customers down the road

PR #526, opened 2025-09-05
feat: update payroll configuration to use prepare endpoint

serikjensen:
This updates the PayrollConfiguration component to use the prepare payroll endpoint instead of the get single payroll endpoint. This is because when we are editing payroll, we need to use the prepare endpoint first since the pay has not been calculated, see slack thread here: https://gustohq.slack.com/archives/C0327091JD8/p1757026541659839

This also updates the payroll list to only fetch unprocessed payrolls since we are currently working on the edit view.

Note that the gross pay has not been calculated when we use the prepare endpoint. This means that the gross pay is displaying as 0 for now. There is a separate ticket for this here https://gustohq.atlassian.net/browse/GWS-5505

## Proof of functionality

https://github.com/user-attachments/assets/004055bc-ed06-41a5-aa39-af5eb2d8387a

### DIFF
--- a/src/components/Payroll/PayrollList/PayrollList.tsx
+++ b/src/components/Payroll/PayrollList/PayrollList.tsx
@@ -1,5 +1,6 @@
 import { usePayrollsListSuspense } from '@gusto/embedded-api/react-query/payrollsList'
 import { usePaySchedulesGetAllSuspense } from '@gusto/embedded-api/react-query/paySchedulesGetAll'
+import { ProcessingStatuses } from '@gusto/embedded-api/models/operations/getv1companiescompanyidpayrolls'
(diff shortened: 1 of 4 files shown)


src/components/Payroll/PayrollList/PayrollList.tsx
```diff
@@ -12,6 +13,7 @@ interface PayrollListBlockProps extends BaseComponentInterface {
 export const PayrollList = ({ companyId, onEvent }: PayrollListBlockProps) => {
   const { data: payrollsData } = usePayrollsListSuspense({
     companyId,
+    processingStatuses: [ProcessingStatuses.Unprocessed],
```
scally, line 16: 💯

PR #533, opened 2025-09-09
feat: calculate via API endpoint on demand

scally:
poll for change of calculation status

delay Calculated event until we have confirmation of calculation from API

### DIFF
--- a/src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
+++ b/src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
@@ -1,8 +1,11 @@
 import { useState, useEffect } from 'react'
 import { useEmployeesListSuspense } from '@gusto/embedded-api/react-query/employeesList'
 import { usePayrollsPrepareMutation } from '@gusto/embedded-api/react-query/payrollsPrepare'
+import { usePayrollsGetSuspense } from '@gusto/embedded-api/react-query/payrollsGet'
+import { usePayrollsCalculateMutation } from '@gusto/embedded-api/react-query/payrollsCalculate'
 import type { Employee } from '@gusto/embedded-api/models/components/employee'
 import type { PayrollPrepared } from '@gusto/embedded-api/models/components/payrollprepared'
+import { PayrollProcessingRequestStatus } from '@gusto/embedded-api/models/components/payrollprocessingrequest'
 import { PayrollEditEmployee } from '../PayrollEditEmployee/PayrollEditEmployee'
 import { PayrollConfigurationPresentation } from './PayrollConfigurationPresentation'
 import type { BaseComponentInterface } from '@/components/Base/Base'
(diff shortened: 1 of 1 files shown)


src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
```diff
@@ -37,15 +35,22 @@ export const Root = ({ onEvent, companyId, payrollId, dictionary }: PayrollConfi
 
   const { LoadingIndicator } = useBase()
 
+  const { data: payrollData, refetch } = usePayrollsGetSuspense(
+    {
+      companyId,
+      payrollId,
+    },
+    { refetchInterval: 5_000 },
```
serikjensen, line 43: ahh this is handy

src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
```diff
@@ -66,8 +71,13 @@ export const Root = ({ onEvent, companyId, payrollId, dictionary }: PayrollConfi
...
   const onCalculatePayroll = async () => {
-    await mutate()
-    onEvent(componentEvents.RUN_PAYROLL_CALCULATED)
+    await calculatePayroll({
+      request: {
+        companyId,
+        payrollId,
+      },
+    })
+    await refetch()
```
serikjensen, line 80: Do we still need to refetch even after the update we made to invalidate the cache?

src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
```diff
@@ -81,7 +91,20 @@ export const Root = ({ onEvent, companyId, payrollId, dictionary }: PayrollConfi
...
+    payrollData.payrollShow?.processingRequest?.status ===
+    PayrollProcessingRequestStatus.CalculateSuccess
+
+  useEffect(() => {
+    if (isCalculated) {
+      onEvent(componentEvents.RUN_PAYROLL_CALCULATED)
+    }
+  }, [isCalculated, onEvent])
+
+  if (isPreparePayrollPending || isCalculating) {
```
serikjensen, line 107: Do we have a follow up ticket to display the loading UX aaron had in the design for when this is calculating? Do we also need to disable the calculate button in PayrollConfigurationPresentation while this is calculating?

serikjensen (approved):

src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
scally: You're right, we shouldn't. I'll double check

src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
scally: `const { data: payrollData, refetch } = usePayrollsGetSuspense(`
afaik I am in fact using a query hook that should suspend...

src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
scally: I removed it

PR #564, opened 2025-09-16
feat: implement new deductions empty state UI

shermango:
- Adds new deductions empty state view UI + basic transitions
  - will need to update new transition states/form capture strategy (new designs will no longer have a radio to designate including YES/ADD transition) in subsequent PR
- Updates dictionary with latest copies
- Adds new icon assets, plus-circle and coins-hand, for use
- Adds tests for empty state view

Screenshot 2025-09-16 at 2 21 42 PM

figma: https://www.figma.com/design/6dxOSiONDiJoa9zY1wOwbs/Frontend-SDK-Partner-Design-File?node-id=5409-26944&t=BXR9ZMisPJETNRZN-0 
jira: https://gustohq.atlassian.net/browse/GWS-5476


src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
```diff
@@ -0,0 +1,22 @@
+.emptyStateContainer {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  min-height: 411px;
```
shermango, line 5: this was just pulled from the figma spec

jeffredodd: You should look and figure out what variable this most aligns to, I feel like we might have defined this before

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.coinHandsIconContainer {
+  border: 1px solid var(--g-colorBorder);
+  border-radius: toRem(8);
+  padding: 8px;
```
jeffredodd, line 11: We should probably be using toRem here instead of hard coded pixels

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.plusCircleIcon {
+  margin-right: toRem(4);
```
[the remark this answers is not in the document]
jeffredodd: We should always be using toRem. I don't know if we import helper functions for other modules??? Should just work

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
serikjensen: Would also recommend using the `toRem` utility which should be available in the global scope. I also tend to round down a pixel in cases like this when the number is especially ugly but that's optional 😂 

```suggestion
  min-height: toRem(410);
```

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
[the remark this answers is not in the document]
serikjensen: Helpers are available globally and get prepended to all scss with vite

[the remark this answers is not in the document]
jeffredodd: Yeah copilot not too smart some days. Sometimes it can catch valid things but more often it doesn't. Feel free to resolve.

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
```diff
@@ -0,0 +1,22 @@
...
+}
+
+.addDeductionButton {
+  &#{&}#{&}#{&} {
```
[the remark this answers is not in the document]
serikjensen: So one oddity with SDK development: designs aren't pixel perfect. The design system used in figma doesn't completely align to what we have in code. We need to eventually get those in sync, it's work that's been hard to prioritize.

In the meantime, don't fret about overriding the core components that you bring in via context from the ui directory (ex. Button, Tabs, Table, TextInput, etc etc). These styles are all centralized and when we do get bandwidth to sync the style updates they will get them for free.

I'd recommend removing this override in the meantime

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
shermango: I checked `src/contexts/ThemeProvider/theme.ts` but there wasn't  any defined variables for `min-height` -- is there any other files I should be checking? I looked through the base scss files but didnt find any thing similar either

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.tsx
```diff
@@ -0,0 +1,103 @@
...
+import PlusCircleIcon from '@/assets/icons/plus-circle.svg?react'
+
+const IncludeDeductionsSchema = z.object({ includeDeductions: z.enum(['Yes', 'No']) })
+export type IncludeDeductionsPayload = z.output<typeof IncludeDeductionsSchema>
+
+interface IncludeDeductionsFormProps extends CommonComponentInterface<'Employee.Deductions'> {
+  employeeId: string
+}
+
+export function IncludeDeductionsFormEmptyState(
```
serikjensen, line 29: I'm a little confused on the naming with the implementation here? If this is just an empty state component it can probably just include the empty state UI pieces without needing to implement a schema, form, or the base component.

If this is eventually going to grow in functionality to contain the form, we should probably update the file naming.

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
jeffredodd: Hmm, yeah I was looking at EmptyData.tsx and its styles and it doesn't look like we've ever defined Min height for those either.

This is probably ok for now but I'd check with Steve, our theme master

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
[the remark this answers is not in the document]
jeffredodd: Hmmm, honestly, why did we need to change the border color? Typically we don't take these kinds of changes that are one off.... 

If its just that the figma looks slightly different that happens sometimes so maybe thats all this is?

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.tsx
```diff
@@ -0,0 +1,103 @@
...
+          ) : (
+            <Grid gridTemplateColumns="1fr">
+              <Components.Heading as="h2">{t('pageTitle')}</Components.Heading>
+              <Components.Text variant="supporting">
+                {t('includeDeductionsDescriptionV2')}
+              </Components.Text>
+              <Components.Text weight="bold" size="lg">
+                {t('includeDeductionsSubtitle')}
+              </Components.Text>
+              <Components.Card className={styles.emptyStateContainer}>
```
serikjensen, line 75: I'd avoid using card here, this is likely just a layout specific container and could be styled here as a one off. Card will get overwritten with component adapters as something more traditionally card like which could be more responsive leaning

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.module.scss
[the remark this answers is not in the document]
jeffredodd: ☝️ What Steve said, sorry Steve I wrote this before I saw your comment

src/components/Employee/Deductions/IncludeDeductionsForm/IncludeDeductionsFormEmptyState.tsx
shermango: renamed component to EmptyState to better match functionality and removed form methods as not required

jeffredodd (approved):

PR #570, opened 2025-09-17
GWS-5534 add payroll type and pay date to PayrollList

scally:
* adjust date formatting to be in line with current figma
* use locales for date formatting
Screenshot 2025-09-17 at 12 36 49 PM


src/components/Payroll/PayrollList/PayrollList.tsx
```diff
@@ -1,11 +1,24 @@
...
 import { usePaySchedulesGetAllSuspense } from '@gusto/embedded-api/react-query/paySchedulesGetAll'
 import { ProcessingStatuses } from '@gusto/embedded-api/models/operations/getv1companiescompanyidpayrolls'
+import type { Payroll } from '@gusto/embedded-api/models/components/payroll'
 import { PayrollListPresentation } from './PayrollListPresentation'
+import type { PayrollType } from './types'
 import type { BaseComponentInterface } from '@/components/Base'
 import { BaseComponent } from '@/components/Base'
 import { componentEvents } from '@/shared/constants'
 
+const selectPayrollType = ({
```
scally, line 11: There was an interesting code design decision here where we need to interpret payroll type from the payroll data because the type doesn't exist on the API payload.

What I chose to do was to have the Presentation layer expect the problem to be solved before it received it as Props. PayrollList then takes a payroll and creates a projection from the payroll with synthesized props from the API payload. The synthesized prop has to be extracted to `types` since both PayrollList and PayrollLisPresentation both need to depend on it.

src/components/Payroll/PayrollList/PayrollListPresentation.tsx
```diff
@@ -1,15 +1,23 @@
...
 import type { PayScheduleList } from '@gusto/embedded-api/models/components/payschedulelist'
 import { useTranslation } from 'react-i18next'
+import type { PayrollType } from './types'
 import { DataView, Flex } from '@/components/Common'
 import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentContext'
 import { useI18n } from '@/i18n'
+import { parseDateStringToLocal } from '@/helpers/dateFormatting'
+import { useLocale } from '@/contexts/LocaleProvider'
+
+interface PresentationPayroll extends Payroll {
```
scally, line 11: this is the payroll projected with some synthesized values

src/components/Payroll/PayrollList/types.ts
```diff
@@ -0,0 +1 @@
+export type PayrollType = 'External' | 'Off-Cycle' | 'Regular'
```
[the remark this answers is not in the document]
jeffredodd: ty

jeffredodd (approved):
LGTM, I don't see anything scary on first glance. Thanks for the types.

serikjensen (approved):

PR #581, opened 2025-09-19
feat: add new Dialog component

serikjensen:
This PR adds a Dialog component consistent with the Figma here https://www.figma.com/design/6dxOSiONDiJoa9zY1wOwbs/Frontend-SDK-Partner-Design-File?node-id=10849-100204&t=n8Ys6vSxrlJ3BUOp-4

This uses the HTML dialog component internally which should take care of the base accessibility (keyboard, screenreader, etc) as well as top level positioning.

## Proof of functionality

https://github.com/user-attachments/assets/f625debb-ff08-4382-b862-4a02c6c2d20a


jeffredodd (approved):
LGTM! I'm assuming the accessibility works on this out of the box but we should probably triple check that we get or meet the same `role="dialog"` functionality. I assume we would....

PR #586, opened 2025-09-22
Introduce PayrollLanding component as first step in RunPayrollFlow

jeffredodd:
## Summary
Introduces a **Payroll Landing** component that centralizes **Payroll History** and **Run Payroll** flows inside tabbed navigation.  
Also resolves type errors in the `PayrollListPresentation` component.  
These errors occurred because the embedded API model now exposes nullable date fields, but `parseDateStringToLocal` expects string parameters.

## Changes
- Centralized **Payroll History** and **Run Payroll** flows inside tabs on the new **Payroll Landing** component.
- Updated the entry point to the **Run Payroll** flow to align with the new tabbed structure.

## Screenshots
Screenshot 2025-09-22 at 9 32 25 AM
Screenshot 2025-09-22 at 9 40 55 AM

### DIFF
--- a/src/components/Payroll/PayrollLanding/PayrollLanding.test.tsx
+++ b/src/components/Payroll/PayrollLanding/PayrollLanding.test.tsx
@@ -56,44 +56,10 @@ describe('PayrollLanding', () => {
   })
 
   describe('tab navigation', () => {
-    it('switches to payroll history component when tab is clicked', async () => {
+    it('switches between tabs correctly', async () => {
       const user = userEvent.setup()
       renderWithProviders(<PayrollLanding {...defaultProps} />)
 
-      // Wait for initial render
-      await waitFor(() => {
-        expect(screen.getByRole('tab', { name: 'Run payroll' })).toBeInTheDocument()
-      })
-
-      // Click on payroll history tab
-      const payrollHistoryTab = screen.getByRole('tab', { name: 'Payroll history' })
-      await user.click(payrollHistoryTab)
-
-      // Check that payroll history tab is now selected
-      await waitFor(() => {
-        expect(screen.getByRole('tab', { name: 'Payroll history' })).toHaveAttribute(
-          'aria-selected',
-          'true',
-        )
-      })
-
-      expect(screen.getByRole('tab', { name: 'Run payroll' })).toHaveAttribute(
-        'aria-selected',
-        'false',
-      )
-
-      // Verify PayrollHistory component content is now visible
-      await waitFor(() => {
-        expect(screen.getByRole('heading', { name: 'Payroll history' })).toBeInTheDocument()
-      })
-      expect(screen.queryByRole('heading', { name: 'Run payroll' })).not.toBeInTheDocument()
-    })
-
-    it('switches back to run payroll component when tab is clicked', async () => {
-      const user = userEvent.setup()
-      renderWithProviders(<PayrollLanding {...defaultProps} />)
-
-      // Wait for initial render
       await waitFor(() => {
         expect(screen.getByRole('tab', { name: 'Run payroll' })).toBeInTheDocument()
       })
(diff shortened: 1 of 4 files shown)


src/components/Payroll/PayrollList/PayrollListPresentation.tsx
```diff
@@ -24,105 +23,112 @@ export const PayrollListPresentation = ({
...
   const { t } = useTranslation('Payroll.PayrollList')
   const { locale } = useLocale()
 
   return (
-    <DataTable
-      emptyState={() => (
-        <Flex flexDirection="column" alignItems="center" gap={24}>
-          <div className={styles.doneIcon}>✓</div>
-          <Text>{t('emptyState')}</Text>
+    <Flex flexDirection="column" gap={16}>
```
jeffredodd, line 32: Wrapped this in the same flex layout I've been using in Payroll History for consistent spacing

.ladle/helpers/I18nWrapper.tsx
```diff
@@ -0,0 +1,18 @@
+import { I18nextProvider } from 'react-i18next'
```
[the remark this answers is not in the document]
serikjensen: Is there a way we can set this on all stories? i thought we already had that configured somewhere?

src/components/Payroll/PayrollLanding/PayrollLanding.stories.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  return (
+    <Tabs
+      tabs={tabs}
+      selectedId={selectedTab}
+      onSelectionChange={setSelectedTab}
+      aria-label="Payroll navigation"
+    />
+  )
+}
```
serikjensen, line 155: Looks like this should actually be importing the `PayrollLanding` component instead of recreating the tabs here?

src/components/Payroll/PayrollLanding/PayrollLanding.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+      <button
+        onClick={() => {
+          onEvent('test-event', {})
+        }}
+      >
+        Test Event
+      </button>
+    </div>
+  ),
+}))
```
serikjensen, line 50: Is it preferable to mock these? i wonder if it would be more aligned with our testing approach to configure the test fixtures so they could get mocked with msw

src/components/Payroll/PayrollLanding/PayrollLanding.test.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+      await waitFor(() => {
+        expect(screen.getByTestId('run-payroll')).toBeInTheDocument()
+      })
+
+      // Check that translated tab labels are used
+      expect(screen.getByRole('tab', { name: 'Run payroll' })).toBeInTheDocument()
+      expect(screen.getByRole('tab', { name: 'Payroll history' })).toBeInTheDocument()
+    })
+  })
```
serikjensen, line 189: Nit: this kinda gets tested implicitly already with the above since we query by these to select the tabs

src/components/Payroll/PayrollLanding/PayrollLanding.tsx
```diff
@@ -0,0 +1,63 @@
...
+      label: t('tabs.runPayroll'),
+      content: (
+        <RunPayroll
+          companyId={companyId}
+          Configuration={PayrollConfiguration}
+          List={PayrollList}
+          Overview={PayrollOverview}
+          EditEmployee={PayrollEditEmployee}
+          onEvent={onEvent}
+        />
```
serikjensen, line 45: This is going to cause an issue where the entire payroll experience is in a single tab. We actually just want to render `PayrollList` for the content here.

Then, we would update the RunPayroll component to use this `PayrollLanding` component instead. This will emit events to RunPayroll which can manage transitions between the blocks

serikjensen (commented):
Looking good! biggest thing is replacing `RunPayroll` in `PayrollLanding` with the `PayrollList` component and then replacing `PayrollList` in `RunPayroll` with the `PayrollLanding` component

PR #608, opened 2025-09-25
Da/payroll overview

dmortal:
Payroll paystub pdf
Cancel payroll

https://github.com/user-attachments/assets/3a8308cc-dc91-45c1-b9c8-367e0bb54900


package-lock.json
```diff
@@ -15818,15 +15887,15 @@
       }
     },
     "node_modules/react-aria-components": {
-      "version": "1.12.1",
-      "resolved": "https://registry.npmjs.org/react-aria-components/-/react-aria-components-1.12.1.tgz",
-      "integrity": "sha512-UTn2y2Pr1QuapXLRoGE/GWHrjcZZSuMf+zhbJjInOHgS+MC4hqXiINufvjQrdjQDzS1llc2aepP9op6+z6QSxA==",
+      "version": "1.12.2",
```
jeffredodd, line 15890: So I think this is why your build is breaking, I thought I put a pin in this, but I believe this verison is busted. I saw React-aria people in Adobe logging issues about tables not rendering. This is the root cause of your issue. https://github.com/Gusto/embedded-react-sdk/pull/556

PR #609, opened 2025-09-25
Only poll for new processing request information when payroll is still calculating

scally:
only poll for new processing request information when payroll is still calculating


src/components/Payroll/PayrollConfiguration/PayrollConfiguration.tsx
```diff
@@ -42,8 +49,11 @@ export const Root = ({
     {
       companyId,
       payrollId,
+      include: ['taxes', 'benefits', 'deductions'],
```
scally, line 52: Consistent now with Payroll Overview

dmortal (approved):

PR #625, opened 2025-09-30
feat: adding alerts to flow context

dmortal:
Passing alerts between payroll components through context


src/components/Payroll/PayrollOverview/PayrollOverviewPresentation.tsx
```diff
@@ -509,8 +512,15 @@ export const PayrollOverviewPresentation = ({
           )}
         </Flex>
       </Flex>
-      {/* TODO: when is this actually saved? */}
-      <Alert label={t('alerts.progressSaved')} status="success"></Alert>
+      {alerts && alerts.length > 0 && (
```
shermango, line 515: think `alerts?.length` should work here -- kind of confused what would happen if there is no `content` value in the `alert` though, do we just show an empty spacer/card? 

for key could possibly use the `content` as well assuming its unique per alert or use that compound key that Copilot is suggesting?

PR #1520, opened 2026-04-09
chore: upgrade @gusto/embedded-api from 0.12.4 to 0.12.5

krisxcrash:
## Summary

- Upgrades `@gusto/embedded-api` from `0.12.4` to `0.12.5`, adapting to all breaking changes in the new version
- Renames response fields across 35 files to match the updated API client types (`payScheduleList` → `paySchedules`, `payScheduleObject` → `paySchedule`, `responseBodies` → `taxRequirementStatesList`, `jobList` → `jobs`, `object` → `paySchedulePreview`)
- Migrates mutation request body params from generic `requestBody` to named types (`federalTaxDetailsUpdate`, `payScheduleCreateRequest`, `payScheduleUpdateRequest`, `i9AuthorizationRequestBody`, `jobsCreateRequestBody`, `jobsUpdateRequestBody`)
- Moves type imports from `models/operations/` to `models/components/` modules where the API client reorganized them
- Handles the new `RFCDate` class that replaces plain string dates for anchor dates, pay periods, and related fields — including guarding against empty-string construction when form fields are unpopulated
- Updates test fixtures and MSW mock handlers to match the new response shapes and request body types

## Test plan

- [x] TypeScript compilation passes cleanly (`npx tsc --noEmit` — 0 errors)
- [x] All 2110 unit tests pass across 170 test files (`npm run test -- --run`)
- [x] Production build succeeds (`npm run build`)
- [x] Pre-commit hooks (lint, format, build, commitlint) all pass
- [x] Verify in the Flows that PaySchedule, FederalTaxes, StateTaxes, Compensation, and Payroll components render correctly
- [x] E2E smoke test against gws-flows (if available)


src/components/Company/PaySchedule/_parts/Edit.tsx
```diff
@@ -103,7 +103,10 @@ export const Edit = () => {
                   options={payPeriodPreview.map((period, index) => {
                     return {
                       value: String(index),
-                      label: dateFormatter.formatPayPeriodRange(period.startDate, period.endDate),
+                      label: dateFormatter.formatPayPeriodRange(
+                        period.startDate.toString(),
+                        period.endDate.toString(),
+                      ),
```
serikjensen, line 109: I would make sure to verify this one for regressions by running company pay schedule

src/components/Company/PaySchedule/PaySchedule.tsx
```diff
@@ -105,8 +105,8 @@ const Root = ({ companyId, children, defaultValues }: PayScheduleProps) => {
     {
       companyId,
       frequency: allValues.frequency,
-      anchorPayDate: formattedAnchorPayDate,
-      anchorEndOfPayPeriod: formattedAnchorEndOfPayPeriod,
+      anchorPayDate: new RFCDate(formattedAnchorPayDate || '1970-01-01'),
+      anchorEndOfPayPeriod: new RFCDate(formattedAnchorEndOfPayPeriod || '1970-01-01'),
```
serikjensen: Where did the fallback for 1970-01-01 come from? i'd have claude scrutinize this a bit. My hunch is that we'd want behavior aligned with the previous even if it means doing something like

```
anchorPayDate: formattedAnchorPayDate ? new RFCDate(formattedAnchorPayDate) : undefined
```

we just want to make sure behavior is the same without regressions

src/components/Company/PaySchedule/PaySchedule.tsx
```diff
@@ -148,10 +148,12 @@ const Root = ({ companyId, children, defaultValues }: PayScheduleProps) => {
...
     reset({
-      frequency: schedule.frequency as Frequency,
-      anchorPayDate: schedule.anchorPayDate ? new Date(schedule.anchorPayDate) : undefined,
+      frequency: schedule.frequency as PayScheduleInputs['frequency'],
+      anchorPayDate: schedule.anchorPayDate
+        ? new Date(schedule.anchorPayDate.toString())
+        : undefined,
       anchorEndOfPayPeriod: schedule.anchorEndOfPayPeriod
-        ? new Date(schedule.anchorEndOfPayPeriod)
+        ? new Date(schedule.anchorEndOfPayPeriod.toString())
```
serikjensen, line 158: Same on these ones

src/components/UNSTABLE_Hooks/hooks/usePayScheduleForm/PayScheduleForm.tsx
```diff
@@ -143,8 +139,8 @@ function PayScheduleFormRoot({ onEvent, dictionary, ...hookProps }: PayScheduleF
                         options={payPeriodPreview.map((period, index) => ({
                           value: String(index),
                           label: dateFormatter.formatPayPeriodRange(
-                            period.startDate,
-                            period.endDate,
+                            period.startDate.toString(),
+                            period.endDate.toString(),
```
serikjensen, line 143: We'll want to make sure that we test the hook implementation for regressions as well. That should be in the sidebar for Hooks -> PayScheduleForm

This one is going out to collective today/tomorrow so we should pay extra attention to any regressions there. Claude can help provide a manual test plan to go through the cases and expected behavior

Image

src/components/UNSTABLE_Hooks/hooks/usePayScheduleForm/usePayScheduleForm.tsx
```diff
@@ -159,15 +163,16 @@ export function usePayScheduleForm({
     {
       companyId,
       frequency: watchedFrequency,
-      anchorPayDate: formattedAnchorPayDate,
-      anchorEndOfPayPeriod: formattedAnchorEndOfPayPeriod,
+      anchorPayDate: new RFCDate(formattedAnchorPayDate || '1970-01-01'),
+      anchorEndOfPayPeriod: new RFCDate(formattedAnchorEndOfPayPeriod || '1970-01-01'),
```
serikjensen: Same feedback as above, i'm not sure that the 1970-01-01 fallback is functionally equivalent to what's there now

serikjensen (commented):
Looking good! left some questions around behavioral differences. also i would def test for regressions thoroughly on both the existing company pay schedule and the new pay schedule hook since that is going out to collective today/tomorrow

src/components/Company/PaySchedule/_parts/Edit.tsx
krisxcrash: Verified — `RFCDate.toString()` returns the same `YYYY-MM-DD` string format that the old API returned as plain strings, so `formatPayPeriodRange(period.startDate.toString(), period.endDate.toString())` is functionally identical to the previous `formatPayPeriodRange(period.startDate, period.endDate)`. Will verify in the flow as part of the manual test plan below.

src/components/Company/PaySchedule/PaySchedule.tsx
krisxcrash: Good call — the `1970-01-01` fallback was a workaround for `RFCDate` throwing on empty strings, but you're right that it's not behaviorally equivalent.

Updated to follow your suggestion. The API type requires `anchorPayDate: RFCDate` (non-optional), so I used a type assertion since the query is always disabled when dates are empty:

```ts
anchorPayDate: formattedAnchorPayDate
  ? new RFCDate(formattedAnchorPayDate)
  : (undefined as unknown as RFCDate),
```

This way:
- When dates are present → valid `RFCDate` is constructed (query runs)
- When dates are empty → `undefined` (query is disabled via `enabled: Boolean(...)`, never fires)
- No fake epoch date that could accidentally be sent to the API

Applied the same fix in both `PaySchedule.tsx` and `usePayScheduleForm.tsx`. TypeScript compilation and all 42 PaySchedule-related tests pass.

src/components/Company/PaySchedule/PaySchedule.tsx
krisxcrash: These `.toString()` calls on `RFCDate` produce `YYYY-MM-DD` strings identical to what the old API returned as plain strings. `new Date('2024-01-15')` behaves the same whether the input was originally a string or came from `RFCDate.toString()`. Will verify in the flow.

src/components/UNSTABLE_Hooks/hooks/usePayScheduleForm/PayScheduleForm.tsx
krisxcrash: I totally understand and will test this thoroughly in the flows. Def couldn't hurt to have a second pair of eyes on it though.

src/components/UNSTABLE_Hooks/hooks/usePayScheduleForm/usePayScheduleForm.tsx
krisxcrash: Fixed — replaced the `1970-01-01` fallback with `undefined` (type-asserted) here as well, same approach as `PaySchedule.tsx`. The `previewEnabled` guard ensures the query never fires when dates are empty, so the `undefined` is never actually sent to the API. See the detailed explanation on the other comment thread.

src/components/UNSTABLE_Hooks/hooks/usePayScheduleForm/PayScheduleForm.tsx
krisxcrash: dates appear to be formatted and logged correctly: 
image

src/components/Company/PaySchedule/PaySchedule.tsx
krisxcrash: image

serikjensen (approved):

PR #1551, opened 2026-04-14
feat: add PolicySettings presentation component for time-off policy creation

krisxcrash:
## Summary
- Adds `PolicySettingsPresentation` component for the time-off policy settings screen (SDK-577)
- Implements toggle-controlled number input fields for accrual maximum, balance maximum, carry over limit, waiting period, and payout on dismissal
- Includes SCSS module with CSS grid layout, scoped theme variable overrides, and responsive toggle alignment
- Adds i18n translations matching Figma design specs
- Includes Storybook stories for hours_worked, fixed, and edit mode variants
- Exports new component and types from the UNSTABLE_TimeOff barrel

## Test plan
- [x] Verify Storybook renders correctly for all three story variants (HoursWorkedVariant, FixedVariant, EditMode)
- [x] Confirm toggle interactions enable/disable corresponding input fields
- [x] Verify form submission passes correct data via onContinue callback
- [x] Visual comparison against Figma design for layout, spacing, and colors
- [x] Unit tests will follow in a separate PR

Resolves SDK-577 (presentation layer only — wiring/functionality is a follow-up)


src/components/UNSTABLE_TimeOff/PolicySettings/PolicySettings.module.scss
```diff
@@ -0,0 +1,21 @@
...
+
+.divider {
+  border: none;
+  border-top: 1px solid var(--g-inputBorderColor);
+  margin: toRem(4) 0;
+}
+
+.settingField :global(.react-aria-NumberField) {
+  max-width: toRem(200);
+}
```
serikjensen: Nit: why do we need this aria numberfield style selector? NumberField should already take up the width of the container

additionally, this style would break for component adapters if we are relying on setting the style on the react aria input since partners may not be using that.

all field components should just be assumed to take up the width of the container and then we should be able to size the container for layout

serikjensen (approved):

serikjensen (approved):

PR #1559, opened 2026-04-14
feat: wire PolicySettings into TimeOff state machine (SDK-579)

krisxcrash:
## Summary

- **Wire PolicySettings container**: Replace the stub with a full container component that fetches policy data via `useTimeOffPoliciesGetSuspense`, maps API fields to form defaults and back, handles the "unlimited skip" logic (auto-advances past the screen), and emits state machine events on save/back/error.
- **Add BACK transition**: New `TIME_OFF_POLICY_SETTINGS_BACK` event and state machine transition to navigate from PolicySettings back to PolicyDetailsForm cleanly.
- **Fix silent form validation on PolicyConfigurationForm**: Added `errorMessage` props to all required fields so users see validation feedback when submitting with empty fields (previously "nothing happened" on submit).

## Test plan

- [x] All 28 existing PolicyConfigurationForm tests pass
- [x] New state machine test for BACK transition passes
- [x] New PolicySettings container unit tests added (API calls, event emission, unlimited skip, field mapping)
- [x] Formatting, linting, and TypeScript checks pass
- [x] Verify in Storybook: PolicyConfigurationForm shows validation errors on empty submit
- [x] Verify in Storybook: PolicySettings renders correctly for hours_worked and fixed variants


src/components/UNSTABLE_TimeOff/PolicySettings/PolicySettings.tsx
```diff
@@ -8,32 +16,138 @@ export interface PolicySettingsProps extends BaseComponentInterface {
...
+
+    void updateTimeOffPolicy({
+      request: {
+        timeOffPolicyUuid: policyId,
+        requestBody: { complete: true, version },
+      },
+    }).then(() => {
+      onEvent(componentEvents.TIME_OFF_POLICY_SETTINGS_DONE)
+    })
+  }, [accrualCategory, onEvent, policyId, updateTimeOffPolicy, version])
```
serikjensen: I think we need to rework this. Running it through an effect is going to be brittle. Also we need to wrap things like this in base submit handler so that we can get proper error handling.

Let's chat about it when you get back from OOO

serikjensen (changes requested):

src/components/UNSTABLE_TimeOff/PolicySettings/PolicySettings.tsx
```diff
@@ -8,32 +16,117 @@ export interface PolicySettingsProps extends BaseComponentInterface {
...
+
+        onEvent(componentEvents.TIME_OFF_POLICY_SETTINGS_DONE)
+      })
+    },
+    [baseSubmitHandler, onEvent, policyId, updateTimeOffPolicy, version],
+  )
+
+  const handleBack = useCallback(() => {
+    onEvent(componentEvents.TIME_OFF_POLICY_SETTINGS_BACK)
+  }, [onEvent])
```
serikjensen: Nit: i don't know that these useCallbacks are actually getting us much here since onEvent is not memoized

src/components/UNSTABLE_TimeOff/PolicySettings/PolicySettings.tsx
```diff
@@ -8,32 +16,117 @@ export interface PolicySettingsProps extends BaseComponentInterface {
...
+    async (data: PolicySettingsFormData) => {
+      await baseSubmitHandler(data, async () => {
+        await updateTimeOffPolicy({
+          request: {
+            timeOffPolicyUuid: policyId,
+            requestBody: buildUpdateRequestBody(data, version),
+          },
+        })
+
+        onEvent(componentEvents.TIME_OFF_POLICY_SETTINGS_DONE)
```
serikjensen: I think we should update this with the result of updateTimeOffPolicy for the event. ex. do


```suggestion
        const result = await updateTimeOffPolicy({
          request: {
            timeOffPolicyUuid: policyId,
            requestBody: buildUpdateRequestBody(data, version),
          },
        })

        onEvent(componentEvents.TIME_OFF_POLICY_SETTINGS_DONE, result.data)
```

i put result.data here, but it's really whatever the inner value of the api result is. The top level result will include things like httpMeta etc for the http call. We want to provide the partner with the data response from the api endpoint. So if you log that result, whatever entry contains the response for updating time off policy is what you want to use here. Lmk if that makes sense!

serikjensen:
** lines 100–102** — Two non-null assertions that could silently throw a runtime `TypeError`:

```ts
const policy = policyResponse.timeOffPolicy!
const version = policy.version!
```

Suspense only catches thrown promises/errors — it won't protect against a successful API response that returns `null` for either field. `version` is especially risky since it feeds directly into the optimistic concurrency check on the PUT; sending `undefined` there would likely produce a confusing API rejection rather than a clear error.

Suggested fix:
```ts
const policy = policyResponse.timeOffPolicy
if (!policy) throw new Error('Unexpected response: missing timeOffPolicy')
const version = policy.version ?? ''
// or throw if version is missing, since the PUT will fail without it:
// if (!policy.version) throw new Error('Missing version — cannot safely update policy')
```

*(Claude)*

serikjensen (approved):

PR #1575, opened 2026-04-16
feat: merge sdk-app and prototype-app

aaronlee777:
## Summary

Merges the separate `prototype-app` into `sdk-app` as a dual-mode application. The sdk-app now supports two modes — **Preview** (existing SDK component browser) and **Design** (prototype sandbox for designing new UI components with the SDK component library) — sharing one data source and configuration. The `prototype-app/` directory is removed entirely.

## Changes

- Added URL-based mode switching (`/design/*` routes) with a Preview/Design toggle in the TopBar
- Moved prototype code (component showcase, contractor profile) to `sdk-app/src/design/`
- Created `DesignLayout` wrapper that provides `GustoProvider` for design routes
- Restyled TopBar to white background with only title + mode switcher
- Added Footer bar for environment badges (API, SDK, demo type), company name, token status, and Settings button
- Mode-aware Sidebar that shows SDK components in Preview and prototypes in Design
- Replaced text arrow (`▾`) with SVG `CaretRightIcon` in sidebar
- Deleted `prototype-app/` directory and all duplicated configuration
- Added breakpoint switcher toggle (fixed bottom-right in Design mode) to preview components at Phone (375px), Small (640px), Medium (768px), Large (1024px), and Full width — aligned with SDK `BREAKPOINTS_VALUES` from `src/shared/constants`
- Fixed stale company name display in TopBar when switching company IDs

## Related

- Builds on the contractor management prototype work from earlier commits on this branch

## Testing

- `npm run sdk-app` and verify:
  - Preview mode (`/`): SDK component browser works as before
  - Design mode (`/design`): Shows prototype grid, component showcase, and contractor profile
  - Mode switcher toggles between modes
  - Footer shows environment badges, company name, token status
  - Settings panel opens from Footer in both modes
  - Sidebar updates items based on current mode
  - Breakpoint switcher in Design mode resizes the content container across all preset widths

---


prototype-app/scripts/start.ts
```diff
@@ -0,0 +1,142 @@
...
+
+  child.on('exit', code => {
+    process.exit(code ?? 0)
+  })
+}
+
+main().catch((err: unknown) => {
+  console.error('Fatal error:', err)
+  process.exit(1)
+})
```
serikjensen: So my overall inclination here is that if we are going to go this route with live data we might consider just creating a designer mode in SDK app that can pull in your prototypes. You could have it be a toggle up top or just have a prototypes section at the bottom of the side bar

I know claude makes it cheap to spin up apps like this, but i worry about the amount of codebase bloat we are going to accumulate trying to have two distinct custom apps in our repository

It's also twice the maintenance overhead etc

If we must have them be separate apps and we can't consolidate to a single development app, i would recommend having claude at a _minimum_ consolidate the start scripts so we are only doing this setup once in sdk rather than multiple times

prototype-app/src/useDemoManager.ts
```diff
@@ -0,0 +1,63 @@
...
+    return () => {
+      clearInterval(pollRef.current)
+    }
+  }, [checkTokenHealth])
+
+  return {
+    ...state,
+    checkTokenHealth,
+  }
+}
```
serikjensen: same here, we should have this consolidated in central utilities, but my overall rec is for you to just have the protoypes running in SDK app so we don't have to duplicate this infrastructure all over the place.

prototype-app/vite.config.ts
```diff
@@ -1,22 +1,96 @@
...
+      },
+    },
+    css: {
+      preprocessorOptions: scssPreprocessorOptions,
+    },
+    server: {
+      port: 5300,
+      strictPort: false,
+      open: true,
+      proxy: proxyConfig,
```
serikjensen: More config boilerplate we're not going to want to duplicate

serikjensen (changes requested):
So left context in the comments, but i think we need to eithere

* Update to have the sdk-app have a design mode and it can have your prototypes (recommended)

_or_

* find a way to consolidate all the boilerplate for spinning up the application such that we have confidence they won't get out of sync

i'd recommend the first path to avoid the bloat of having two applications with basically identical infrastructure in the codebase

aaronlee777:
@serikjensen I am in full agreement with all of your feedback.

sdk-app/src/design/useCompanyName.ts
```diff
@@ -0,0 +1,18 @@
+import { useState, useEffect } from 'react'
+
+export function useCompanyName(companyId: string) {
+  const [name, setName] = useState<string>('')
+
+  useEffect(() => {
+    if (!companyId) return
+
+    fetch(`/api/v1/companies/${companyId}`, { signal: AbortSignal.timeout(10000) })
```
serikjensen, line 10: You should be able to just have claude use the @gusto/embedded-api package instead of needing to call fetch directly

serikjensen (approved):
One comment recommending to use embedded api, but overall this looks good!

PR #1576, opened 2026-04-16
feat: employee home address management and dashboard integration

dmortal:
## Summary

Adds the **Employee.HomeAddress.Management** experience: manage current home address, show pending future-dated changes, browse and edit or delete historical addresses, and add new addresses with an effective date. Extends **`useHomeAddressForm`** so management flows can force create vs update, target a specific address row, and control default values. Wires the employee dashboard state machine to this component and emits home-address lifecycle events (created, updated, deleted).

## Changes

- New management UI (`HomeAddress`, `HomeAddressView`) with modals, history table, and pending-address handling; helpers for pending future address selection.
- **`useHomeAddressForm`** enhancements for submission mode, empty vs current defaults, and update-by-UUID; partner hook field tweaks where needed for modals.
- Dashboard and exports/registry updates so partners can render the component from embedded flows; new i18n namespace and event constants.
- Vitest coverage for the management component (MSW-backed) and pending-address helpers; Employee list folder naming alignment.

## Testing

- `npm run test -- --run src/components/Employee/HomeAddress/management/HomeAddress.test.tsx`
- `npm run test -- --run src/components/Employee/HomeAddress/management/getPendingFutureHomeAddress.test.ts`
- Both commands pass locally.
- Manual: run `npm run sdk-app` and exercise the employee home address management entry from the dashboard.


src/components/Employee/Profile/shared/useHomeAddressForm/useHomeAddressForm.tsx
```diff
@@ -50,11 +50,44 @@ export interface UseHomeAddressFormProps {
...
+  /**
+   * When set with `submissionMode: 'alwaysUpdate'`, the form defaults and PUT target the home address with this UUID
+   * (e.g. editing a row from address history). When omitted, the active address is used.
+   */
+  updateTargetUuid?: string
+  /**
+   * Increment when opening a create/edit modal so the form resets even when the target address and
+   * default field values match the previous open (e.g. reopening the same row after cancel).
+   */
+  formSessionId?: number
```
serikjensen: I would leave the same comment here as i did on the work address https://github.com/Gusto/embedded-react-sdk/pull/1576

i think we can keep hooks simple by having the `homeAddressId` present, that will determine if it is in create/update mode, it'll also set the default values for the home address when present. We could create a convenience useActiveHomeAddressForm for profile that would get the active home address id by default and pass it to the hook.

That should allow us to remove these props.

I would also recommend using a `key` in the UI in order to handle the form reset/unmount the component vs. needing the formSessionId to be passed to the hook

src/components/Employee/HomeAddress/management/useHomeAddressManagement.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { firstLastName } from '@/helpers/formattedStrings'
+import { SDKInternalError } from '@/types/sdkError'
+import { componentEvents, type EventType } from '@/shared/constants'
+
+export interface UseHomeAddressManagementParams {
+  employeeId: string
+  onEvent: OnEventType<EventType, unknown>
+}
+
+export interface UseHomeAddressManagementResult {
```
serikjensen: You _should_ now be able to extend BaseHookReady. it looks like we may need to add `actions` to that one as well since this looks like it has actions. We'd want to update to use the same interface to keep these consistent tho

Even if it's just meant to be internal it'd be nice to keep all the hooks interfaces standardized

src/components/Employee/Profile/shared/useHomeAddressForm/useHomeAddressForm.tsx
```diff
@@ -79,61 +85,89 @@ export interface UseHomeAddressFormReady extends BaseFormHookReady<
...
+      street2: fetchedHomeAddress?.street2 ?? partnerDefaults?.street2 ?? '',
+      city: fetchedHomeAddress?.city ?? partnerDefaults?.city ?? '',
+      state: fetchedHomeAddress?.state ?? partnerDefaults?.state ?? '',
+      zip: fetchedHomeAddress?.zip ?? partnerDefaults?.zip ?? '',
+      courtesyWithholding:
+        fetchedHomeAddress?.courtesyWithholding ?? partnerDefaults?.courtesyWithholding ?? false,
+      effectiveDate:
+        fetchedHomeAddress?.effectiveDate?.toString() ?? partnerDefaults?.effectiveDate ?? '',
+    }
+  }, [isCreateMode, fetchedHomeAddress, partnerDefaults])
```
serikjensen: Nit: feels like this change isn't needed yeah? we could just change the other to use fetchedHomeAddress instead of currentHomeAddress?

src/components/Employee/Profile/shared/useHomeAddressForm/useHomeAddressForm.tsx
```diff
@@ -79,61 +85,89 @@ export interface UseHomeAddressFormReady extends BaseFormHookReady<
...
 
   const formMethods = useForm<HomeAddressFormData, unknown, HomeAddressFormOutputs>({
     resolver: zodResolver(schema),
     mode: validationMode,
     shouldFocusError,
     defaultValues: resolvedDefaults,
     values: resolvedDefaults,
-    resetOptions: { keepDirtyValues: true },
+    resetOptions: { keepDirtyValues: false },
```
serikjensen: To my knowledge we need to keep this as true to make sure the form state for default values is correctly set up. This is because the form initializes before we have the default values sometimes

src/components/Employee/Profile/shared/useHomeAddressForm/useHomeAddressForm.tsx
```diff
@@ -79,61 +85,89 @@ export interface UseHomeAddressFormReady extends BaseFormHookReady<
...
+    resolvedDefaults.street2,
+    resolvedDefaults.city,
+    resolvedDefaults.state,
+    resolvedDefaults.zip,
+    resolvedDefaults.courtesyWithholding,
+    resolvedDefaults.effectiveDate,
+    formMethods,
+  ])
+  /* eslint-enable react-hooks/exhaustive-deps */
+
```
serikjensen: You should be able to get rid of this by having keepDirtyValues set to true above

src/components/Employee/HomeAddress/management/getPendingFutureHomeAddress.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    expect(getPendingFutureHomeAddress(undefined)).toBeUndefined()
+    expect(getPendingFutureHomeAddress([])).toBeUndefined()
+  })
+
+  it('returns the earliest future inactive address', () => {
+    const now = new Date('2026-03-15T12:00:00')
+    const past = base({
+      uuid: '1',
+      active: false,
+      effectiveDate: '2026-01-01' as unknown as EmployeeAddress['effectiveDate'],
```
serikjensen: Nit: can we ask claude to fix the typing here?

serikjensen (approved):

PR #1579, opened 2026-04-17
feat(SDK-640): add employee work address management

dmortal:
## Summary

Introduces **Employee.WorkAddress** management: view current work address and history, add a future address, edit current or historical rows (with effective date when editing inactive addresses), and delete inactive history rows. Company locations and work addresses load from existing APIs; `companyId` is derived from the employee record. Submit and delete paths use `baseSubmitHandler` / `useBaseSubmit` for consistent error handling.

## Changes

- New management UI (`WorkAddress` / `WorkAddressView`) with modals, DataView history, and pending-future-address messaging.
- Extended **`useWorkAddressForm`** (submission modes, `updateTargetUuid`, `formSessionId` for reliable form reset when reopening modals) and shared hook field portal support for modal dropdowns/date picker.
- i18n, component exports, sdk-app **`Employee.WorkAddress`** registry entry, and work-address-related component events.

## Demo

_Delete this section or add screenshots of the work address screen, edit/change modals, and history actions._

## Related

- Jira: [SDK-640](https://gustohq.atlassian.net/browse/SDK-640)

## Testing

- `npm run tsc` (pass)
- Manual: open **Change address** and **Edit** from current and from history; change rows; cancel and reopen; confirm fields reset and match the selected address; submit create/update; delete an inactive row; verify portal dropdowns/date fields render inside the modal.

[SDK-640]: https://gustohq.atlassian.net/browse/SDK-640?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


serikjensen (commented):
I think there's a lot of complexity introduced here due to needing to differentiate between the case where we are automatically getting the active work address (profile) and the management case where we may be updating different work addresses

I added a proposal here where we might simplify this. if we update useWorkAddressForm to take a `workAddressId` prop, we can have the same create/update modes and default props automatically and we can keep the hook api consistent with the existing api

We could then create a dedicated active work address form hook that could be a convenience helper for the onboarding case.

I attached the plan for review here, lmk if that seems reasonable or if it's an oversimplification/there's more to be considered!

[simplify-work-address-hook-plan.md](https://github.com/user-attachments/files/26902096/simplify-work-address-hook-plan.md)

serikjensen (approved):

PR #1598, opened 2026-04-23
fix: remove default garnishment selection in deductions form

lbgustie:
## Summary

Remove the default "Garnishment" pre-selection when adding a new employee deduction. The radio group now starts empty, and sub-forms only appear after the user explicitly selects a deduction type.

## Changes

- Deduction type radio group no longer pre-selects "Garnishment" in add mode
- Sub-forms (ChildSupport, Garnishment, CustomDeduction) and the `<hr>` separator are hidden until a selection is made
- Edit mode behavior is unchanged — it still infers the type from the existing deduction

## Testing

- `npm run test -- --run src/components/Employee/Deductions/DeductionsForm/DeductionsForm.test.tsx`
- New test verifies no form renders until a deduction type is selected
- Existing tests updated to explicitly select a radio option before asserting on form content

### DIFF
--- a/src/components/Employee/Deductions/DeductionsForm/DeductionsForm.test.tsx
+++ b/src/components/Employee/Deductions/DeductionsForm/DeductionsForm.test.tsx
@@ -107,6 +107,18 @@ describe('DeductionsForm', () => {
       })
     })
 
+    it('does not show a form until a deduction type is selected', async () => {
+      renderDeductionsForm()
+
+      await waitFor(() => {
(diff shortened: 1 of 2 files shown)


serikjensen (changes requested):
I think there's enough going on in this one that we really need to make sure we understand what's going on. onsite day is happening but will review by tomorow in the AM. Putting a hold on this until then

serikjensen (approved):
Looks good! apologies for the delay on this one!

PR #1599, opened 2026-04-23
fix: remove italic styling and normalize text size in company address list

lbgustie:
## Summary

- Remove the HTML `<address>` tag in the company locations table which was applying browser-default italic styling
- Normalize both address lines to the same text size (city/state/zip was previously smaller)

## Changes

- Replaced `<address>` wrapper with a CSS module flex column layout (`.addressCell` in `List.module.scss`)
- Removed `size="sm"` from the city/state/zip text so both lines render at the same size
- Uses `toRem(4)` for gap spacing to match codebase convention

## Testing

- View in SDK dev app at `/company/Locations`
- Verify address text is not italicized and both lines are the same font size

### DIFF
--- a/src/components/Company/Locations/LocationsList/List.tsx
+++ b/src/components/Company/Locations/LocationsList/List.tsx
@@ -1,5 +1,6 @@
 import { useTranslation } from 'react-i18next'
 import { useLocationsList } from './useLocationsList'
+import styles from './List.module.scss'
 import PencilSvg from '@/assets/icons/pencil.svg?react'
 import { DataView, EmptyData, useDataView, VisuallyHidden } from '@/components/Common'
 import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentContext'
@@ -33,7 +34,7 @@ export const List = () => {
         title: t('locationListCol1'),
         render: location => {
           return (
-            <div style={{ display: 'flex', flexDirection: 'column', gap: 4 }}>
+            <div className={styles.addressCell}>
               <Components.Text as="div">{getStreet(location)}</Components.Text>
               <Components.Text as="div">{getCityStateZip(location)}</Components.Text>
             </div>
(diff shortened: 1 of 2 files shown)


src/components/Company/Locations/LocationsList/List.tsx
```diff
@@ -33,14 +33,10 @@ export const List = () => {
...
           return (
-            <>
-              <address>
-                <Components.Text as="div">{getStreet(location)}</Components.Text>
-                <Components.Text as="div" size="sm">
-                  {getCityStateZip(location)}
-                </Components.Text>
-              </address>
-            </>
+            <div style={{ display: 'flex', flexDirection: 'column', gap: 4 }}>
```
jeffredodd: Can you ask AI to use our `<Flex>` component here instead of a div?

serikjensen: +1

src/components/Company/Locations/LocationsList/List.tsx
```diff
@@ -33,14 +34,10 @@ export const List = () => {
...
           return (
-            <>
-              <address>
-                <Components.Text as="div">{getStreet(location)}</Components.Text>
-                <Components.Text as="div" size="sm">
-                  {getCityStateZip(location)}
-                </Components.Text>
-              </address>
-            </>
+            <div className={styles.addressCell}>
```
serikjensen: the `address` element here is useful for accessibility. You should be able to apply these flex styles to the address element and provide any other style overrides you need in order to override the browser defaults without having to not use the html tag

serikjensen (commented):
One more comment here!

src/components/Company/Locations/LocationsList/List.tsx
jeffredodd: Fixed in b0ff308

jeffredodd (approved):

PR #1613, opened 2026-04-27
feat(2/5): add SelectEmployees presentational component (SDK-563)

jeffredodd:
## Summary

- Add `SelectEmployeesPresentation` — shared stateless UI consumed by both the time-off and holiday container components
- Renders employee table with search, multi-select, optional starting-balance column, optional reassignment warning, and Back/Continue actions
- Add translations (`Company.TimeOff.SelectEmployees`)
- Storybook stories: default, partial/full selection, search filtered, empty results, with/without reassignment warning, pagination

Depends on: #1611

## Test plan

- [ ] `npm run test -- --run src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesPresentation`
- [ ] `npm run storybook` → `Domain/TimeOff/SelectEmployees` — all 7 stories render correctly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1614, opened 2026-04-27
feat(3/5): add SelectEmployees container for time-off/sick policies (SDK-565)

jeffredodd:
## Summary

- Add `SelectEmployeesTimeOff` container component for adding employees to time-off/sick policies
- Fetches active employees via `useEmployeesListSuspense`, manages selection/search/pagination state
- **Standalone mode**: calls `useTimeOffPoliciesAddEmployeesMutation` with selected UUIDs and optional per-employee starting balances
- **Wizard mode**: emits `TIME_OFF_ADD_EMPLOYEES_DONE` with selected UUIDs (no API call)
- Shows reassignment warning (employees can only be in one time-off policy)

Depends on: #1613

## Test plan

- [ ] `npm run test -- --run src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff`

### DIFF
--- a/src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
+++ b/src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
@@ -1,0 +1,131 @@
+import { useCallback, useMemo, useState } from 'react'
+import { useEmployeesListSuspense } from '@gusto/embedded-api/react-query/employeesList'
+import { useTimeOffPoliciesAddEmployeesMutation } from '@gusto/embedded-api/react-query/timeOffPoliciesAddEmployees'
+import { SelectEmployeesPresentation } from './SelectEmployeesPresentation'
+import type { EmployeeItem } from './SelectEmployeesPresentationTypes'
+import { useBase } from '@/components/Base/useBase'
+import { usePagination } from '@/hooks/usePagination/usePagination'
+import { componentEvents } from '@/shared/constants'
+
+interface SelectEmployeesTimeOffProps {
+  companyId: string
+  policyId: string
+  mode?: 'standalone' | 'wizard'
+}
+
+export function SelectEmployeesTimeOff({
+  companyId,
+  policyId,
+  mode = 'standalone',
+}: SelectEmployeesTimeOffProps) {
+  const { onEvent, baseSubmitHandler } = useBase()
+  const [selectedUuids, setSelectedUuids] = useState(new Set<string>())
+  const [searchValue, setSearchValue] = useState('')
+  const [balances, setBalances] = useState<Record<string, string>>({})
+  const { currentPage, itemsPerPage, getPaginationProps, resetPage } = usePagination()
+
+  const { data: employeesData, isFetching } = useEmployeesListSuspense({
+    companyId,
+    terminated: false,
+    page: currentPage,
+    per: itemsPerPage,
+  })
+
+  const employees = useMemo<EmployeeItem[]>(
+    () =>
+      (employeesData.showEmployees ?? []).map(e => ({
+        uuid: e.uuid,
+        firstName: e.firstName,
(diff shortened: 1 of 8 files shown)


src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
```diff
@@ -0,0 +1,131 @@
...
+    return employees.filter(e =>
+      `${e.firstName ?? ''} ${e.lastName ?? ''}`.toLowerCase().includes(lower),
+    )
+  }, [employees, searchValue])
+
+  const pagination = useMemo(
+    () => getPaginationProps(employeesData.httpMeta.response.headers, isFetching),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    [employeesData.httpMeta.response.headers, isFetching, currentPage, itemsPerPage],
+  )
```
serikjensen: Didn't we get a util to centralize the pagination state?

src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
```diff
@@ -0,0 +1,131 @@
...
+          timeOffPolicyUuid: policyId,
+          requestBody: {
+            employees: [...selectedUuids].map(uuid => ({
+              uuid,
+              ...(balances[uuid] !== undefined && { balance: balances[uuid] }),
+            })),
+          },
+        },
+      })
+      onEvent(componentEvents.TIME_OFF_ADD_EMPLOYEES_DONE)
```
serikjensen: Could we add the api response from add employees here?

src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
```diff
@@ -0,0 +1,131 @@
...
+    })
+  }, [mode, baseSubmitHandler, addEmployees, policyId, selectedUuids, onEvent])
```
serikjensen: Nit: seems like a lot of deps for a memoized function, i'm assuming this will likely just re-render as needed still due to referential equality checks

serikjensen (approved):

src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
jeffredodd: Good call — `usePagination` is the centralizing util here, and the additional `useMemo` wrapper (with the eslint-disable) was unnecessary overhead. Switched to a direct `getPaginationProps(...)` call inline, matching the pattern in `PayrollList` and `useContractors`.

src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
jeffredodd: Done — captured the return value and forwarded `response.timeOffPolicy` so partners get the updated policy object in the event payload without a separate refetch.

src/components/UNSTABLE_TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
jeffredodd: Agreed — `selectedUuids` is a new `Set` reference on every toggle, so the memoization was essentially a no-op (and `balances` was also missing from the deps, making it subtly buggy). Removed `useCallback` from `handleContinue` entirely — it recreates on every render either way, so no point pretending otherwise.

PR #1618, opened 2026-04-27
fix(SDK-464): validate integer percentages in payment method split

mariechatfield:
## Summary

Improve client-side validation of percentage splits for employee payment method.

## Changes

**Before:** Splitting an employee payment method by percentage allowed you to input decimal values; the server replied with a 422 and requested integers.

Screenshot 2026-04-27 at 12 51 49 PM

Error message at top of form, no inline validation:

> There was a problem with your submission.
> - If the payment amount is split by Percentage, split amount must be an integer.

---

**Now:** Client-side validation prevents you from typing or pasting a decimal in the percentage input. If you _do_ manage to use an invalid number (possibly via the hook?) we give a more descriptive error message.

Screenshot 2026-04-27 at 1 56 27 PM

Error message inline, below individual field marked as invalid:
> Percentage should be a whole number between 0 and 100

## Related

- Jira ticket: [SDK-464](https://gustohq.atlassian.net/browse/SDK-464)

## Testing

- `npm run test PaymentMethod`
- `npm run sdk-app`
  - http://localhost:5200/employee/PaymentMethod
  - Add a second bank account
  - Split paycheck by percentage
  - Attempt to type or paste a decimal; should round to nearest integer
  - Attempt to submit with an invalid integer (e.g. 101); should show more descriptive error message

[SDK-464]: https://gustohq.atlassian.net/browse/SDK-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Employee/PaymentMethod/PaymentMethod.test.tsx
+++ b/src/components/Employee/PaymentMethod/PaymentMethod.test.tsx
@@ -116,6 +116,46 @@ describe('PaymentMethod - Percentage Split Validation', () => {
       expect(result.success).toBe(false)
     })
 
+    it('should reject decimal percentages', () => {
+      const invalidData = {
+        type: 'Direct Deposit' as const,
+        isSplit: true as const,
+        hasBankPayload: false as const,
+        splitBy: 'Percentage' as const,
+        splitAmount: {
+          'account-1': 80.001,
+          'account-2': 19.9,
+        },
+        priority: {
+          'account-1': 1,
+          'account-2': 2,
+        },
+      }
+
+      const result = CombinedSchema.safeParse(invalidData)
+      expect(result.success).toBe(false)
+      expect(result.error?.issues[0]?.code).toEqual('invalid_union')
+
+      if (!result.success && result.error.issues[0]?.code === 'invalid_union') {
+        expect(result.error.issues[0].errors).toContainEqual([
+          {
+            code: 'invalid_type',
+            expected: 'int',
+            format: 'safeint',
+            message: 'Invalid input: expected int, received number',
+            path: ['splitAmount', 'account-1'],
+          },
+          {
+            code: 'invalid_type',
+            expected: 'int',
+            format: 'safeint',
+            message: 'Invalid input: expected int, received number',
+            path: ['splitAmount', 'account-2'],
+          },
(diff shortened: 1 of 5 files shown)


src/components/Employee/PaymentMethod/PaymentMethod.test.tsx
```diff
@@ -116,6 +116,46 @@ describe('PaymentMethod - Percentage Split Validation', () => {
...
+            path: ['splitAmount', 'account-1'],
+          },
+          {
+            code: 'invalid_type',
+            expected: 'int',
+            format: 'safeint',
+            message: 'Invalid input: expected int, received number',
+            path: ['splitAmount', 'account-2'],
+          },
+        ])
```
mariechatfield, line 155: This is a bit of a different format from the other unit tests, but I wanted to make sure we were catching the actual `int` validation error.

Because this is part of a discriminated union, we get _a bunch_ of error codes about the entire shape not matching the union but it is broken down into individual error codes. This was my best attempt at capturing a snapshot of the relevant zod schema issue if we pass a decimal

src/i18n/en/Employee.PaymentMethod.json
```diff
@@ -45,6 +45,7 @@
   "validations": {
     "percentageError": "If payment method amount is split by Percentage, all split amounts must add up to exactly 100.",
     "percentageErrorWithTotal": "Splits must total 100%. Currently {{total}}%.",
+    "percentageAmountError": "Percentage should be a whole number between 0 and 100",
```
mariechatfield, line 48: Is this the correct voice and structure? I tried to follow the example of routing and account number `[field name] should be a [brief description]` without a trailing period

serikjensen: Let's check in with @aaronlee777 on copy here

serikjensen (commented):
Looks good! recommended we check in with @aaronlee777 on the copy and then i'll go ahead and approve. great work!

serikjensen (approved):
Looks like aaron is good to go on copy!

PR #1619, opened 2026-04-27
fix(SDK-169): respect set-only contract for mailing/filing address flags

hukid:
## Summary

* Disables the mailing/filing-address checkbox in the location edit form when the location is already the company's mailing/filing address, with helper copy explaining how to transfer the designation.
* Omits the corresponding field from the `PUT /v1/locations/:id` request body when its checkbox is locked, matching the gws-flows reference UI and the underlying API contract.
* Updates the locations MSW mock to mirror the API's set-only behavior so the test suite guards against regressions.

## Why

Investigating SDK-169 surfaced that the underlying API treats `mailing_address` and `filing_address` as **set-only** flags: a company must always have exactly one of each, so a request to flip a `true` value back to `false` is silently ignored.

| Current value | Sent value | API behavior |
|---|---|---|
| `false` | `true` | Sets this location as mailing/filing (overrides any other) |
| `false` | `false` or omitted | No-op |
| `true` | `true` or omitted | No-op |
| **`true`** | **`false`** | **Silently ignored — value stays `true`** |

The previous SDK form let users uncheck either option and faithfully sent `false` on the wire, where the API dropped the change. From a user's perspective the selection wasn't persisting — the symptom reported in SDK-169.

The reference [gws-flows location form](https://github.com/Gusto/gws-flows/blob/main/app/controllers/company/locations_controller.rb) handles this by **disabling** the already-set checkboxes and **omitting** them from the PUT body. This PR mirrors that behavior in the SDK.

See the investigation comment on [SDK-169](https://gustohq.atlassian.net/browse/SDK-169) for the full breakdown. This was previously acknowledged in #249 (CheckboxGroup migration) and deferred at that time.

## What changed

* `src/components/Company/Locations/LocationForm/Form.tsx` — sets `isDisabled` and a context-aware description on the mailing/filing options based on the current location's flags (read from the existing `LocationsFormProvider` context).
* `src/components/Company/Locations/LocationForm/LocationForm.tsx` — computes `isMailingLocked` / `isFilingLocked` from the loaded location, passes them into `LocationsFormProvider`, and conditionally omits those fields from the PUT request body.
* `src/components/Company/Locations/LocationForm/useLocationForm.ts` — extends the form context type with the two new locked flags.
* `src/i18n/en/Company.Locations.json` (+ regenerated `src/types/i18next.d.ts`) — adds `mailingAddressDescriptionLocked` / `filingAddressDescriptionLocked` strings.
* `src/test/mocks/apis/company_locations.ts` — adds a `getLocation` handler for `GET /v1/locations/:id` (used in the new edit-mode tests) and updates the PUT handler to mirror the API's set-only contract.
* `src/components/Company/Locations/LocationForm/LocationForm.test.tsx` — adds an `edit mode` describe block with three new tests.

## Test plan

Automated:

- [x] New unit tests cover (1) prepopulation from the loaded location, (2) the disabled-when-locked checkbox state, and (3) the locked field being omitted from the PUT request body.
- [x] `npm run test -- --run` passes (2,273 / 2,273).
- [x] `npm run tsc` clean.
- [x] ESLint clean on changed files.

Manual:

- [ ] In the SDK dev app, open `Company.LocationsFlow`, ensure the company has at least two addresses, and edit the address that is currently the mailing address. Verify the *Mailing address* checkbox is checked, disabled, and shows the locked-state helper copy. The *Filing address* checkbox should remain interactive.
- [ ] Edit a different (non-mailing) address and check the *Mailing address* box, then save. Confirm the mailing-address designation transfers to the new location (the pill in the list view moves accordingly).
- [ ] Edit any address without changing the mailing/filing options and save — request should succeed without sending no-op flags.

## Notes / out of scope

* The existing locations POST/PUT MSW mocks read camelCase keys from `request.json()`, but Speakeasy serializes the wire format as snake_case, so several echoed fields have always been `undefined` in test responses. Not affecting any current test, but worth a follow-up cleanup. (Not addressed here to keep this PR scoped.)
* The full mailing/filing transfer flow (checking the box on a different location to move the designation) already works via the API — the SDK's create flow is unchanged.

## Screenshot

When address types are set for the address:
image

When address types are not set for the address:
image

[SDK-169]: https://gustohq.atlassian.net/browse/SDK-169?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Company/Locations/LocationForm/LocationForm.tsx
```diff
@@ -39,9 +40,12 @@ function Root({
   useI18n('Company.Locations')
   const { onEvent, baseSubmitHandler } = useBase()
 
+  const queryClient = useQueryClient()
   const { mutateAsync: createLocation, isPending: isPendingCreate } = useLocationsCreateMutation()
   const { mutateAsync: updateLocation, isPending: isPendingUpdate } = useLocationsUpdateMutation()
   const addressType = ['mailingAddress', 'filingAddress'] as const
+  const isMailingLocked = location?.mailingAddress === true
+  const isFilingLocked = location?.filingAddress === true
```
serikjensen, line 48: Nit: i wonder if just hiding the checkboxes in this case might be preferable

src/components/Company/Locations/LocationForm/LocationForm.tsx
```diff
@@ -63,19 +67,37 @@ function Root({
...
+      // of *inactive* queries too (refetchType: 'all') because the next form
+      // mount initializes useForm defaults synchronously from cache; otherwise
+      // useSuspenseQuery returns the stale cached value without re-suspending
+      // and the form seeds with a stale addressType on first open.
+      const refreshLocationsCache = () =>
+        queryClient.invalidateQueries({
+          queryKey: ['@gusto/embedded-api', 'Locations'],
+          refetchType: 'all',
+        })
+
```
serikjensen, line 92: You might verify if this works without the manual invalidation. We have it configured to invalidate queries after any mutation so this is potentially not needed

serikjensen (approved):

src/components/Company/Locations/LocationForm/LocationForm.tsx
hukid: I'll let @aaronlee777 to figure out a better pattern, keep this behavior for consistency with Flow.

src/components/Company/Locations/LocationForm/LocationForm.tsx
hukid: It doesn't work without manual invalidation, so I added this. I believe the invalidated query is scoped to the location which is changed however we need invalidate another object at the same time in this case.

serikjensen: Right, it should be invalidating all queries. We may want to look into this further

PR #1625, opened 2026-04-28
fix(SDK-426): add empty state to RecoveryCasesList

mariechatfield:
## Summary

Make the empty state for recovery cases more visually consistent with information requests.

## Changes

**Before:** `RecoveryCasesList` showed an empty table when there were no active cases
Screenshot 2026-04-28 at 8 12 36 AM

**Now:** `RecoveryCasesList` shows the same empty state as the `InformationRequestsList` and hides the description since it's not relevant
Screenshot 2026-04-28 at 8 16 21 AM

## Related

- Jira ticket: [SDK-426](https://gustohq.atlassian.net/browse/SDK-426)

## Testing

- `npm run test -- RecoveryCasesList`
- `npm run sdk-app`
  - http://localhost:5200/payroll/RecoveryCases compared to http://localhost:5200/informationrequests/InformationRequestList

[SDK-426]: https://gustohq.atlassian.net/browse/SDK-426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Payroll/RecoveryCases/RecoveryCasesList/RecoveryCasesList.tsx
+++ b/src/components/Payroll/RecoveryCases/RecoveryCasesList/RecoveryCasesList.tsx
@@ -11,7 +11,7 @@ import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentCon
 import { useComponentDictionary, useI18n } from '@/i18n'
 import { DataView } from '@/components/Common/DataView/DataView'
 import { useDataView } from '@/components/Common/DataView/useDataView'
-import { Flex, FlexItem } from '@/components/Common'
+import { EmptyData, Flex, FlexItem } from '@/components/Common'
 import { recoveryCasesEvents } from '@/shared/constants'
 import type { BadgeProps } from '@/components/Common/UI/Badge/BadgeTypes'
 import { formatNumberAsCurrency } from '@/helpers/formattedStrings'
@@ -116,6 +116,9 @@ function Root({ companyId, dictionary, onEvent }: RecoveryCasesListProps) {
 
   const dataViewProps = useDataView({
     data: recoveryCases,
+    emptyState: () => (
+      <EmptyData title={t('emptyTableTitle')} description={t('emptyTableDescription')} />
+    ),
     columns: [
       {
         key: 'originalDebitDate',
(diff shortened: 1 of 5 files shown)


src/i18n/en/Payroll.RecoveryCasesList.json
```diff
@@ -1,6 +1,8 @@
 {
   "title": "Recovery cases",
   "description": "One or more payments couldn't be processed due to a bank error. Resolve open recovery cases to continue running payroll.",
+  "emptyTableTitle": "No recovery cases",
+  "emptyTableDescription": "There are no recovery cases that need your response at this time.",
```
mariechatfield, line 5: @aaronlee777 is this the right copy? I used the same verbiage as information requests but I'm not sure "need your response" is the right phrasing for recovery cases

src/components/Payroll/RecoveryCases/RecoveryCasesList/RecoveryCasesList.tsx
```diff
@@ -169,7 +172,7 @@ function Root({ companyId, dictionary, onEvent }: RecoveryCasesListProps) {
           <Heading as="h2" styledAs="h4">
             {t('title')}
           </Heading>
-          <Text>{t('description')}</Text>
+          {recoveryCases.length > 0 && <Text>{t('description')}</Text>}
```
jeffredodd, line 175: Super minor, i'm wondering if there is a chance the API will ever return recoveryCases as null. In that case it might be better to do `recoveryCases && recoveryCases.length > 0` but i'm guessing here I could be wrong and this is never null.

jeffredodd (approved):
LGTM just a minor question

src/components/Payroll/RecoveryCases/RecoveryCasesList/RecoveryCasesList.tsx
mariechatfield:
```tsx
const recoveryCases = (data.recoveryCaseList ?? []).filter(rc => rc.status !== 'recovered')
```

This is where `recoveryCases` is defined above, so I think it should already handle nulls from the server.

jeffredodd: Oh sweet yeah I should have looked up! Thanks for the response

PR #1627, opened 2026-04-28
fix: add isWithinBox support to DataCards

aaronlee777:
## Summary
- Adds `isWithinBox` prop to `DataCards`, matching the existing pattern in `DataTable`
- When inside a Box, cards strip their border, border-radius, and box-shadow, and render divider lines between items instead
- Adds `overflow: hidden` to Box so card content respects the Box border-radius

## Demo

**Before**
Screenshot 2026-04-28 at 4 29 00 PM

**After**

Screenshot 2026-04-28 at 4 42 33 PM

### DIFF
--- a/src/components/Common/DataView/DataCards/DataCards.tsx
+++ b/src/components/Common/DataView/DataCards/DataCards.tsx
@@ -2,6 +2,7 @@ import { useId } from 'react'
 import { useTranslation } from 'react-i18next'
 import styles from './DataCards.module.scss'
 import type { useDataViewPropReturn, SelectionMode } from '@/components/Common/DataView/useDataView'
+import type { TableProps } from '@/components/Common/UI/Table/TableTypes'
 import { useSelectionState } from '@/components/Common/DataView/useSelectionState'
 import { Flex } from '@/components/Common/Flex/Flex'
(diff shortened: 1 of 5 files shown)


src/components/Common/DataView/DataCards/DataCards.tsx
```diff
@@ -68,7 +71,11 @@ export const DataCards = <T,>({
   }
 
   return (
-    <div data-testid="data-cards">
+    <div
+      className={styles.root}
+      data-testid="data-cards"
+      data-within-box={isWithinBox || undefined}
```
serikjensen: nit i think my preference would be doing this with conditionally applying css classes instead of adding a lot of data attributes

serikjensen (approved):

src/components/Common/DataView/DataCards/DataCards.tsx
jeffredodd: Yeah I think I can +1 to this, doing this in our styles is going to be more clear for partners hopefully.

aaronlee777: Updated!

serikjensen (approved):

PR #1647, opened 2026-04-30
fix(SDK-735): improve validation error for net pay in off-cycle payroll

mariechatfield:
## Summary

Improve validation error when setting net amount for an employee in an off-cycle payroll.

## Changes

- Update the validation error message for net pay to use an i18n string instead of zod schema defaults
- ~Update styling on the Calculate button so it stays aligned with the net pay input whether or not an error message is visible~

## Demo

**Before:** error message reads "Too small: expected number to be >0" and the "Calculate" button is misaligned

Screenshot 2026-04-30 at 12 11 28 PM

**After:** error message reads "Net amount must be greater than zero cents" ~and the "Calculate" button is correctly aligned with the input with/without the error message~

Screenshot 2026-04-30 at 12 43 38 PM

## Related

- Jira ticket: [SDK-735](https://gustohq.atlassian.net/browse/SDK-735)

## Testing

- `npm run sdk-app`
- http://localhost:5200/payroll/PayrollFlow
- Run an off-cycle payroll (e.g. for today's date, paid out next week, bonus cycle, for a single employee)
- For an employee, use the triple dot menu to select Set net earnings
- Attempt to calculate with $0.00 net earnings and observe the error

[SDK-735]: https://gustohq.atlassian.net/browse/SDK-735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Payroll/GrossUpModal/GrossUpModal.module.scss
```diff
@@ -17,11 +17,18 @@
   }
 }
 
-.calculateButton:global(.react-aria-Button) {
-  min-height: toRem(40);
-  padding-block: toRem(7);
+// Needs three classes to win specifity for margin and height compared to
+// `.root:global(.react-aria-Button)` in Button.module.scss
+// Extra class may be removed if we update loading order to include common UI classes first
+.content .calculateButton:global(.react-aria-Button) {
```
mariechatfield: I'm not entirely confident about these changes but I tried to leave (too many?) comments explaining each change.

Is this a dev thing or in production too? I would have expected `Button.module.scss` to be loaded before `GrossUpModal.modules.scss`. Since it's not, the order implies that `Button` should win over rules in `GrossUpModal` with the same specificity (`0, 2, 0`). If I need to increase specificity to win against another rule, I usually leave a comment explaining why and add an extra class since class-based specificity is the easiest to compare imo.

Image

serikjensen (commented):
hrmmm i'm wondering if we can find a cleaner way to navigate this alignment. let's chat about it tomorrow and maybe bring aaron in

src/components/Payroll/GrossUpModal/GrossUpModal.module.scss
```diff
@@ -17,11 +17,18 @@
...
   flex-shrink: 0;
-  align-self: flex-end;
+  // Aligned from flex-start since height of input changes with error message.
+  // Adjust own margin top to account for height of input label which is static
+  margin-top: toRem(24);
+  // Match the exact height of the input
+  min-height: toRem(40);
+  height: toRem(40);
+  max-height: toRem(40);
```
mariechatfield: @serikjensen would love to chat more about the right way to fix this alignment. I'm guessing the main issue is that this is hard-coding assumptions about the input height based on the common UI input and with the component adapter or custom theming (e.g. input label font size) that could easily break

My initial thoughts are that.... 
- if we expect some kind of action button next to an input to be a common UX pattern we may need to add some kind of new prop to the component API
- but that would be a breaking change and it adds an extra non-standard requirement to anyone using their own components
- especially if it's something like material where the library may be opinionated about layout / may want the input itself to take full width especially on smaller screens

That may mean that the longer term solution is to update _this specific_ instance and move the calculate button below the input? Or make some other adjustment so that we're not relying on center aligning a button to an input that may change height (whether it's our own input or a custom one via the adapter).

I can also just remove this CSS entirely, since it was an opportunistic fix but it may be complex enough to merit its own ticket.

src/components/Payroll/GrossUpModal/GrossUpModal.tsx
```diff
@@ -122,6 +122,7 @@ export function GrossUpModal({ isOpen, onCalculateGrossUp, onApply, onCancel }:
               name="netPay"
               label={t('netPayLabel')}
               format="currency"
+              errorMessage={t('validations.netPay')}
```
jeffredodd, line 125: My only real concern here is what we do if we get another error from the GrossUp endpoint. I'm not sure what error we would get aside from it maybe being down. Just a thought.

mariechatfield: It looks like most of the server errors are displayed at the top of the modal in a little alert, and this is primarily only client-side validation. Before it was auto-generated from the zod schema but (as far as I can tell) there's not a good way to give the schema translation overrides via the schema.

jeffredodd: Yeah thats fair, i'm blanking on how, but I do know when we get field errors back we do try to map them onto a field and they get bubbled up via context. Mostly cause we want to be accurate and a11y by mapping errors to fields specifically. But i've not worked with this in a while so I feel rusty even writing this :)

mariechatfield: @jeffredodd you're correct, this _will_ override anything we get from the server. I took a closer look at the code with claude.

I can override the message generated for this field like `.positive({ error: 'custom message' })` _but_ I don't have access to the translation context where the `GrossUpFormSchema` is currently defined (file-level).

Claude suggests I could move the schema definition inside the component like so:

```tsx
export function GrossUpModal() {                                                                                                                                                                     
  const { t } = useTranslation('Payroll.GrossUpModal')                                                                                                                                                  
                                                                                                                                                                                                          
  const schema = useMemo(                                                                               
    () => z.object({ netPay: z.number().positive({ message: t('validations.netPay') }) }),                                                                                                            
    [t]                                                                                   
  )                                                                                                                                                                                                     
     
  const formHandlers = useForm<GrossUpFormValues>({ resolver: zodResolver(schema), ... })                                                                                                               
}
```

I don't _love_ this, since we don't have any other schemas defined inside a component like this. Do you know of any other way we could provide a translation key to the schema and then have the `useField` hook translate it on demand?

src/components/Payroll/GrossUpModal/GrossUpModal.module.scss
mariechatfield: cc @aaronlee777 these are the changes I made that we discussed earlier today. I'm going to remove this override from this PR and we can have a followup discussion about what we want to do next. (I think I'm in favor of updating the design here if this really is the only place we have a button next to an input)

src/components/Payroll/GrossUpModal/GrossUpModal.tsx
mariechatfield: Per our discussion in Slack: I believe `net_pay` can have validation errors from the server. This is a wider question for the entire codebase, around how we want to handle errors from the server compared to schema validation errors.

Right now, my current solution is following the patterns in other parts of the codebase so I suggest we merge this one as is, but create a ticket for the larger question of reconciling error messages from various contexts? cc @serikjensen

jeffredodd (approved):

PR #1656, opened 2026-05-01
feat: update Company header styles with consistent spacing and supporting text variant

aaronlee777:
## Summary

- Standardize section headers across Company components by wrapping heading/description pairs in `Flex` with `gap={4}` and applying `variant="supporting"` to description text
- Add `alignItems="stretch"` to parent layout containers for consistent alignment

## Changes

- Updated header patterns in AssignSignatory (Create + Invite), DocumentSigner, FederalTaxes, Industry, LocationsList, and PayScheduleList
- Wrapped heading + description pairs in `Flex flexDirection="column" gap={4}`
- Applied `variant="supporting"` to description `Text` components
- Added `alignItems="stretch"` to parent `Flex` containers

## Testing

- `npm run storybook` — verify header spacing and text styling in affected Company components
- `npm run test -- --run` — ensure no regressions

### DIFF
--- a/src/styles/_Base.scss
+++ b/src/styles/_Base.scss
@@ -19,8 +19,4 @@
     opacity: 1;
     margin: 0;
   }
-
-  section {
-    width: 100%;
-  }
 }
--- a/src/styles/_Base.scss
+++ b/src/styles/_Base.scss
@@ -19,4 +19,8 @@
     opacity: 1;
     margin: 0;
   }
+
+  section {
+    width: 100%;
+  }
 }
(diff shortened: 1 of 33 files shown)


src/styles/_Base.scss
```diff
@@ -19,4 +19,8 @@
     opacity: 1;
     margin: 0;
   }
+
+  section {
+    width: 100%;
+  }
```
serikjensen: This is a pretty dangerous rule to add globally. You can already see with IncludeDeductions.tsx that you had to modify the icon container to avoid regressions with this. 

My rec would be to solve this case by case with an applied className to avoid unexpected/frustrating behavior

serikjensen (approved):

PR #1670, opened 2026-05-04
feat!: migrate Employee.FederalTaxes to hook architecture and add steady-state edit mode

dmortal:
## Summary

Migrates the `Employee.FederalTaxes` component to the hook-based SDK architecture (new `useFederalTaxesForm` partner hook), wires it into the Employee Dashboard flow as a steady-state edit screen, and adds an `isOnboarding` prop so the same component can render either the onboarding CTA ("Continue" + auto-advance via `EMPLOYEE_FEDERAL_TAXES_DONE`) or a Cancel + Save layout with a success alert for in-place edits.

## Breaking changes

- **`EMPLOYEE_FEDERAL_TAXES_UPDATED` event payload.** The event now carries the updated `EmployeeFederalTax` entity (i.e. `result.data` from the hook) instead of the full Update Federal Taxes endpoint response. Consumers that previously read `payload.employeeFederalTax` should switch to using the payload directly. See `docs/workflows-overview/employee-onboarding/employee-onboarding.md` for the updated event contract.

## Changes

- **New `useFederalTaxesForm` hook** at `src/components/Employee/FederalTaxes/shared/useFederalTaxesForm/` (schema, fields, hook, barrel) following the migration skill; types / error codes / `FILING_STATUS_VALUES` re-exported from the SDK root barrel. The hook only requires `filingStatus` by default (matching the API); the bundled `<Employee.FederalTaxes>` component promotes the rest via `optionalFieldsToRequire` to preserve the original all-required UX.
- **`FederalTaxes` component refactored** to wrap `BaseBoundaries`, delegate to `Root` / `Ready`, use `SDKFormProvider` + `form.Fields`. Old `Head` / `Actions` / `FederalForm` / `useFederalTaxes` context removed. The Filing Status select uses `getOptionLabel` to render localized labels.
- **`isOnboarding` prop** (default `false`):
  - `true` (passed by `OnboardingFlow` / `SelfOnboardingFlow`): single Continue button, emits `EMPLOYEE_FEDERAL_TAXES_DONE` after save.
  - `false` (used by `Dashboard`): Cancel + Save buttons. Cancel emits `CANCEL`; Save submits, shows a dismissible "Successfully updated federal tax settings." alert, and keeps the user on the screen (no `_DONE`).
- **Dashboard flow integration**: `dashboardStateMachine` transitions `index → federalTaxes` on `EMPLOYEE_FEDERAL_TAXES_EDIT` and back on `CANCEL` / `EMPLOYEE_FEDERAL_TAXES_DONE`, mounting a new `FederalTaxesContextual`. Shared "return to index" reducer extracted.
- **i18n**: added `cancelCta`, `saveCta`, `successAlert` to `Employee.FederalTaxes.json`; types regenerated.
- **Docs**: new `docs/hooks/useFederalTaxesForm.md`, hook inventory row in `docs/hooks/hooks.md`, and `docs/workflows-overview/employee-onboarding/employee-onboarding.md` updated to document `isOnboarding` and the new event payload.
- **Tests**: new `FederalTaxes.test.tsx` covering rendering, prefill, onboarding-mode (Continue + \`_DONE\`), and steady-state-mode (Cancel/Save + success alert + no \`_DONE\`); mock-handler factories added in `src/test/mocks/apis/employee_federal_taxes.ts`.

## Testing

- \`npm run lint\` — clean
- \`npm run test -- --run\` — 190 files / 2,437 tests passing
- \`npm run test -- --run src/components/Employee/FederalTaxes\` — 10 tests passing (rendering + onboarding mode + steady-state mode)
- \`npm run build\` — clean

Manual:

1. Open the Employee Dashboard, switch to the **Taxes** tab, click **Edit** next to Federal taxes → form opens with Cancel + Save buttons.
2. Click **Cancel** → returns to dashboard without saving.
3. Edit a field, click **Save** → success alert appears at the top, user stays on the form; alert is dismissible via the × button.
4. Run admin and self-service onboarding flows → form still renders a single "Continue" CTA and advances on submit as before.


src/components/Employee/FederalTaxes/shared/useFederalTaxesForm/useFederalTaxesForm.tsx
```diff
@@ -0,0 +1,261 @@
...
+  FederalTaxesFormData,
+  FederalTaxesFields
+> {
+  data: {
+    employeeFederalTax: EmployeeFederalTax
+  }
+  status: { isPending: boolean; mode: 'update' }
+  actions: {
+    onSubmit: (
+      callbacks?: FederalTaxesSubmitCallbacks,
```
serikjensen: i think we should only provide the callbacks if there is more than one api operation that happens inside of the hook, ex. what happens with employee details. Otherwise onSubmit just takes no arguments and returns the api result.

I have added this to the claude instruction but for some reason it doesn't seem to like it and keeps trying to add the callbacks 😂

src/components/Employee/FederalTaxes/FederalTaxes.tsx
```diff
@@ -1,138 +1,208 @@
...
-    employeeUuid: employeeId,
-  })
-  const employeeFederalTax = fedData.employeeFederalTax!
-
-  const { mutateAsync: updateFederalTaxes, isPending } =
-    useEmployeeTaxSetupUpdateFederalTaxesMutation()
+  const federalTaxes = useFederalTaxesForm({
+    employeeId,
+    defaultValues,
+  } satisfies UseFederalTaxesFormProps)
```
serikjensen: For consistent behavior with the previous implementation, you may want to set `optionalFieldsToRequire` with the fields that are required

by default, the hook just requires the api required fields so that partners can have flexibility to require or leave as optional. But in our form i think we have them as required

docs/hooks/useFederalTaxesForm.md
```diff
@@ -0,0 +1,477 @@
...
+
+Updates an employee's federal tax (W-4) withholding information — filing status, multiple-jobs flag, dependents, other income, deductions, and extra withholding.
+
+```tsx
+import { useFederalTaxesForm, SDKFormProvider } from '@gusto/embedded-react-sdk'
+```
+
+The federal tax record is created automatically with the employee, so this hook is always in update mode. Only the revised 2020 W-4 format (`rev_2020_w4`) is supported for updates.
+
+> **`<FederalTaxes>` component note:** When using the `Employee.FederalTaxes` component, set `isOnboarding` to `true` to render a single "Continue" submit button that emits `EMPLOYEE_FEDERAL_TAXES_DONE` after a successful save (so the parent flow can advance). Leave it `false` (default) for steady-state edit screens — the component then renders Cancel + Save buttons, where Cancel emits `CANCEL` and Save submits the form, surfaces a success alert, and keeps the user on the screen.
```
serikjensen: wondering if this blurb is needed since this is the hook documentation?

serikjensen (commented):
couple of comments! looking good, only required change would be to update the optionalFieldsToRequire prop to include the field names so that the fields are required consistent with the previous experience

serikjensen (approved):

PR #1674, opened 2026-05-05
feat!: migrate Employee.Profile to hook architecture and add steady-state edit mode

dmortal:
## Summary

Migrates \`Employee.Profile\` to the hook architecture established by \`FederalTaxes\` and \`HomeAddress\` — relocating the existing onboarding variant into an \`onboarding/\` subdirectory and adding a new \`management/\` variant for steady-state editing within \`DashboardFlow\`.

## Changes

- **\`Profile/onboarding/\`** — existing \`AdminProfile\` and \`EmployeeProfile\` relocated here; imports updated to new paths
- **\`useEmployeeDetailsForm\`** — props refactored to a discriminated union: create mode requires \`{ companyId }\` (no \`employeeId\`), update mode requires \`{ employeeId }\` (\`companyId\` optional but unused); removes the need to pass an empty string for \`companyId\` in update callers
- **\`Profile/management/Profile.tsx\`** — new steady-state edit component using \`useEmployeeDetailsForm\` (update-only); shows first/middle/last name, email, SSN, DOB with Cancel/Save actions and an inline success alert
- **\`DashboardFlow\`** — \`employee/update\` event now transitions to a \`profile\` state; returns to \`index\` on \`CANCEL\` (success stays in-place, matching FederalTaxes pattern)
- **\`employeeOnboarding.ts\` / \`employeeManagement.ts\`** — explicit variant exports added; \`Profile/index.ts\` barrel removed; all internal consumers updated to direct paths

## Breaking changes

This PR is marked \`feat!\` for two reasons:

1. **\`useEmployeeDetailsForm\` props changed to a discriminated union.** Callers must now pass either \`{ companyId }\` (create mode) or \`{ employeeId }\` (update mode). The old shape where both were optional is no longer valid. SDK-internal callers have been updated; any partner using this hook directly must update their call sites.

2. **\`Profile\` barrel export removed.** \`Employee/Profile/index.ts\` is gone. Partners importing \`Profile\` from the old path must switch to the explicit variant:
   - Onboarding: \`@gusto/embedded-react-sdk/employeeOnboarding\` (or \`../Profile/onboarding/Profile\` internally)
   - Management: \`@gusto/embedded-react-sdk/employeeManagement\` (or \`../Profile/management/Profile\` internally)

## Testing

\`\`\`bash
# All Profile tests
npm run test -- --run src/components/Employee/Profile/onboarding/Profile.test.tsx

# Dashboard integration
npm run test -- --run src/components/Employee/Dashboard/

# Full build
npm run build
\`\`\`

All 49 Profile tests and 7 Dashboard tests pass. Build clean.

## Related

- Follows pattern established in: #1670 (FederalTaxes hook migration)


src/components/Employee/Profile/management/Profile.tsx
```diff
@@ -0,0 +1,144 @@
...
+}
+
+function ProfileRoot({ employeeId, className, dictionary, onEvent }: ProfileProps) {
+  useI18n('Employee.Profile')
+  useComponentDictionary('Employee.Profile', dictionary)
+  const { t } = useTranslation('Employee.Profile')
+  const Components = useComponentContext()
+
+  const employeeDetails = useEmployeeDetailsForm({
+    companyId: '',
```
serikjensen: any reason we're not supplying the company id here? if there's a valid case for not having the company id maybe we need to make company id optional?

dmortal: updating this - companyId is required but only in create mode which is not applicable to steady state

src/components/Employee/Profile/shared/useEmployeeDetailsForm/useEmployeeDetailsForm.tsx
```diff
@@ -46,16 +46,18 @@ export interface EmployeeDetailsSubmitCallbacks {
...
   optionalFieldsToRequire?: EmployeeDetailsOptionalFieldsToRequire
   defaultValues?: Partial<EmployeeDetailsFormData>
   validationMode?: UseFormProps['mode']
   shouldFocusError?: boolean
 }
 
+export type UseEmployeeDetailsFormProps =
+  | (UseEmployeeDetailsFormSharedProps & { companyId: string; employeeId?: never })
+  | (UseEmployeeDetailsFormSharedProps & { employeeId: string; companyId?: never })
+
```
serikjensen, line 60: i worry a little bit about this for the component usage 🤔 

ex. if you are in create mode, you provide the companyId, then you need to remove the company id once you have the employee id?

This is the case you get in employee profile in the partial update case. You set the company id on the hook and then optionally supply the employee id after the update is successful to make sure it gets placed into update mode. In that case, this would introduce the overhead of also needing to omit the company id when the employee id got supplied

Profile usage:

```
  const employeeDetails = useEmployeeDetailsForm({
    companyId,
    employeeId: resolvedEmployeeId,
    withSelfOnboardingField: true,
    optionalFieldsToRequire,
    defaultValues: employeeDetailsDefaults,
    shouldFocusError: false,
  })
```

I think if i was articulating the proper scenarios here

## Create mode
* Provide a company id
* Don't provide an employee id

## Update mode
* Provide an employee id
* Company id can be present but isn't used

So basically i think that company id should be required if no employee id is present. But if an employee id is present, company id can also be there.

dmortal: done

serikjensen (approved):

PR #1675, opened 2026-05-05
docs: omit implementation details from hooks reference

serikjensen:
## Summary

Tightens the hook reference docs to focus on what partners need to build forms, removing implementation-detail leaks that don't help integrators and that previously could mislead readers into wiring forms in non-recommended ways.

## Changes

- Removed the `Advanced: Fields Metadata` section from `hooks.md` so partners aren't steered toward consuming `fieldsMetadata` directly. Field components consume metadata internally; the recommended path is `Fields` (with `FieldComponent` for UI overrides).
- Reworded the `Fields.Ssn` section to describe the masked placeholder + auto-waived `REQUIRED` rule behaviorally instead of via `fieldsMetadata.ssn.hasRedactedValue`.
- Replaced internal symbols with behavioral descriptions: `NAME_REGEX` (Employee Details), `FLSA_OVERTIME_SALARY_LIMIT` (Compensation), `rev_2020_w4` (Federal Taxes).
- Stripped schema-builder vocabulary (`'never'` / `'always'` / "predicate", "schema declares/enforces") from Pay Schedule, Compensation, and `hooks.md` in favor of partner-facing language.
- Dropped the internal `Rule` column from the Compensation required-fields table; the remaining columns convey the same partner-relevant info.
- Removed raw HTTP endpoint paths (e.g. `GET /v1/work_addresses/{work_address_uuid}`, `POST`) from Work Address mode descriptions in favor of mode-only language.

What was intentionally preserved:

- `fieldsMetadata` in return-type signatures and the `SignEmployeeFormFieldsMetadata` row in the Exported Types table — they accurately describe the public return shape without prescribing direct consumption.
- `react-hook-form` / `Zod` / `UseFormReturn` references — partner-relevant framework context, especially for the `Advanced: Hook Form Internals` escape hatch.
- Cross-references to other public hooks (e.g. `useCurrentWorkAddressForm`, `useWorkAddressManagement`).

## Testing

- Docs-only change; no code touched.
- Re-grepped \`docs/hooks/\` to confirm no remaining \`NAME_REGEX\`, \`FLSA_OVERTIME_SALARY_LIMIT\`, \`rev_2020_w4\`, \`/v1/\`, \`'never'\`/\`'always'\` requiredness, or "predicate rule" references outside legitimate framework/behavioral usage.
- Visual review of each modified \`docs/hooks/*.md\` file confirmed updates read cleanly.


docs/hooks/hooks.md
mariechatfield: Is the semi-colon before the JSX required? It always looks a little odd to my eye

mariechatfield, line 91: Would you mind adding an example of two forms? I think I finally pictured it after reading the pros/cons of each approach but actually showing it in code might help with comprehension

mariechatfield: Do we not have any default validation messages for hooks?

mariechatfield (approved):

docs/hooks/hooks.md
serikjensen: Not currently! but this is a good catch, the test here is incorrect. It displays with "REQUIRED"

Going to correct

i'd be open to providing default validation messages, we'd just want to be sure that we aren't putting translations in hooks and we'd also want to be sure that we are providing the translations in the downstream component where the hook is used. I figured that might be easy to miss if the validation errors appeared to be rendering properly

serikjensen: yup!

serikjensen: Updated!

PR #1692, opened 2026-05-06
fix: render terminated employees on the time-off policy detail

jeffredodd:
## Summary
- The Employees tab on the time-off policy detail screen fetched its lookup data with `terminated: false`. Terminated employees can still be attached to a policy, so the map in `deriveEmployees` missed them and their rows rendered with empty names and job titles (screenshot 1 in plan).
- Refactored to fetch by `uuids` directly from `policy.employees`. We already know which employees the policy has, so we get exactly that set without filtering by termination state, and avoid over-fetching the full company roster just to enrich a handful of rows.
- Adds a regression test asserting the hook is called with `uuids` (the policy's attached employee uuids) and not `terminated: false`.

This is PR 8 in the time-off bug-fix slate captured in `~/.claude/plans/ok-i-want-to-partitioned-balloon.md`.

## Test plan
- [x] `npx tsc --noEmit` — clean
- [x] `npm run test -- --run src/components/UNSTABLE_TimeOff/TimeOffPolicyDetail/` — 37/37 passing
- [ ] Manual: attach an employee to a sick policy, dismiss them, reopen the policy detail Employees tab — confirm name and job title still render

### DIFF
--- a/src/components/UNSTABLE_TimeOff/TimeOffPolicyDetail/TimeOffPolicyDetail.test.tsx
+++ b/src/components/UNSTABLE_TimeOff/TimeOffPolicyDetail/TimeOffPolicyDetail.test.tsx
@@ -40,16 +40,10 @@ vi.mock('@gusto/embedded-api/react-query/timeOffPoliciesGet', () => ({
   }),
 }))
 
+const mockUseEmployeesListSuspense = vi.fn()
+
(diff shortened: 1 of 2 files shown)


jeffredodd:
Closing — the underlying issue was misdiagnosed. The existing code with `terminated: false` correctly populates names; the `uuids` filter approach in this PR was what caused names to render blank.

PR #1704, opened 2026-05-07
fix: hide holiday pay option when company already has a policy

jeffredodd:
## Summary

- A Gusto company can have **only one** holiday pay policy at a time. The API rejects `POST /companies/{id}/holiday_pay_policy` when one already exists. Today the `PolicyTypeSelector` radio always offers Holiday pay as a third option, so users with an existing policy can pick it, submit, and hit the API error.
- Match `gws-flows` ([app/views/company/time_off/time_off_management/new.html.erb#L12-L20](https://github.com/Gusto/gws-flows/blob/main/app/views/company/time_off/time_off_management/new.html.erb#L12-L20)) by completely omitting the radio option from the DOM when a holiday policy is already present. Detected via the existing `useHolidayPayPoliciesGet` query — same pattern `PolicyList` uses.
- Added presentation- and container-level tests, plus a `HolidayPolicyAlreadyExists` Storybook story.

## Test plan

- [x] `npm run test -- --run src/components/UNSTABLE_TimeOff/PolicyTypeSelector` (26/26 pass)
- [x] `npm run test -- --run src/components/UNSTABLE_TimeOff` (336/336 pass)
- [x] `npx eslint src/components/UNSTABLE_TimeOff/PolicyTypeSelector` clean
- [ ] Storybook: `Domain/TimeOff/PolicyTypeSelector → HolidayPolicyAlreadyExists` shows only Time off / Sick leave
- [ ] sdk-app smoke: with an existing holiday policy, "Create policy" shows two options; after deleting the policy, three options return

### DIFF
--- a/src/components/UNSTABLE_TimeOff/PolicyTypeSelector/PolicyTypeSelector.tsx
+++ b/src/components/UNSTABLE_TimeOff/PolicyTypeSelector/PolicyTypeSelector.tsx
@@ -21,10 +21,7 @@ export function PolicyTypeSelector(props: PolicyTypeSelectorProps) {
 function Root({ companyId, defaultPolicyType }: PolicyTypeSelectorProps) {
   const { onEvent } = useBase()
 
-  const holidayQuery = useHolidayPayPoliciesGet(
-    { companyUuid: companyId },
-    { throwOnError: () => false },
-  )
+  const holidayQuery = useHolidayPayPoliciesGet({ companyUuid: companyId })
   const holidayPolicyExists = Boolean(holidayQuery.data?.holidayPayPolicy)
 
   const handleContinue = (policyType: PolicyType) => {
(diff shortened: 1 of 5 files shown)


src/components/UNSTABLE_TimeOff/PolicyTypeSelector/PolicyTypeSelector.tsx
```diff
@@ -17,9 +18,15 @@ export function PolicyTypeSelector(props: PolicyTypeSelectorProps) {
   )
 }
 
-function Root({ defaultPolicyType }: PolicyTypeSelectorProps) {
+function Root({ companyId, defaultPolicyType }: PolicyTypeSelectorProps) {
   const { onEvent } = useBase()
 
+  const holidayQuery = useHolidayPayPoliciesGet(
+    { companyUuid: companyId },
+    { throwOnError: () => false },
```
dmortal: is this deliberate?

dmortal (approved):

src/components/UNSTABLE_TimeOff/PolicyTypeSelector/PolicyTypeSelector.tsx
jeffredodd: Good catch, it was from debugging during an outage yesterday. Will remove.

PR #1717, opened 2026-05-07
fix: await delete handler before closing policy delete dialog

krisxcrash:
## Summary
- The delete confirmation dialog in `PolicyListPresentation` closed optimistically before the async delete resolved
- If the API call failed, the user saw no dialog and the error was not clearly surfaced
- Now `handleConfirmDelete` awaits `onDeletePolicy` before dismissing the dialog, keeping the loading state visible during the API call

## Test plan
- [x] All 25 PolicyList tests pass
- [ ] Manual: Delete a policy and verify the dialog stays open with loading spinner until the delete completes

### DIFF
--- a/src/components/UNSTABLE_TimeOff/PolicyList/PolicyListPresentation.tsx
+++ b/src/components/UNSTABLE_TimeOff/PolicyList/PolicyListPresentation.tsx
@@ -45,9 +45,9 @@ export function PolicyListPresentation({
     setDeletePolicyDialogState({ isOpen: false, policy: null })
   }
 
-  const handleConfirmDelete = () => {
+  const handleConfirmDelete = async () => {
     if (deletePolicyDialogState.policy) {
-      onDeletePolicy(deletePolicyDialogState.policy)
+      await onDeletePolicy(deletePolicyDialogState.policy)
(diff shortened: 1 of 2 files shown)


src/components/UNSTABLE_TimeOff/PolicyList/PolicyListPresentation.tsx
```diff
@@ -45,9 +45,9 @@ export function PolicyListPresentation({
     setDeletePolicyDialogState({ isOpen: false, policy: null })
   }
 
-  const handleConfirmDelete = () => {
+  const handleConfirmDelete = async () => {
     if (deletePolicyDialogState.policy) {
-      onDeletePolicy(deletePolicyDialogState.policy)
+      await onDeletePolicy(deletePolicyDialogState.policy)
```
jeffredodd, line 50: Good catch

jeffredodd (approved):

src/components/UNSTABLE_TimeOff/PolicyList/PolicyListPresentation.tsx
krisxcrash: ty!!

PR #1727, opened 2026-05-07
feat(sdk-app): manual token mode with named presets

jeffredodd:
## Summary
- Adds an Auto / Manual mode toggle in the Settings panel. In Manual mode the user pastes a flow token + entity IDs and the proxy uses those values without auto-provisioning or token polling.
- Persists the active mode + config in localStorage (\`sdk-app-manual-config\`) and rehydrates the dev server on app boot via two new middleware endpoints (\`/sdk-app/api/set-manual-token\`, \`/sdk-app/api/restore-auto-token\`).
- Adds named saved presets under \`sdk-app-manual-config-saves\`: a dropdown lets users save the current draft under a name (e.g. \"Acme Test\") and recall it later. Editing fields clears the selection so the draft state is obvious. Apply still has to be clicked to commit.

## Why
Demos sometimes need to run against a specific company or partner-supplied credentials. Auto-provisioning works well for the default flow but can't be redirected. Saved presets make it easy to keep multiple demo contexts ready (engineer's own demo, customer-specific data, etc.) without re-pasting tokens.

## Test plan
- [ ] \`npm run sdk-app\` boots in Auto mode unchanged.
- [ ] In Settings → Mode, toggle to Manual; manual fields appear, Demo Management + Entity IDs are hidden.
- [ ] Paste a known-good flow token + companyId, click Apply; an SDK component (e.g. \`/employee/Profile\`) renders live API data.
- [ ] Refresh the page — manual config persists and the component still renders.
- [ ] Save the current draft as a named preset; reload; confirm the preset is still in the dropdown and selecting it repopulates fields.
- [ ] Switch back to Auto — auto-provisioned token resumes working without restarting Vite (proxy display flips back to \`flow-token\`, \`/sdk-app/api/validate-token\` returns valid).
- [ ] Token-expired polling does not fire while in Manual (status pill reads \"Unknown\").


krisxcrash (approved):

serikjensen (approved):
This looks good! we'll want to rigorously test SDK app over the next few days to guard against regressions

PR #1736, opened 2026-05-08
refactor(EditCompensation): consume useJobForm + useCompensationForm hooks

serikjensen:
Migrates the Compensation flow's edit screen onto the public hook contract introduced by #1735.

Opening as a **draft stacked on `refactor/job-compensation-hooks-split`** because the migration validates the hook contract end-to-end against the existing integration tests. Will rebase onto `main` and unmark draft once #1735 lands.

## Summary

- Replace `BaseComponent` + `EditCompensationPresentation` with a single `BaseBoundaries`-wrapped Root that instantiates `useJobForm` and `useCompensationForm` and threads them through `composeSubmitHandler`. The job-then-compensation submit chain is preserved (POST job → PUT stub comp on create, PUT job → PUT comp on update). Events (`EMPLOYEE_JOB_CREATED`/`UPDATED`, `EMPLOYEE_COMPENSATION_UPDATED`) and the `onSaved(Compensation)` contract are unchanged.
- Delete the duplicate internal `compensationSchema.ts` and the `EditCompensationPresentation`/test pair: form behavior is now covered by the hooks' unit tests (`useJobForm.test.tsx`, `useCompensationForm.test.tsx`, `compose.test.tsx`). The integration suite (`Compensation.test.tsx`) is **unmodified** and continues to assert screen-level behavior.
- Promote `title` (job) and `flsaStatus`/`rate`/`paymentUnit` (comp) via `optionalFieldsToRequire: { update: [...] }` so the screen always renders these fields as required. The hooks default these to required only on `'create'`; the screen-level UX always required them.
- Render fields inline with `formHookResult={...}` props because the screen interleaves job and compensation fields. Keep the FLSA change warning as screen-local presentation state via `useWatch` on the comp form's `hookFormInternals.formMethods.control`. The hook still exposes `data.canTriggerCarveOut` for partners who want to drive their own confirmation prompt.

## Hook-design fixes that landed on #1735 because of this migration

The migration revealed two contract mismatches that #1735 needed before this PR could pass the existing integration tests without modification:

1. `flsaStatus` validator made `.optional()` so the field can render its empty placeholder when nothing is provided. Schema-level `requiredFieldsConfig` still enforces it on submit in create mode. New fallback chain inherits from the employee's primary job's current compensation when adding a secondary job.
2. Reorder the FLSA rate validation so `RATE_EXEMPT_THRESHOLD` surfaces before `RATE_MINIMUM` for `EXEMPT` + low rate (matches pre-migration UX).

## Test plan

- [x] `npm run test -- --run src/components/Employee/Compensation` — 85/85 ✅ (integration suite untouched)
- [x] `npm run lint:check src/components/Employee/Compensation` — 0 errors
- [x] Smoke-test `Employee.Compensation` flow end-to-end in Storybook / sdk-app (no-jobs onboarding, single-Exempt edit, multi-job FLSA carve-out)
- [x] Rebase onto `main` and unmark draft after #1735 merges

## Verifying Compensation

https://github.com/user-attachments/assets/897eaad7-6a58-48ce-8b60-e7e5fb10b258

https://github.com/user-attachments/assets/d051a9ae-c7b4-4654-8491-758bd0fd94cc


jeffredodd (approved):
LGTM

krisxcrash (approved):

PR #1747, opened 2026-05-08
fix: remove hardcoded colors in PolicySettings for theme compliance

krisxcrash:
## Summary
- Removes hardcoded hex color overrides (`#414651`, `#535862`, `#181d27`) from `PolicySettings.module.scss`
- These overrode `--g-inputLabelColor` and `--g-inputDescriptionColor` CSS custom properties, bypassing the theme token system
- With these removed, PolicySettings inherits proper colors from ThemeProvider, ensuring WCAG 1.4.3 contrast compliance across partner themes

## Test plan
- [ ] Verify PolicySettings form renders with correct label/description colors in Storybook
- [ ] Confirm no visual regression in light theme (colors should be identical or very close)
- [ ] Validate contrast ratios meet 4.5:1 minimum


src/components/UNSTABLE_TimeOff/PolicySettings/PolicySettings.module.scss
```diff
@@ -1,8 +1,5 @@
 .policySettings {
-  --g-inputLabelColor: #414651;
```
jeffredodd, line 2: Are these already tokened? I'm guessing so!

jeffredodd (approved):

src/components/UNSTABLE_TimeOff/PolicySettings/PolicySettings.module.scss
krisxcrash: Yes! They're already tokened via ThemeProvider. Updated the approach — instead of deleting the overrides entirely (which would change the visual hierarchy since the defaults are darker), I now reference the existing theme tokens:

- `--g-inputLabelColor: var(--g-colorBodySubContent)` (was hardcoded `#414651` / neutral[700])
- `--g-inputDescriptionColor: var(--g-colorSecondaryContent)` (was hardcoded `#535862` / neutral[600])

This preserves the original design intent (lighter labels for the toggle+input pattern) while making it theme-adaptive — partners with custom themes will get the correct colors.

PR #1755, opened 2026-05-08
feat(sdk-app): theme panel resolved values, shared resizable panels, color swatch polish

jeffredodd:
## Summary

- **Theme panel resolved values**: The theme editor now shows the resolved (default) theme token values as placeholder text in inputs and as actual colors in swatches — so you can see what each token is set to before making any overrides, in both light and dark mode
- **Color swatch contrast ring**: Color swatches now have a layered inset box-shadow ring so they're visible even when set to white or a very light color
- **Shared resizable panels**: Code, Theme, and Settings panels now all share a single `RightPanelShell` wrapper with a drag handle for resizing — switching between panels preserves the panel width (persisted to localStorage)
- **Cursor bleed fix**: The col-resize cursor was bleeding over the entire app due to absolute-positioned handle inside a sticky container; fixed by switching to a flex-row layout in `RightPanelShell` where the handle is a natural sibling with no absolute positioning
- **Theme shortcut**: Changed theme panel shortcut from `Cmd+T` (conflicts with browser new tab) to `Cmd+J`

## Test plan

- [ ] Open the theme panel — all token rows show the resolved color/value as placeholder text and color swatches show the current theme color
- [ ] Switch to dark mode — swatches and placeholders update to dark theme values
- [ ] Override a color token — swatch and input show the override; "Reset" button appears; clearing resets to resolved value
- [ ] White/light color swatches have a visible ring border for contrast
- [ ] Drag the panel resize handle to resize — width persists when switching between Theme, Settings, and Code panels
- [ ] Hover over the app content area (not the handle) — no col-resize cursor
- [ ] `Cmd+J` opens/closes the theme panel; `Cmd+T` opens a new browser tab as expected
- [ ] Shortcut helper shows updated `⌘J` binding for theme panel

### DIFF
--- a/sdk-app/src/App.tsx
+++ b/sdk-app/src/App.tsx
@@ -21,6 +21,15 @@ import { CodePanel } from './CodePanel'
 import { CurrentComponentProvider } from './CurrentComponentContext'
 import { useDemoChrome } from './useDemoChrome'
 import { findDemoChrome, SDK_NATIVE_CHROME_ID } from './demoChromes/registry'
+import {
+  ThemePanel,
+  useThemePanel,
+  ThemeEditorContext,
+  useThemeEditorState,
+  DesignSystemContext,
+  useDesignSystemState,
+} from './ThemePanel'
+import { RightPanelShell } from './RightPanelShell'
 
 const THEME_CYCLE: ThemeMode[] = ['system', 'light', 'dark']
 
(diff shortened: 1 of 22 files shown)


serikjensen (approved):

jeffredodd:
Two things, @copilot+claude-opus-4.7 can you fix the CI formatting issue? Also can you review copilot review feedback and see if we should implement it?

PR #1700, opened 2026-05-06
refactor: rebuild Employee.Compensation around a state machine + presentation/connected split

serikjensen:
## Summary

Rebuilds `Employee.Compensation` from the `mode`-flag, children-override
pattern into a robot3 state machine plus focused presentation/connected
component pairs. Mode-aware behavior (initial onboarding vs. managing
additional jobs) is moved out of the component and into contextual wrappers,
making `EditCompensation` mode-agnostic and configurable via `title`,
`submitCtaLabel`, `onSaved`, and `onCancel`.

### Highlights

- **State machine**: `initialEditJob` → `done` / `viewJobs`; `viewJobs` →
  `editJob` (add or edit); `editJob` → `viewJobs` via `RETURN_TO_LIST` or
  `CANCEL`. The machine never carries a mode flag — branching is encoded as
  distinct states with distinct contextual components.
- **Presentation/Connected split**: `JobsList[Presentation]` and
  `EditCompensation[Presentation]`, mirroring the pattern used by
  Deductions, StateTaxes, Locations, and Dismissal.
- **Tightened machine memoization**: `Compensation` recreates the machine only
  on scalar inputs that genuinely affect initial state (`employeeId`,
  `startDate`, `initialState`, `currentJobId`). `defaultValues` is
  intentionally omitted with a documented `eslint-disable` so partner-supplied
  object refs cannot blow away in-flight machine state. Other state-machine
  flows in the repo currently include `defaultValues` in deps and have the
  same latent fragility — Compensation is now the reference fix.
- **Removed dead `register` for hidden `flsaStatus` input**: RHF carries
  values from `defaultValues` through to submit without a registered field;
  proven by a new test in `EditCompensationPresentation.test.tsx`.
- **Schema extracted** to `compensationSchema.ts`; legacy `useCompensation.ts`
  removed.
- **Reference doc**: `docs/reference/jobs-and-compensations.md` captures the
  underlying API/model rules and the test plan derived from gws-flows EMF +
  zenpayroll.

### Tests

- **Integration suite preserved**: `Compensation.test.tsx` (21 tests) still
  covers every state-machine transition end-to-end across the no-jobs,
  single-Nonexempt, single-non-Nonexempt, and multi-job branches plus 2%
  shareholder visibility and rate validation messages.
- **New presentation-level unit tests**:
  - `JobsListPresentation.test.tsx` — list rendering, primary delete
    suppression, action wiring.
[...]

PR #1757, opened 2026-05-09
fix(Compensation): stabilize flow state machine across getJobs refetches

serikjensen:
## Summary

PR #1700 introduced a `robot3` state machine wrapper around `Employee.Compensation` whose `initialState` derived from a live `getJobs` query result. Each render recomputed the routing decision, so the `getJobs` invalidation that fires after every mutation re-seated the machine with a fresh snapshot — most visibly causing the JobsList view to flash on-screen between an `EditCompensation` chained job-then-compensation submit and the flow's transition to the next step.

This PR snapshots the routing decision via `useState` lazy initialization from the first suspense-resolved jobs payload (guaranteed fresh by `useSuspenseQuery`) and never recomputes. The state machine identity stays stable across all subsequent refetches, so the user's place in the flow is preserved through chained mutations.

Callers that legitimately need to re-derive the routing for a different employee should remount via `key`, which matches the existing contract used by `OnboardingFlow`, `PayrollExecutionFlow`, and `DismissalFlow`.

## Why this is safe

- `useSuspenseQuery` guarantees `jobsData` is populated before `CompensationFlow` renders, so the lazy initializer always sees real data.
- Dependent refetches that legitimately change routing semantics (e.g. switching to a different employee) cause an unmount+remount from upstream, not a re-render of the same instance.
- `initialState` and `currentJobId` remain in the `useMemo` deps array as honest signals — they're now stable identities sourced from `useState`, so they never trigger machine recreation but the eslint-disable on the deps line is preserved for clarity.

## Test plan

- [ ] Compensation tests pass locally (`npm run test -- --run src/components/Employee/Compensation`)
- [ ] Manual: trigger an onboarding-style chained job-then-compensation save; confirm the JobsList view does not flash between mutations
- [ ] Manual: confirm steady-state JobsList navigation (add/edit secondary, edit primary) still routes correctly when remounted via `key` for a different employee

### DIFF
--- a/src/components/Employee/Compensation/Compensation.tsx
+++ b/src/components/Employee/Compensation/Compensation.tsx
@@ -1,4 +1,4 @@
-import { useMemo } from 'react'
+import { useMemo, useState } from 'react'
 import { createMachine } from 'robot3'
 import { useTranslation } from 'react-i18next'
 import { useJobsAndCompensationsGetJobsSuspense } from '@gusto/embedded-api/react-query/jobsAndCompensationsGetJobs'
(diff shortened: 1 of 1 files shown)


serikjensen:
Wrong branch placement on my part — the state machine fix was supposed to land on the migration PR (#1736), not as an independent fix. Cherry-picked onto `refactor/migrate-edit-compensation-to-hooks` instead. Deleting this branch.

PR #1764, opened 2026-05-11
docs(SDK-810): date normalization audit and solution spike

mariechatfield:
## Summary

Audit the existing usage of dates, date normalization functions, and and logic across the app and identify an approach to make date handling more consistent and avoid timezone errors when going between formats.

## Related

- Jira ticket: [SDK-810](https://gustohq.atlassian.net/browse/SDK-810)

[SDK-810]: https://gustohq.atlassian.net/browse/SDK-810?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/rfc/2026-05-11--date-normalization.md
+++ b/rfc/2026-05-11--date-normalization.md
@@ -63,9 +63,10 @@ Functions that parse them: `normalizeToDate` handles `YYYY-MM-DD` safely via reg
 Wraps dates for API write mutations. Serializes internally via `.toISOString().slice(0, 10)` (UTC).
 
 - **Safe:** `new RFCDate("YYYY-MM-DD")` — string input is UTC midnight; serializes back correctly
-- **Unsafe:** `new RFCDate(new Date(someString))` — local-midnight `Date` is UTC midnight minus the offset; wrong day for UTC+ users ⚠️
+- **Safe (but redundant):** `new RFCDate(new Date("YYYY-MM-DD"))` — `new Date` parses the string as UTC midnight; `RFCDate` serializes back via UTC; roundtrip is correct. The intermediate `Date` is unnecessary and confusing, but not buggy.
+- **Unsafe:** `new RFCDate(new Date(localMidnightDate))` — if the `Date` was constructed via `new Date(year, month, day)` (local midnight), then UTC serialization reads the previous day for UTC+ users ⚠️
 
-Three files use the unsafe pattern: `useEmployeeDetailsForm` (`dateOfBirth`), `useHomeAddressForm` (`effectiveDate`), `useWorkAddressForm` (`effectiveDate`).
+Three files use `new RFCDate(new Date(string))`: `useEmployeeDetailsForm` (`dateOfBirth`), `useHomeAddressForm` (`effectiveDate`), `useWorkAddressForm` (`effectiveDate`). Because these fields are typed as `z.iso.date()` in their Zod schemas, the value is always a `YYYY-MM-DD` string — so the UTC roundtrip cancels and no data-integrity bug occurs. The pattern is redundant noise, not a bug.
 
 ### Central Utilities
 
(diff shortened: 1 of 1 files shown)


rfc/2026-05-11--date-normalization.md
```diff
@@ -0,0 +1,226 @@
...
+| `PayPeriod.fromPayPeriodData`                                        | `new Date(YYYY-MM-DD)` → UTC midnight                                                                                                          |
+| `Payroll/helpers.ts` (compensation effective dates)                  | `new Date(effectiveDate)` on string → UTC midnight                                                                                             |
+| `useEmployeeStateTaxesForm` deserialization                          | `new Date(trimmed)` on string → UTC midnight                                                                                                   |
+| `CalendarPreview.tsx:32`                                             | `new Date(dateValue.toString())` — behavior varies                                                                                             |
+| `useEmployeeDetailsForm`, `useHomeAddressForm`, `useWorkAddressForm` | `new RFCDate(new Date(string))` — UTC roundtrip (**live data-integrity bug** for UTC+ users: wrong birth date / effective date sent to API)    |
```
mariechatfield: Found an active bug for these hooks by auditing dates. I'll create a new bug ticket (https://gustohq.atlassian.net/browse/SDK-828) and work on those / add tests so we can address it independently of the rest of the date normalization or migration.

rfc/2026-05-11--date-normalization.md
```diff
@@ -0,0 +1,227 @@
...
+
+### Adapter Boundary Is Fixed
+
+`DatePickerProps` and `DateRangePickerProps` are the partner-facing API and must keep `Date` at the boundary. Everything above this boundary is internal and can change.
+
+Conversions happen at two edges:
+
+- **Inbound** (adapter → form): `value.getFullYear/getMonth/getDate` → `new CalendarDate(...)` — reads local, safe
+- **Outbound** (form → adapter): `new Date(cd.year, cd.month - 1, cd.day)` — local midnight, safe
+- **Display**: `cd.toDate(getLocalTimeZone()).toLocaleDateString(...)`
```
serikjensen: On this one, i'm wondering if we should just standardize on iso string without time components as the external partner contract. We could add some redundancy to where we parse the string `.split('T')[0]` to ensure we only get the date portion?

My thought here is that the time components could still introduce bugs depending on where the partner is creating the Date object? So wondering if we just need to get out of the Date game here altogether

It'd be a breaking change, but i think the partner migration would be pretty straightforward.

Thoughts?

serikjensen (commented):
Excellent findings! sounds like there's an opportunity to just standardize on CalendarDate for our internal operations. 

One thing i didn't see included was `defaultValues`.  Some of the date pickers that live in forms allow for `defaultValues` setting the date via props. This caused a BQE bug recently. So if we did make the move to standardize on a specific format, i might recommend iso string sans time as the universal value for all the things partner facing for clarity. That would include adapters with date values as outlined below as well as default values for forms.

My other question is if we migrate to CalendarDate for all internal operations, does this play well with existing endpoints? I know we have RFCDate that is used as a step in api requests so wondering if that could be updated to just get a date string from CalendarDate, or if there's a reason we can't just standardize on the dates that the API is using

Thanks for putting this together!

rfc/2026-05-11--date-normalization.md
mariechatfield: From our standup conversation:

- Switching the API at the component adapter to a string with a specific format seems like a reasonable suggestion
- I need to do some research on the internalized/date library to confirm we want to double down on it outside of react-aria
- I will update this RFC and then open it for review, but we don't expect to actually make any of the recommended changes immediately (apart from the bugfix that's already been merged)

PR #1767, opened 2026-05-11
fix(SDK-828): utc roundtrip bug in date picker field

mariechatfield:
## Summary

Fix existing UTC roundtrip bugs in the `DatePickerField`:
  - Get the calendar date parts from the `Date` object directly instead of parsing from `toISOString`
  - `toISOString` displays the date in UTC; midnight on May 11 in UTC+2 is 10pm on May 10 in UTC... so we get `May 10` instead of `May 11`

Add tests to datepickers and formatters to validate that UTC+ timezones handle saving the correct values.

## Demo

First: set your local timezone to a UTC+ zone

Screenshot 2026-05-11 at 12 43 51 PM

On main: attempt to set an employee's birthday to another date. 
- The read-only date is one day _after_ the original value in the input
- Choosing a new date on the dropdown picker will result in a different date being updated in the input
- Using the up and down arrows to increment skips multiple days

https://github.com/user-attachments/assets/fff3693a-bed3-45f9-b31a-acdb7b2037a7

On this branch: attempt the same. 
- Read-only date matches the original input value.
- New date via the dropdown picker _or_ the arrow keys works as expected.

https://github.com/user-attachments/assets/b7362b9f-5883-43eb-a76c-e767daebf2e0

## Related

- Jira ticket: [SDK-828](https://gustohq.atlassian.net/browse/SDK-828)

[SDK-828]: https://gustohq.atlassian.net/browse/SDK-828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Common/Fields/DatePickerField/DatePickerField.test.tsx
```diff
@@ -0,0 +1,218 @@
...
+  describe('UTC+ timezone edge cases', () => {
+    beforeEach(() => {
+      vi.stubEnv('TZ', 'Europe/Paris') // UTC+2 in summer: local midnight June 15 = UTC June 14 22:00
+    })
+
+    afterEach(() => {
+      vi.unstubAllEnvs()
+    })
+
+    it('Date mode: submits the correct date when local midnight falls in UTC previous day', async () => {
```
mariechatfield, line 165: This is the test that failed when we first introduced it before the code change. (TDD with claude 🚀 )

jeffredodd: Yay Claude TDD Club! 🎉

src/helpers/dateFormatting.ts
```diff
@@ -203,22 +206,7 @@ export const normalizeDateToLocal = (date: Date | null): Date | null => {
...
-  const numbers = parts.map(Number)
-  const year = numbers[0]!
-  const month = numbers[1]!
-  const day = numbers[2]!
-  if (isNaN(year) || isNaN(month) || isNaN(day) || month < 1 || month > 12 || day < 1 || day > 31) {
-    return null
-  }
-
-  return new Date(year, month - 1, day)
+  return new Date(date.getFullYear(), date.getMonth(), date.getDate())
```
jeffredodd, line 209: Just a general thought regarding the way the code was setup. 

I think what you're fixing is solved by this code change. 

What i'm trying to remember is why we were splitting time zone at all, it might be worth finding the original ticket that authored this code just to validate the test cases are covered for both why this was created and what we're trying to solve now.

serikjensen (commented):
Wanted to check in on the feedback here https://github.com/Gusto/embedded-react-sdk/pull/1764#pullrequestreview-4274206849 my thought is we may just be able to simplify the contract for date pickers since Date is being a pain

src/helpers/dateFormatting.ts
mariechatfield: I think this is the PR you're remembering: https://github.com/Gusto/embedded-react-sdk/pull/423

We are still dropping the hours component and _just_ considering the calendar date, which I _think_ is what this was concerned about. Does that ring any bells?

jeffredodd: Thank you for digging this up. I was hoping I had a more concrete issue here but I guess not.  I think i'm ok with this code then given all of this.

jeffredodd (approved):

serikjensen (approved):

PR #1773, opened 2026-05-11
fix: prevent skeleton/gap pop-in for empty TransitionPayrollAlert

jeffredodd:
## Summary

`TransitionPayrollAlert` is rendered unconditionally on `PayrollLanding`, but it has nothing to show unless the company has unprocessed transition pay periods. Today, even when it ends up returning `null`, the component still produces a visible artifact in the layout:

1. While its Suspense queries are loading, `BaseComponent` shows its default 200px skeleton fallback above the tabs.
2. After load, `BaseComponent` keeps an empty `FadeIn` `<div>` in the DOM even when `Root` returns `null`. Because the parent `Flex` uses `gap={32}`, that empty wrapper still consumes a gap slot — producing a small, visible "pop-in" as content finishes loading.

This PR hoists the Suspense queries above `BaseComponent`. A new gate component does the data fetching, computes `groupedPayPeriods`, and returns `null` *before* mounting `BaseComponent` when there's nothing to render. The outer wrapper becomes `<Suspense fallback={null}>`, so:

- Loading state → nothing in the DOM
- Resolved, no transition pay periods → nothing in the DOM
- Resolved, has transition pay periods → `BaseComponent` + `FadeIn` + alert mount as usual

No layout shift, no skeleton flash, and no behavioral change when there *is* a transition alert to show.

## Fix Video

Uploading Screen Recording 2026-05-11 at 2.41.16 PM.mov…

## Test plan

- [ ] `npm run test -- --run src/components/Payroll/TransitionPayrollAlert/` passes (10/10)
- [ ] Visit `PayrollLanding` for a company with no transition pay periods — no skeleton box and no gap above the tabs at any point during page load
- [ ] Visit `PayrollLanding` for a company with an active transition pay period — alert still renders correctly with skip / run-payroll handlers wired up
- [ ] Skip-payroll mutation still surfaces errors through the alert (no regression from removing the outer `BaseComponent` wrap around the loading state)

### DIFF
--- a/src/components/Payroll/TransitionPayrollAlert/TransitionPayrollAlert.tsx
+++ b/src/components/Payroll/TransitionPayrollAlert/TransitionPayrollAlert.tsx
@@ -1,4 +1,4 @@
-import { useMemo, useState, useCallback } from 'react'
+import { Suspense, useMemo, useState, useCallback } from 'react'
 import { usePaySchedulesGetPayPeriodsSuspense } from '@gusto/embedded-api/react-query/paySchedulesGetPayPeriods'
 import { usePaySchedulesGetAllSuspense } from '@gusto/embedded-api/react-query/paySchedulesGetAll'
 import { usePayrollsSkipMutation } from '@gusto/embedded-api/react-query/payrollsSkip'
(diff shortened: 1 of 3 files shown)


jeffredodd:
@copilot can you fix the issues with CI?

src/components/Payroll/TransitionPayrollAlert/TransitionPayrollAlert.tsx
```diff
@@ -1,39 +1,79 @@
...
+          FallbackComponent={NullFallback}
+          onReset={resetQueries}
+          onError={handleGateError}
+        >
+          <Suspense fallback={null}>
+            <TransitionPayrollAlertGate {...props} />
+          </Suspense>
+        </ErrorBoundary>
+      )}
+    </QueryErrorResetBoundary>
```
serikjensen: I think what you actually want here is BaseBoundaries. You could update that component to accept a SuspenseFallback that defaults to the current and then set it to null?

I worry a little bit about this approach tho where we are doing a NullFallback for the error state. Can't this component still get into an error state? Confused also why would be hiding the load in this case?

serikjensen (approved):

PR #1790, opened 2026-05-12
fix: filter non-numeric input on balance fields (SDK-863)

krisxcrash:
## Summary
- Balance fields used plain `TextInput` which accepted non-numeric characters (letters, symbols, etc.)
- Added an `isNumericInput` validation predicate that rejects any value not matching `/^\d*\.?\d*$/` — this allows digits, a single optional decimal point, and empty input
- Keeps the existing `TextInput` component to avoid type mismatches, re-render issues, and accessibility regressions that `NumberInput` introduced

## Test plan
- [x] Verify balance fields reject alphabetic input (`abc`)
- [x] Verify balance fields reject a second decimal point (`1.2` → typing `.` is blocked)
- [x] Verify valid numeric input still passes through (`4`, `1.5`)
- [x] Run `npm run test -- --run src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesPresentation.test.tsx`

### DIFF
--- a/src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesPresentation.tsx
+++ b/src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesPresentation.tsx
@@ -74,16 +74,22 @@ export function SelectEmployeesPresentation({
                       return <Text>{originalBalances?.[employee.uuid] ?? '0'}</Text>
                     }
                     return (
-                      <Components.TextInput
+                      <Components.NumberInput
                         name={`balance-${employee.uuid}`}
                         label={t('startingBalanceColumn')}
                         shouldVisuallyHideLabel
                         aria-labelledby={`employee-name-${employee.uuid} ${balanceColHeaderId}`}
-                        value={balances?.[employee.uuid] ?? ''}
-                        onChange={(value: string) => {
-                          onBalanceChange(employee.uuid, value)
+                        value={
+                          balances?.[employee.uuid] != null
+                            ? Number(balances[employee.uuid])
+                            : 0
+                        }
+                        onChange={(value: number) => {
+                          onBalanceChange(employee.uuid, String(value))
                         }}
-                        placeholder="0"
+                        min={0}
+                        minimumFractionDigits={1}
+                        maximumFractionDigits={2}
                       />
                     )
                   },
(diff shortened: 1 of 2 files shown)


src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesPresentation.tsx
```diff
@@ -81,7 +81,10 @@ export function SelectEmployeesPresentation({
                         aria-labelledby={`employee-name-${employee.uuid} ${balanceColHeaderId}`}
                         value={balances?.[employee.uuid] ?? ''}
                         onChange={(value: string) => {
-                          onBalanceChange(employee.uuid, value)
+                          const numeric = value.replace(/[^0-9.]/g, '')
```
jeffredodd: Should we just use isNumeric? 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger

krisxcrash: Good call — replaced the regex-replace-then-compare approach with a single `isNumericInput` predicate (`/^\d*\.?\d*$/`). This handles digits, one optional decimal, and empty input in one check. Went with a regex predicate rather than `Number()` / `Number.isFinite()` since we need to allow partial input like `"1."` while typing, which `Number("1.")` would accept but feels fragile to rely on across engines. Also added test coverage for the multiple-dot case.

src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesPresentation.tsx
```diff
@@ -81,7 +81,10 @@ export function SelectEmployeesPresentation({
...
+                          if (numeric !== value) return
+                          if ((numeric.match(/\./g) ?? []).length > 1) return
```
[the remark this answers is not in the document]
krisxcrash: Addressed — added a test that starts with `balances: { '1': '1.2' }` and types `.`, asserting `onBalanceChange` is not called. This exercises the single-decimal-point constraint.

jeffredodd (approved):

PR #1791, opened 2026-05-12
fix: add max={20000} to policy settings number fields (SDK-864)

krisxcrash:
## Summary
- `NumberInputField` components for accrual maximum, balance maximum, carry-over limit, and waiting period had no upper bound, allowing values that trigger API 500 errors.
- Added `max={20000}` to all four fields.

## Test plan
- [ ] Verify each field rejects values above 20,000
- [ ] Verify values at or below 20,000 are accepted


src/components/TimeOff/PolicySettings/PolicySettingsPresentation.tsx
```diff
@@ -73,6 +73,7 @@ export function PolicySettingsPresentation({
                       placeholder={t('policySettings.numberOfHoursPlaceholder')}
                       isDisabled={!accrualMaximumEnabled}
                       min={0}
+                      max={20000}
```
jeffredodd, line 81: Just a thought, is it possible to get this somehow with Zod validations? Max is helpful but it doesn't guide the user tothe 'why' this doesn't go above 20000

mariechatfield: If we really wanted to be pedantic we could use `8760` as the max since that's the number of hours in a year but imo the `20,000` is so absurdly high for "max hours accrued in a year" or "max hours in balance" or "max rollover hours" that we don't strictly need a validation error.

mariechatfield (approved):

PR #1794, opened 2026-05-12
fix: surface errors inline inside edit balance modal (SDK-867)

krisxcrash:
## Summary
- Mutation errors from the edit balance modal went to `BaseLayout` Alert which rendered behind the modal overlay.
- Added an `errorMessage` prop to `EditEmployeeBalanceModal` with an inline `Alert` component.
- `TimeOffPolicyDetail` now catches errors from the balance update and passes them to the modal. Error is cleared on modal close.
- Added `editBalanceModal.updateError` i18n key as fallback.

## Test plan
- [ ] Trigger a balance update error and verify the error displays inside the modal
- [ ] Close and reopen the modal, verify the error is cleared
- [ ] Successful updates still close the modal and show the success flash


src/components/TimeOff/TimeOffPolicyDetail/TimeOffPolicyDetail.tsx
```diff
@@ -164,6 +164,7 @@ function Root({ policyId }: TimeOffPolicyDetailProps) {
     name: string
   } | null>(null)
   const [editBalanceState, setEditBalanceState] = useState<EditBalanceState | null>(null)
+  const [balanceError, setBalanceError] = useState<string | undefined>()
```
jeffredodd: Lets sync on this one. I think there is a better way to do this thats more in line with Base.tsx + will be better for hooks in the future too.

jeffredodd (changes requested):

jeffredodd (approved):

PR #1796, opened 2026-05-12
fix: sort policies alphabetically by name for stable display order (SDK-869)

krisxcrash:
## Summary
- Policy list used API response order directly, causing policies to appear in different order after deletion and refetch.
- Added `.sort((a, b) => a.name.localeCompare(b.name))` after mapping policies, before appending the holiday row.

## Test plan
- [ ] Verify policies appear in alphabetical order
- [ ] Delete a policy and verify the remaining policies maintain alphabetical order
- [ ] Verify the holiday row always appears last


mariechatfield (approved):
Works as described! It _was_ confusing that the holiday policy row is last but it sounds like that's on purpose?

Image

PR #1799, opened 2026-05-12
fix: humanize snake_case field names in API error messages (SDK-872)

krisxcrash:
## Summary
- API error messages contained raw snake_case field names (e.g., "accrual_rate is invalid").
- Added `humanizeFieldName` and `humanizeMessage` utilities to convert snake_case to Title Case in error messages.
- `toSDKFieldError` now applies humanization before returning error objects.

## Test plan
- [ ] Trigger an API validation error with a snake_case field name and verify the displayed message uses humanized text (e.g., "Accrual rate is invalid")


src/types/sdkError.ts
```diff
@@ -117,10 +117,23 @@ function hasErrorsArray(error: unknown): error is ErrorWithErrors {
   )
 }
 
+function humanizeFieldName(snakeCaseField: string): string {
+  if (!snakeCaseField) return snakeCaseField
```
dmortal, line 121: I wonder if instead this should return some generic value

dmortal (approved):

jeffredodd (approved):

PR #1800, opened 2026-05-12
fix: send separate PUT with complete:true after wizard final step (SDK-873)

krisxcrash:
## Summary
- `complete: true` was only sent via `PolicySettings` update, but the wizard flow skipped this step for some accrual methods.
- Removed `complete: true` from `PolicySettings.buildUpdateRequestBody`.
- Added a `markComplete` parameter to `submitDiff` in `SelectEmployeesTimeOff` that sends a separate PUT with `{ complete: true, version }` after employees are added (matching gws-flows behavior).
- Wizard mode now calls `submitDiff(toAdd, [], true)` to mark the policy complete.

## Test plan
- [ ] Complete a wizard flow and verify the policy is marked as complete
- [ ] Verify the separate PUT request is sent after the add-employees call
- [ ] Verify standalone policy settings save no longer sends `complete: true`


src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
```diff
@@ -167,6 +168,7 @@ function SelectEmployeesTimeOffInner({
   const { mutateAsync: addEmployees } = useTimeOffPoliciesAddEmployeesMutation()
   const { mutateAsync: removeEmployees, isPending: isRemovePending } =
     useTimeOffPoliciesRemoveEmployeesMutation()
+  const { mutateAsync: updatePolicy } = useTimeOffPoliciesUpdateMutation()
```
dmortal: do we need to handle loading state here? and errors?

src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
```diff
@@ -223,6 +225,20 @@ function SelectEmployeesTimeOffInner({
...
           policyResult = response.timeOffPolicy
         }
+        if (markComplete && policyResult) {
+          const version =
+            typeof policyResult === 'object' &&
+            policyResult !== null &&
+            'version' in policyResult
+              ? ((policyResult as { version: string }).version ?? '')
+              : ''
+          await updatePolicy({
```
dmortal: looks like we are not handling errors here

dmortal (changes requested):

dmortal (approved):

PR #1801, opened 2026-05-12
fix: clear policyResetDate for per_anniversary_year and send null to API (SDK-874)

krisxcrash:
## Summary
- Stale `policyResetDate` values could be sent for `per_anniversary_year`, which should have `policy_reset_date: null` per the API spec.
- Changed `policyResetDate` fallback from `undefined` to `null` in both `buildCreateRequestBody` and `buildUpdateRequestBody`, ensuring the field is explicitly cleared.
- Added an `onChange` callback on the `resetDateType` RadioGroup to reset month/day to defaults when switching to `per_anniversary_year`.

## Test plan
- [ ] Select per_anniversary_year and verify the API payload includes `policy_reset_date: null`
- [ ] Switch from per_calendar_year (with a custom date) to per_anniversary_year and verify stale dates are cleared


src/components/TimeOff/TimeOffManagement/PolicyConfigurationForm/PolicyConfigurationFormPresentation.tsx
```diff
@@ -73,6 +73,13 @@ export function PolicyConfigurationFormPresentation({
     }
   }, [resetMonth, getValues, setValue])
 
+  useEffect(() => {
+    if (resetDateType === 'per_anniversary_year') {
+      setValue('resetMonth', 1)
+      setValue('resetDay', 1)
+    }
+  }, [resetDateType, setValue])
+
```
mariechatfield, line 103: Do we always use `useWatch` combined with a `useEffect` like this? I tend to really dislike `useEffect` / observer patterns in general and I'm trying to understand how `react-hook-form` thinks about it.

For example, I would typically prefer using something like "when we actually modify the resetDateType value, if we are setting it to per_anniversary_year, make this change to the other parts of the field" but that might be what this is doing already? I'm honestly having a hard time figuring out `useWatch` vs `subscribe` in the docs: https://react-hook-form.com/docs/useform/subscribe

mariechatfield (commented):
I am so confused. I cannot, for the life of me, get changing to "by employee anniversary" to stick.

I save the form:
Image

The payload of the `PUT` and it's response both show `policy_reset_date: null`

Image
Image

But after submitting the settings page with no changes, the response for _that_ request shows up with `policy_reset_date: "01-01"`. And it's not in the payload for that call.

Image
Image

I am deeply confused but this happens over and over again. Do I have bad data? Am I doing this wrong?

krisxcrash:
> I am so confused. I cannot, for the life of me, get changing to "by employee anniversary" to stick.
> 
> I save the form: Image
> 
> The payload of the `PUT` and it's response both show `policy_reset_date: null`
> 
> Image Image
> But after submitting the settings page with no changes, the response for _that_ request shows up with `policy_reset_date: "01-01"`. And it's not in the payload for that call.
> 
> Image Image
> I am deeply confused but this happens over and over again. Do I have bad data? Am I doing this wrong?

You don't have bad data. I believe this is related to `complete: true`. When PolicySettings submits the update it sends complete: true without including policyResetDate, and the backend appears to default `policy_reset_date` to "01-01" when it's missing from a "completing" PUT. That's why the first save looks correct (the config step sends `policy_reset_date: null`) but the settings step undoes it.

This should actually be fixed by #1800 which moves `complete: true` out of PolicySettings entirely. Once that merges I'd expect the reset date to stick. I'll rebase this on top of that PR and verify.

src/components/TimeOff/TimeOffManagement/PolicyConfigurationForm/PolicyConfigurationFormPresentation.tsx
krisxcrash: Good question. `useWatch` is already needed here for rendering -- the component conditionally shows/hides the month/day selectors based on `resetDateType`. Given that, `useEffect` is the standard react-hook-form way to react to watched value changes.

`subscribe` is more for non-rendering side effects (it doesn't trigger re-renders), so it wouldn't replace `useWatch` here since we need the value for the conditional render anyway.

mariechatfield: Got it, thank you for explaining!

mariechatfield:
Sounds good! Especially since I know the initial API request is correct, as soon as that rebase is done I'm happy to approve

dmortal (approved):

src/components/TimeOff/TimeOffManagement/PolicyConfigurationForm/PolicyConfigurationFormPresentation.tsx
serikjensen: I am in agreement with Marie tho! you do have access to the onChange for the field components, so you can always try running this through a callback instead of an effect for when the resetDateType changes.

serikjensen (approved):

PR #1810, opened 2026-05-13
fix(SDK-878, SDK-880): responsive policy detail cards follow Tabs mode

aaronlee777:
## Summary

Make the policy detail cards and the employees-tab search bar share the same width as the Tabs control at every viewport size.

- **Wide container** (≥ 40rem / 640px): tab strip renders as a row. Cards cap at `30rem`, search bar caps at `20rem` (unchanged from before).
- **Narrow container** (< 40rem): tab strip renders as a Select dropdown that fills the container width. Cards and search bar now also span the full container, so the Select dropdown and the content beneath it line up.

## Approach

The previous version of this PR set a `data-tabs-mode` attribute on the SDK's `Tabs` root and read it from descendant CSS. That broke the contract that `Tabs` is swappable via `ComponentsContext` — if a partner supplied their own `Tabs`, the attribute would not exist and the layout would silently fall back to fixed widths.

Replaced with **CSS container queries** declared at the consumer level:

- `TimeOffPolicyDetail.module.scss` — wraps the two `Box` cards in a `.cardsContainer` (the query container). Above the `40rem` threshold the inner `.cards` grid switches from `1fr` to `minmax(0, 30rem)`.
- `EmployeeTable.module.scss` — `.root` becomes the query container; the `.searchContainer` is `width: 100%` by default and caps at `20rem` above the same threshold.

Both wrappers subscribe to the same `40rem` breakpoint Tabs uses (`BREAKPOINTS.SMALL` in `src/shared/constants.ts`) without depending on Tabs' DOM. Any partner-provided `Tabs` adapter works unchanged.

## Changes

- `Tabs.tsx` — reverted `data-tabs-mode` attribute.
- `TimeOffPolicyDetail.module.scss` — container-query rules; removed `.descriptionCard` / `.descriptionCardUnlimited`.
- `TimeOffPolicyDetailPresentation.tsx` — `.cardsContainer` > `.cards` wrapper around both `Box` components; removed unused `Flex` import.
- `EmployeeTable.module.scss` — `container-type: inline-size` on `.root`; `.searchContainer` cap moved inside an `@container (min-width: 40rem)` rule.

## Related

- Jira: [SDK-878](https://gustohq.atlassian.net/browse/SDK-878), [SDK-880](https://gustohq.atlassian.net/browse/SDK-880)

## Testing

- `npm run sdk-app` and open the TimeOff policy detail (`TimeOffFlow` → policy detail).
- Resize the main content area through the 640px / 40rem threshold:
  - **Above the breakpoint** (Details tab): both cards share a 30rem track, left-aligned.
  - **Below the breakpoint** (Details tab): tab strip collapses to a Select; both cards span the full container, matching the Select's edges.
  - **Above the breakpoint** (Employees tab): search bar capped at 20rem.
  - **Below the breakpoint** (Employees tab): search bar spans the full container, matching the Select's edges.
- Optional partner-adapter check: render with a custom `Tabs` adapter in `sdk-app` and confirm the cards / search bar still resize correctly (they should — the queries no longer depend on Tabs' DOM).
- `npm run test -- --run src/components/TimeOff`
- `npm run test -- --run src/components/Common/UI/Tabs`

[SDK-878]: https://gustohq.atlassian.net/browse/SDK-878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Common/UI/Tabs/Tabs.tsx
+++ b/src/components/Common/UI/Tabs/Tabs.tsx
@@ -27,7 +27,11 @@ export function Tabs({ tabs, selectedId, onSelectionChange, className, ...ariaPr
   }))
 
   return (
-    <div ref={containerRef} className={classNames(styles.root, className)}>
+    <div
+      ref={containerRef}
+      className={classNames(styles.root, className)}
+      data-tabs-mode={shouldUseDropdown ? 'dropdown' : 'tabs'}
+    >
       {!shouldUseDropdown ? (
         <AriaTabs
           className={styles.tabsContainer}
--- a/src/components/Common/UI/Tabs/Tabs.tsx
+++ b/src/components/Common/UI/Tabs/Tabs.tsx
@@ -27,11 +27,7 @@ export function Tabs({ tabs, selectedId, onSelectionChange, className, ...ariaPr
   }))
 
   return (
-    <div
-      ref={containerRef}
-      className={classNames(styles.root, className)}
-      data-tabs-mode={shouldUseDropdown ? 'dropdown' : 'tabs'}
-    >
+    <div ref={containerRef} className={classNames(styles.root, className)}>
       {!shouldUseDropdown ? (
         <AriaTabs
           className={styles.tabsContainer}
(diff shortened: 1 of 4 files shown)


krisxcrash (approved):

serikjensen (approved):

serikjensen (changes requested):
We're going to have to figure out a diff approach here! this one won't work with adapters

serikjensen (approved):
Looks good!

PR #1829, opened 2026-05-14
feat(e2e): scenario schema + driven E2E infrastructure

jeffredodd:
> **Consolidation note:** This PR absorbs the work originally drafted as PR #1827 (`feat(e2e): scenario JSON schema + fragments + validator`). The schema commit (`7cc95d16`) is rebased into this branch and every file from #1827 is present here with identical content. PR #1827 is being closed as superseded so reviewers have a single review unit for all foundational scenario work (schema + fragments + validator + loader + hash + cache + runner + fixture + reporter + scripts). No code was lost.

## Summary

Complete infrastructure for scenario-driven E2E testing, enabling per-domain test isolation and the API upgrade workflow.

- **Scenario schema + loader + hash** (commits from prior branches): JSON schema for scenario definitions, `$ref` resolution with cycle detection, template substitution (`{{ts}}`, `{{relative:+Nd}}`), structural hashing for cache identity
- **Cache layer** (`e2e/scenario/cache.ts`): Atomic file-based cache with token validation probe. Reuses provisioned companies across runs when scenario structure hasn't changed.
- **Runner** (`e2e/scenario/runner.ts`): Orchestrates demo provisioning via `createDemoAndProvision`, then decorates entities (locations, employees with addresses/jobs/compensation/onboarding, contractors, pay schedules, payroll processing including prepare/calculate/submit/poll). Validates `expectedContext` paths post-decoration.
- **Fixture** (`e2e/utils/localTestFixture.ts`): `scenario` fixture reads annotation, provisions via runner, auto-tags tests with `@<domain>` for `--grep` filtering. Backwards-compatible — tests without scenario annotations use the legacy `localConfig` path.
- **Reporter** (`e2e/reporters/scenario-reporter.ts`): Custom Playwright reporter emitting per-domain/scenario pass/fail/skip aggregation to `e2e/reports/results.json`. Registered in all 3 Playwright configs.
- **Scripts**: `npm run e2e:scenarios:prewarm [--domain X]` and `npm run e2e:scenarios:clear`
- **CI**: Uploads `e2e/reports/` artifact alongside `playwright-report/`

Notion tasks: #007-046 (Phase A — Groundwork)

## Test plan

- [x] `npx vitest run --config vitest.scenario.config.ts` — all 32 scenario module tests pass
- [x] `npx tsc --noEmit` — clean typecheck
- [x] `npm run build` — clean build
- [ ] CI checks pass
[...]

PR #1832, opened 2026-05-14
feat(e2e): employee domain — scenarios + spec rewrites

jeffredodd:
## Summary
- Add employee-onboarding and employee-self-onboarding scenarios
- Rewrite employee spec files to use scenario fixture
- Move to e2e/tests/employee/ directory

## 
https://github.com/user-attachments/assets/f0b734d4-a75a-4768-af0b-2fbc0c9d494e.webm

https://github.com/user-attachments/assets/71a799a4-16f0-4ef0-a205-4353f2ea1e67.webm

https://github.com/user-attachments/assets/213bfb51-235d-4fe4-8b2d-310ea40ea8a0.webm

Base: PR #1829 (e2e/infrastructure)

Notion tasks: #068-073


jeffredodd:
@cursoragent can you resolve conflicts here and make sure we're rebased off main?

serikjensen:
Claude recs:

**1. Incomplete migration in employee-onboarding.spec.ts**

The file was moved and given a `beforeEach` scenario annotation, but the tests still navigate with `companyId=123` and don't consume the `scenario` fixture. Every other new spec in this PR drops the hardcoded ID. This will fail in CI's domain-matrix mode where `companyId=123` isn't a real company.

```ts
// Fix: convert to scenario pattern, matching the new spec files:
test('completes the happy path successfully', async ({ page, scenario }) => {
  test.skip(!scenario.flowToken, 'Requires scenario provisioning')
  await page.goto('/?flow=employee-onboarding')
  // ... rest of test
})
```

**2. Unsafe type cast in pickTerminationCandidateId (employeeFlowDrivers.ts:998)**

`r.json()` is cast directly with `as EmployeeSummary[]` on a live API response. If the endpoint returns an unexpected shape or a non-array (e.g., a paginated `{ employees: [...] }` wrapper), the `.find()` call silently returns `undefined` on every attempt, the polling loop exhausts, and termination is skipped with a misleading error.

```ts
// Fix: guard the array assumption at runtime
const data: unknown = r.ok ? await r.json() : []
const employees: EmployeeSummary[] = Array.isArray(data) ? (data as EmployeeSummary[]) : []
```

**3. Silent pass in employee-self-onboarding.spec.ts (lines 609–623)**

Same pattern flagged in #1833: the test returns early with a green result if any "completed" text appears after a single Continue click, and the fallback `heading | Continue button` assertion would pass on virtually any loaded page. The full multi-step flow is never verified.

```ts
// Fix: remove the early-return branch; assert the specific completion heading:
await expect(
  page.getByRole('heading', { name: /completed|that's it/i }),
).toBeVisible({ timeout: 60_000 })
```

---

serikjensen (approved):

jeffredodd:
@cursoragent Can you investigate the feedback and see if we should address Steve's cursor feedback?

jeffredodd:
@cursoragent can you resolve the conflicts

PR #1844, opened 2026-05-15
refactor(SDK-823): organize employee documents by journey

mariechatfield:
## Summary

_Without renaming any files or changing exports_ move the components and hooks used for employee documents and forms into a top-level `Employee/Documents` directory, split into `onboarding/` and `shared/`.

In the next PR, we will add a `management/` directory with a new component that allows onboarded employees to view and sign documents from the dashboard, using the `shared/` components.

## Changes

**Before**                             
```                                                                           
├──DocumentSigner/                                                                             
│  ├── DocumentList/                                                                                                 
│  ├── EmploymentEligibility/                                                                                      
│  ├── I9SignatureForm/                                                                                              
│  ├── SignatureForm/                                                                          
│  ├── shared/                         
│  │   └── useSignEmployeeForm/                                                                                    
│  └── DocumentSigner.tsx
└─EmployeeDocuments/            
```                                                              
                                      
  **After**         
```                                                                                              
└──Documents/
   ├── onboarding/
   │   ├── DocumentSigner/
   │   │   ├── DocumentList/                                                                                         
   │   │   ├── EmploymentEligibility/
   │   │   ├── I9SignatureForm/                                                                                      
   │   │   └── DocumentSigner.tsx                                                              
   │   └── EmployeeDocuments/ 
   └── shared/                                                                                                       
       ├── SignatureForm/                                                                      
       └── useSignEmployeeForm/ 
```

## Related

- Jira ticket: [SDK-823](https://gustohq.atlassian.net/browse/SDK-823)

[SDK-823]: https://gustohq.atlassian.net/browse/SDK-823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1846, opened 2026-05-15
docs(SDK-897): harden endpoint inventory derivation script

mariechatfield:
## Summary

We auto-generate `docs/reference/endpoint-inventory.json` and `docs/reference/endpoint-reference.md` by walking all the components exported by the SDK and determining which embedded API endpoints they call.

Harden this script so that it emits more stable and reliable results.

## Changes

Easiest to look at the diffs for the generated code commit by commit.

- Sort blocks and their endpoints (alphabetically by path, then GET → POST → PUT → PATCH → DELETE)         
- Ensure component names match SDK usage
  - Fix flow names to use the correct top-level namespace (`Contractor.PaymentFlow` instead of `Contractor.Payments.PaymentFlow`)
  - Use the public export alias when a component is re-exported under a different name (`TimeOff.PolicySettings` instead of `TimeOff.PolicySettingsPresentation`)      
- Discover previously missing flows that were nested more than three directories deep (`Employee.DashboardFlow`, `Payroll.DismissalFlow`, etc.)
- Ensure flows are not duplicated in the blocks section (`TimeOff.TimeOffFlow`)

## Related

- Jira ticket: [SDK-897](https://gustohq.atlassian.net/browse/SDK-897)
- Figma: [Design link]
- Related PRs: #1844 first exposed the issue when moving directories accidentally deleted endpoint docs

[SDK-897]: https://gustohq.atlassian.net/browse/SDK-897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build/deriveEndpointInventory.ts
+++ b/build/deriveEndpointInventory.ts
@@ -1,12 +1,22 @@
-import { readFileSync, readdirSync, writeFileSync, mkdirSync, statSync, existsSync } from 'fs'
+import {
+  readFileSync,
+  readdirSync,
+  writeFileSync,
+  mkdirSync,
+  statSync,
+  existsSync,
+  rmSync,
+} from 'fs'
 import { join, dirname, relative, resolve } from 'path'
 import { fileURLToPath } from 'url'
-import { Project, SyntaxKind } from 'ts-morph'
+import { spawnSync } from 'child_process'
+import { Project, SourceFile, SyntaxKind } from 'ts-morph'
 
 const __filename = fileURLToPath(import.meta.url)
 const __dirname = dirname(__filename)
 
 const ROOT = join(__dirname, '..')
+const SRC_DIR = join(ROOT, 'src')
 const FUNCS_DIR = join(ROOT, 'node_modules/@gusto/embedded-api/src/funcs')
 const OPS_DIR = join(ROOT, 'node_modules/@gusto/embedded-api/src/models/operations')
 const COMPONENTS_DIR = join(ROOT, 'src/components')
@@ -14,8 +24,6 @@ const JSON_OUTPUT_PATH = join(ROOT, 'docs/reference/endpoint-inventory.json')
 const MD_OUTPUT_PATH = join(ROOT, 'docs/reference/endpoint-reference.md')
 
 const isVerifyMode = process.argv.includes('--verify')
-
-// Removed NON_DOMAIN_DIRS - no longer needed with export-based discovery
 
 interface Endpoint {
   method: string
@@ -29,11 +37,27 @@ interface BlockEntry {
 
 interface FlowEntry {
   blocks: string[]
-  endpoints: Endpoint[]
-  variables: string[]
 }
 
-// --- AST-based extraction from @gusto/embedded-api ---
+interface BlockMapping {
+  blockName: string
+  componentDir: string
+}
+
+interface FlowMapping {
+  flowName: string
+  flowDir: string
+}
+
+interface Inventory {
+  blocks: Record<string, BlockEntry>
+  flows: Record<string, FlowEntry>
+}
+
+interface DerivationResult {
+  inventory: Inventory
+  funcLookup: Map<string, Endpoint>
+}
 
 function createApiProject(): Project {
   return new Project({
@@ -94,8 +118,6 @@ function buildParamNameMap(project: Project, funcPaths: string[]): Record<string
   return paramNameMap
 }
 
-// --- Build a lookup from func name -> { method, path } ---
-
 function collectRawFuncPaths(
   project: Project,
 ): { funcName: string; path: string; method: string }[] {
@@ -136,8 +158,7 @@ function collectRawFuncPaths(
   return results
 }
 
-function buildFuncLookup(): Map<string, Endpoint> {
-  const project = createApiProject()
+function buildFuncLookup(project: Project): Map<string, Endpoint> {
   const rawFuncs = collectRawFuncPaths(project)
   const paramNameMap = buildParamNameMap(
     project,
@@ -154,8 +175,6 @@ function buildFuncLookup(): Map<string, Endpoint> {
 
   return lookup
 }
-
-// --- Recursively find source files in a directory ---
 
 function walkDir(dir: string): string[] {
   const results: string[] = []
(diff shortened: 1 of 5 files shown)


build/tsconfig.json
mariechatfield, line 1: Make sure the build scripts have valid types by indicating this directory runs in the context of node.

lint-staged.config.js
mariechatfield, line 1: This script is more thorough _but_ more expensive. Take it out of the pre-commit hook, which means you _may_ end up getting a failed build on your branch if you don't run it yourself

build/deriveEndpointInventory.ts
```diff
@@ -175,35 +194,67 @@ function walkDir(dir: string): string[] {
...
 
-// --- Extract @gusto/embedded-api imports from component files ---
-
-function extractApiImports(filePaths: string[]): Set<string> {
+function collectTransitiveApiImports(
+  project: Project,
+  entryFilePaths: string[],
+  componentDir: string,
+  otherBlockDirs: Set<string>,
+): Set<string> {
```
mariechatfield, line 202: This is where the bulk of the improvement happens. We're now using a ts-morph project to walk the imports instead of using regex-based patterns to look for imports that match a text pattern.

This is what allows us to match components and endpoints at any depth while avoid over-attributing endpoints to components that are siblings or that reference each other (like a flow importing blocks).

build/deriveEndpointInventory.ts
```diff
@@ -244,22 +288,23 @@ function parseNamespaceExports(): Map<string, string> {
...
-        const importPath = match[2]
+      // export { EmployeeDocuments as Documents } from './Documents'  →
+      //   nameToken = "EmployeeDocuments as Documents"
+      //   importPath = "./Documents"
+      // Uses the alias when present so the map key matches the public export name.
+      const exportPattern =
+        /export\s+\{\s*(\w+(?:\s+as\s+\w+)?)[^}]*?\}\s+from\s+['"](\.[^'"]+)['"]/g
+      for (const [_fullMatch, nameToken, importPath] of content.matchAll(exportPattern)) {
+        const asIndex = nameToken.indexOf(' as ')
+        const componentName = asIndex >= 0 ? nameToken.slice(asIndex + 4).trim() : nameToken.trim()
```
mariechatfield, line 304: Gets the component name from the alias, if present, so that we document the name the library consumer will actually use

build/deriveEndpointInventory.ts
```diff
@@ -302,85 +333,78 @@ function discoverBlocks(): BlockMapping[] {
...
 }
 
-function deriveFlowBlocks(
-  flowDir: string,
-  blockDirToName: Map<string, string>,
-  blockMappings: BlockMapping[],
+function deriveBlocksFromImport(
+  resolvedImportPath: string,
+  rawImportedNames: string,
+  dirToBlocks: Map<string, string[]>,
```
mariechatfield, line 368: Shared function that replaces the shallow / static lookup:

```ts
const segments = importPath.split('/')
const candidateDirs = [
  join(COMPONENTS_DIR, segments[0], segments[1] ?? ''),
  join(COMPONENTS_DIR, segments[0], segments[1] ?? '', segments[2] ?? ''),
]

// iterate candidateDirs:
//   if `dirToBlocks` has dir as a key:
//     then add block names to set and break on first match
```

Instead, we use the existing `dirToBlocks` mapping to identify which directory is a best match for the import path, and extract the block names from there.

build/deriveEndpointInventory.ts
```diff
@@ -389,126 +413,92 @@ function deriveFlowBlocks(
...
+      const key = `${ep.method} ${ep.path}`
+      if (seen.has(key)) return false
+      seen.add(key)
+      return true
+    })
+    .sort((a, b) => {
+      const pathOrder = a.path.localeCompare(b.path)
+      if (pathOrder !== 0) return pathOrder
+      return (METHOD_RANK[a.method] ?? 99) - (METHOD_RANK[b.method] ?? 99)
+    })
```
mariechatfield, line 488: Makes the outputs more consistent (easier diffs to read long term) _and_ makes the tables nicer for a human since sort the endpoints by name and then by method.

Image
Image

build/deriveEndpointInventory.ts
```diff
@@ -678,7 +642,21 @@ function generate() {
   }
 }
 
-// --- Verify mode: compare against committed files, exit 1 if stale ---
+function printDiff(committedPath: string, freshContent: string): void {
```
mariechatfield, line 645: Since this may fail in CI instead of in your pre-commit hook, make sure we include the diff in the build output for a slightly nicer dev experience

jeffredodd (approved):
LGTM

One thing to note, I noticed we don't have 'hooks' in our inventory, we might want to write a JIRA ticket so we make sure to come back and do that work.

mariechatfield:
Ooh good point. I'm also cooking up an `api-contract.yaml` script generator that should help us keep track of when we're making breaking changes to exports. It looks a bit like this:

```yaml
# This file is auto-generated by build/deriveApiContract.ts.
# Do not edit manually — run "npm run api-contract:derive" to regenerate.
#
# This file describes the public SDK surface: exported components, hooks, and their types.
# Changes to this file may indicate breaking changes to the API contract with integrators.
# Additions are generally safe; removals or renames are breaking.

exports:
  ACCOUNT_TYPES:
    category: Constant
Company:
    category: Component.Namespace
    comment: "@deprecated Use `CompanyOnboarding` (or future `CompanyManagement`) instead. Will be removed in a future major version."
    children:
      - AssignSignatory
      - BankAccount
      - CreateSignatory
      - DocumentList
      - DocumentSigner
      - FederalTaxes
      - Industry
      - InviteSignatory
      - LocationForm
      - Locations
      - OnboardingFlow
      - OnboardingOverview
      - PaySchedule
      - SignatureForm
      - StateTaxes
      - StateTaxesForm
      - StateTaxesList
  useHomeAddressForm:
    category: Function.Hook
```

Obviously wouldn't include the actual endpoints being used by the hooks but might help us think about the surface area of the API especially as we get closer to a 1.0.0 eventually

https://gustohq.atlassian.net/browse/SDK-898 to track "making sure hooks are included in the endpoint inventory"

PR #1852, opened 2026-05-15
feat(SDK-823): add DocumentManager block and wire into DashboardFlow

mariechatfield:
## Summary

Adds `EmployeeManagement.DocumentManager` — shows a view-only PDF or delegates to `SignatureForm` when the form requires signing. Integrates into `DashboardFlow` via a new `documentManager` state triggered by `EMPLOYEE_VIEW_FORM_TO_SIGN`.

## Demo

https://github.com/user-attachments/assets/87ada630-c219-498d-b06b-b6f660634fc6

Screenshot 2026-05-15 at 3 08 10 PM
Screenshot 2026-05-15 at 3 08 14 PM

## Related

- Jira ticket: [SDK-823](https://gustohq.atlassian.net/browse/SDK-823)
- Related PRs: Follow up to #1844 

## Testing

1. `npm run sdk-app`
2. Go to `Employee > OnboardingFlow` and finish onboarding the _last two employees_
  a. Standard company should have two employees in pending state, one will just need you to click through and save. Another will need address + SSN + tax information
3. Switch to `Employee > DashboardFlow` and use the settings panel to switch to one of your newly onboarded employees
4. Go to `Documents` tab. You should see at least one document per newly onboarded employee to sign
  - [ ] You can view documents that have already been signed
  - [ ] You can download documents that have already been signed
  - [ ] You can sign documents that need a signature
  - [ ] Signing a document keeps you on the same page but shows the signed version for download
  - [ ] Back button goes back to the main Dashboard flow (first tab, not Documents)

No stories added for this component, since it requires fetching data.
This component does not show up in the sdk-app sidebar because it is exported from `EmployeeManagement` only, not from the deprecated `Employee` domain.

[SDK-823]: https://gustohq.atlassian.net/browse/SDK-823?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Employee/Documents/management/DocumentManager.tsx
```diff
@@ -0,0 +1,94 @@
...
+export function DocumentManager({
+  FallbackComponent,
+  ...props
+}: DocumentManagerProps & { FallbackComponent?: BaseBoundariesProps['FallbackComponent'] }) {
+  return (
+    <BaseBoundaries componentName="Employee.DocumentManager" FallbackComponent={FallbackComponent}>
+      <DocumentManagerRoot {...props} />
+    </BaseBoundaries>
+  )
+}
```
serikjensen: So i think claude is getting confused between two patterns emerging in the codebase. We are currently doing the sdk hook migration and ironing out how that looks.

components that are powered by SDK hooks
* Do not use suspense queries
* Use `composeErrorHandler` to aggregate the error response and pass it to BaseLayout
* Compose isLoading returned from the used hooks to manually control the loading state

Components that are not powered by SDK hooks
* Use suspense queries
* Use BaseComponent as the top level wrapper for the component
* Get error handling for free through BaseComponent
* Get loading state for free through suspense

This is kinda doing a mix of the two, we should probably push it to be one or the other. The component itself will not be powered by SDK hooks from what i'm seeing so it's worth figuring out long term how we want to handle such components in the codebase.

In the meantime, we either need to

* Update this component to use `composeErrorHandler`, pass the hook results into composeErrorHandler to aggregate the errors, and then ensure the errors are passed into BaseLayout which will allow for proper error handling

_or_

* Update this to be more aligned with other components using BaseComponent. Update the DocumentManager to use BaseComponent instead of BaseBoundaries which will have it do the suspense and error handling automatically
* Update to use suspense queries from the embedded api instead

I would probably lean towards the latter here

mariechatfield: Thanks for all of this context and explanation! I think I follow the distinction, especially the importance of using the `BaseComponent` and suspense so that we don't have to manually handle loading state everywhere that uses async queries.

I _think_ the latest commit implements your recommended strategy but if that's still not quite right it might be easier to pair if you have time

src/components/Employee/Dashboard/DocumentsView.tsx
```diff
@@ -40,8 +40,18 @@ export function DocumentsView({ forms = [], isLoading = false, onViewForm }: Doc
...
+            <Components.Badge status="warning">
+              {t('documents.signingStatus.notSigned')}
+            </Components.Badge>
+          )
+        }
+        return (
+          <Components.Badge status="success">
+            {t('documents.signingStatus.signed')}
+          </Components.Badge>
+        )
```
mariechatfield, line 54: Image

Based on @aaronlee777's suggestion to use a badge in this view

serikjensen (approved):
Looks great! thanks for updating!

PR #1853, opened 2026-05-15
docs(SDK-897): include journey namespaces in endpoint inventory

mariechatfield:
## Summary

The endpoint inventory script was omitting all the journey-based namespaces (`EmployeeOnboarding` and `EmployeeManagement`) because their barrel files didn't match `domain/index.ts` but instead are `domain.ts`

## Changes

- Include `domain.ts` style barrel files
- Include deprecation status and calculation in the endpoint inventory script
  - This is used to de-duplicate and choose the best namespace for the component, _not_ to surface deprecation warnings in the inventory output (yet)
- **Behavioral change for existing flows**: `Employee.DashboardFlow`, `Employee.OnboardingFlow`, and `Employee.SelfOnboardingFlow` now list some blocks under `EmployeeOnboarding.*` and `EmployeeManagement.*` instead of `Employee.*` — this is intentional and correct; the de-duplication logic prefers the non-deprecated namespace when the same component dir appears in both

## Testing

- Ran `npm run endpoints:derive` and confirmed `EmployeeOnboarding` and `EmployeeManagement` blocks now appear in the output
- Ran `npm run endpoints:verify` to confirm the generated files match the committed output
- Spot-checked de-duplication: blocks shared between deprecated and non-deprecated namespaces (e.g. `EmployeeDocuments`) resolve to the preferred non-deprecated entry

## Related

- Jira ticket: [SDK-897](https://gustohq.atlassian.net/browse/SDK-897)

[SDK-897]: https://gustohq.atlassian.net/browse/SDK-897?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/reference/endpoint-inventory.json
+++ b/docs/reference/endpoint-inventory.json
@@ -1648,6 +1648,866 @@
         "companyId",
         "timeOffPolicyUuid"
       ]
+    },
+    "EmployeeOnboarding.EmployeeList": {
+      "endpoints": [
+        {
+          "method": "GET",
+          "path": "/v1/companies/:companyId/employees"
+        },
+        {
+          "method": "DELETE",
+          "path": "/v1/employees/:employeeId"
+        },
+        {
+          "method": "PUT",
+          "path": "/v1/employees/:employeeId/onboarding_status"
+        }
+      ],
+      "variables": [
+        "companyId",
+        "employeeId"
+      ]
+    },
+    "EmployeeOnboarding.OnboardingSummary": {
+      "endpoints": [
+        {
+          "method": "GET",
+          "path": "/v1/employees/:employeeId"
+        },
+        {
+          "method": "GET",
+          "path": "/v1/employees/:employeeId/onboarding_status"
+        }
+      ],
+      "variables": [
+        "employeeId"
+      ]
+    },
+    "EmployeeOnboarding.Landing": {
+      "endpoints": [
+        {
+          "method": "GET",
+          "path": "/v1/companies/:companyId"
+        },
+        {
+          "method": "GET",
+          "path": "/v1/employees/:employeeId"
+        }
+      ],
+      "variables": [
+        "companyId",
+        "employeeId"
+      ]
+    },
+    "EmployeeOnboarding.DocumentSigner": {
+      "endpoints": [
+        {
+          "method": "GET",
+          "path": "/v1/employees/:employeeId"
+        },
+        {
+          "method": "GET",
+          "path": "/v1/employees/:employeeId/forms"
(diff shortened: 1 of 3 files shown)


docs/reference/endpoint-inventory.json
```diff
@@ -1661,7 +2521,20 @@
         "Company.OnboardingOverview",
         "Company.PaySchedule",
         "Company.StateTaxes",
-        "Employee.OnboardingFlow"
+        "EmployeeOnboarding.OnboardingFlow"
+      ]
```
mariechatfield, line 2525: This is the reason for the deprecation math: we should try to guide library users into the `EmployeeOnboarding` version of the onboarding flow instead of the `Employee.OnboardingFlow`

docs/reference/endpoint-inventory.json
```diff
@@ -1725,6 +2610,49 @@
...
+      "blocks": [
+        "Employee.HomeAddress",
+        "Employee.PaymentMethod",
+        "EmployeeManagement.FederalTaxes",
+        "EmployeeManagement.Profile",
+        "EmployeeManagement.StateTaxes",
+        "EmployeeManagement.WorkAddress",
+        "EmployeeOnboarding.PaymentMethod"
+      ]
+    },
```
mariechatfield, line 2623: But even in the new namespaces, some blocks are only exported from the deprecated versions. So we can't just find-replace the old namespaces indiscriminately

docs/reference/endpoint-inventory.json
```diff
@@ -1725,6 +2610,49 @@
...
+        "EmployeeManagement.TerminateEmployee",
+        "EmployeeManagement.TerminationSummary",
+        "Payroll.DismissalFlow",
+        "Payroll.PayrollLanding"
+      ]
+    },
+    "EmployeeOnboarding.OnboardingFlow": {
+      "blocks": [
+        "Employee.PaymentMethod",
+        "EmployeeManagement.EmployeeDocuments",
```
mariechatfield: This kills me, but it is accurate. We should be exporting `EmployeeDocuments` from onboarding and _not_ from management, since it's only used in onboarding. Right now it's swapped

mariechatfield: whoops, just merged main where I had a commit that marked the export from `EmployeeManagement` as deprecated 😂 and now this line has been updated to `Employee.EmployeeDocuments`, as intended

serikjensen (approved):

PR #1821, opened 2026-05-14
feat: payment method sub-flows in Dashboard flow

dmortal:
## Summary

Integrates payment method sub-flows (add bank account, split paycheck, delete bank account) into the Dashboard flow via the robot3 state machine. Replaces the old monolithic `PaymentMethod.tsx` with separate `management/` and `onboarding/` variants backed by shared hooks.

## Changes

- Refactored `PaymentMethod` into `management/` (standalone flow) and `onboarding/` (Dashboard-embedded) variants with shared hooks under `shared/`
- Wired `EMPLOYEE_BANK_ACCOUNT_CREATE` and `EMPLOYEE_SPLIT_PAYCHECK` events into `dashboardStateMachine` — both return directly to `index` with a dismissable success alert held in FSM context
- Added `EMPLOYEE_BANK_ACCOUNT_DELETED` and `EMPLOYEE_DISMISS` self-transitions on `index` state to show/clear the delete success alert without leaving the view
- Replaced payment data fetching in `useEmployeeCompensation` with `usePaymentMethodList` called directly in `JobAndPayView`
- Fixed bank account empty state in `JobAndPayView`: shows payment method label/value instead of an empty DataView
- Fixed delete dialog artifact (dialog + alert briefly co-visible) by closing the dialog optimistically before the async delete call
- Fixed `BankForm` submit doing nothing: added `accountType: 'Checking'` default to `useBankForm` (Zod schema requires it once `hasBankPayload` is set)

## Testing

```bash
npm run test -- --run   # 205 tests, 2610 assertions — all pass
npm run lint            # warnings only, no errors in changed files
npm run format          # no changes after running
```

Manual (Playwright, `http://localhost:5200/employee/DashboardFlow`):
- Add bank account → returns to Dashboard index with "Bank account successfully added." alert
- Delete bank account → dialog closes, returns to Dashboard with "Bank account successfully deleted." alert (no co-render artifact)
- Split paycheck → returns to Dashboard index with "Split payment successfully updated." alert
- All three alerts are dismissable and clear on next FSM transition
- Payment section with no bank accounts shows "Payment method" / "Check" or "Direct deposit" label

PR #1859, opened 2026-05-18
refactor: inline bank-account delete UI state into PaymentMethod views

serikjensen:
## Summary

Removes the `useDeleteBankAccount` helper hook and inlines its UI state into the three consumer views, aligning `usePaymentMethodList` with the established `useEmployeeList` / `useHomeAddressManagement` action pattern. Follow-up cleanup on the hook layer introduced in #1821.

After this change `usePaymentMethodList.actions.onDelete(uuid)` is the single delete action surface — the same shape as `useEmployeeList.actions.onDelete(employeeId)` — and the hook owns everything API-shaped (data, in-flight status via `deletePendingBankAccountUuid`, error handling). The views own only the pure UI affordances they each need:

- `management/ListView.tsx` and `onboarding/ListView.tsx` track `deleteConfirmUuid` (which row's confirm modal is open) and `deletedAccountNumber` (recent-delete success Alert) locally.
- `Dashboard/JobAndPayView.tsx` tracks only `deleteConfirmUuid` — it doesn't render a success Alert; the existing `EMPLOYEE_BANK_ACCOUNT_DELETED` event handles the dashboard-level signal.

This matches the pattern already used by `useHomeAddressManagement` + `HomeAddressView` for confirm-then-delete UX:

- Hook exposes `actions.<delete>(uuid)` + an in-flight status field.
- View owns a single `useState<string | null>(deleteConfirmUuid)` and derives the row for display via `bankAccounts.find(a => a.uuid === deleteConfirmUuid)`.

### Incidental UX bug fixes

Two small bugs in the old helper fall out of inlining:

- The confirm modal previously closed before awaiting the delete request, so a failed delete left the user with no modal and a silent error. The modal now stays open on failure (resets only inside `if (result)`), matching `HomeAddressView`.
- The success Alert previously set `deletedAccountNumber` unconditionally even when `baseSubmitHandler` swallowed an API error. The Alert now only shows after a confirmed-successful delete.

### Type-only tweak

Loosened `DeleteBankAccountDialog`'s `hiddenAccountNumber` prop from `string | undefined` to `string?` so the dialog can accept an `EmployeeBankAccount` (where the field is optional) directly without a structural-typing mismatch. Runtime behavior is unchanged.

## Test plan

- [x] `npm run test -- --run src/components/Employee/PaymentMethod src/components/Employee/Dashboard` (8 files, 66 tests, all pass)
- [x] Lint clean on all touched files
- [ ] Manual smoke: delete a bank account from each of the three surfaces (employee-management list, onboarding list, dashboard Job & Pay tab) and confirm the modal closes on success, the success Alert appears on the two ListViews, and the modal stays open on a forced API failure.

### Demo

https://github.com/user-attachments/assets/22c76150-eedb-46b7-a74d-4229f8c25275


serikjensen:
Realized the delete hook was just an internal helper, and the partner api was already where it needed to be for the payment list hook. Closing this one

PR #1861, opened 2026-05-18
feat(compensation): add management/EditCompensation for steady-state edits

serikjensen:
> Second PR in the steady-state compensation series. Stacked on #1860 (now merged).

## Summary

- Adds `management/EditCompensation` — a dedicated steady-state edit surface composing `useJobForm` and `useCompensationForm` via `composeSubmitHandler`. PUT job (immediate: 2% shareholder / WA WC) then PUT compensation (effective-dated: title, rate, unit, FLSA, effective date, min-wage adjustment).
- Adds `withTitleField?: boolean` to `useJobForm` (mirrors `withHireDateField`): when `false`, drops `title` from schema validation, the rendered field, and the PUT/POST body. Onboarding stays on the default (`true`) — no behavior change.
- `useCompensationForm.title` now falls back to `currentJob.title` when the loaded comp row has no scheduled title change, so the input seeds from the employee's current title for steady-state edits.

## Why this shape

Follows the "Steady-state edit" pattern documented in [`docs/hooks/jobs-and-compensations.md`](docs/hooks/jobs-and-compensations.md): title is effective-dated and rides the comp row; 2% shareholder / WC are immediate and ride the job. The hook-level `withTitleField` toggle keeps onboarding's title-on-job behavior intact while letting management drive title through the comp form.

## Out of scope here (covered by later PRs in the series)

- `management/ListView` + pending-change banner + delete dialogs (PR 3).
- `management/Compensation` flow root + state machine + `management/AddJob` (PR 4).
- Dashboard wiring (PR 5).

## Notes

- **No Storybook stories** in this PR. The Compensation directory has no existing story coverage and there's no MSW-based story pattern established in the repo. Unit tests cover every variant the plan listed (S-Corp, WA, carve-out, prefill, body shape, submit ordering). Happy to add stories if reviewers want them.
- **Upstream contract verification still pending** — gws-flows' single-branch behavior silently drops job-only fields on the future-comp path. The SDK intentionally diverges by issuing the two calls separately. Tracked outside this PR.

## Test plan

- [x] \`npm run test -- --run src/components/Employee/Compensation\` — 106 tests pass (97 baseline + 3 \`withTitleField\` + 6 \`management/EditCompensation\`).
- [x] \`npx tsc --noEmit\` — clean.
- [x] \`npm run lint:check\` — clean (only pre-existing warnings).


src/components/Employee/Compensation/shared/useJobForm/useJobForm.tsx
```diff
@@ -69,10 +69,19 @@ export interface UseJobFormProps {
    * Defaults to `true`.
    */
   withHireDateField?: boolean
+  /**
+   * When `false`, hides `Fields.Title` (becomes `undefined`), removes
+   * `title` from schema validation, and skips sending `title` on PUT/POST.
+   * Use this for steady-state compensation edits where title is rendered
+   * via `CompFields.Title` and rides the future-dated compensation row
```
mariechatfield: `rides` the future-dated compensation row? Was that meant to be something like writes or overrides? Or am I the one hallucinating language instead of claude

src/components/Employee/Compensation/shared/useJobForm/useJobForm.tsx
```diff
@@ -254,11 +270,16 @@ export function useJobForm({
                     : 'hireDate is required to create a job. Pass it via JobSubmitOptions when withHireDateField is false.',
                 )
               }
+              if (titleToSend === undefined) {
+                throw new SDKInternalError(
+                  'title is required to create a job. Set withTitleField: true (the default) on useJobForm for create flows.',
+                )
+              }
```
mariechatfield, line 278: Nice, love to see defensive guarding like this

src/i18n/en/Employee.Compensation.json
```diff
@@ -43,6 +43,23 @@
...
+    "wageLabel": "Wage",
+    "wageFrequencyLabel": "Wage frequency",
+    "wageFrequencyOptions": {
+      "Hour": "Hourly",
+      "Week": "Weekly",
+      "Month": "Monthly",
+      "Year": "Annually",
+      "Paycheck": "Per paycheck"
+    },
+    "effectiveDateWarning": "Changing an employee's start date may impact when they get paid, their time off balance and accruals, and possibly create the need for an additional off-cycle payroll.",
```
mariechatfield: Do we use smart quotes for customer-facing copy?

src/components/Employee/Compensation/shared/useJobForm/useJobForm.tsx
serikjensen: haha claude is getting insufferable with copy

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.test.tsx
```diff
@@ -0,0 +1,347 @@
...
+
+    await screen.findByRole('heading', { name: 'Edit compensation' })
+
+    // Only the compensation-form title field is rendered, never the job-form one.
+    // useJobForm with withTitleField:false produces undefined for Fields.Title,
+    // so just one labelled input should exist.
+    expect(screen.getAllByLabelText('Job title')).toHaveLength(1)
+  })
+
+  it('submits PUT /v1/jobs/:id before PUT /v1/compensations/:id', async () => {
```
mariechatfield: Do we only care that we PUT compensations after jobs, or do we need to assert that compensations is only called if the PUT to jobs is successful / after it resolves?

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
```diff
@@ -0,0 +1,290 @@
...
+    jobId,
+    compensationId,
+    withEffectiveDateField: true,
+    optionalFieldsToRequire: {
+      update: ['title', 'flsaStatus', 'rate', 'paymentUnit', 'effectiveDate'],
+    },
+    shouldFocusError: false,
+  })
+
+  if (jobForm.isLoading || compensationForm.isLoading) {
```
mariechatfield, line 117: Aha I see now what you meant about how the "driven by SDK hooks" components should look vs the ones using suspense + base component

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
```diff
@@ -0,0 +1,290 @@
...
+    if (!compensationResult) return
+
+    onEvent(componentEvents.EMPLOYEE_COMPENSATION_UPDATED, compensationResult.data)
+  })
+
+  const errorHandling = composeErrorHandler([submitResult])
+  const isPending = jobForm.status.isPending || compensationForm.status.isPending
+
+  return (
+    <section className={classNames(styles.container, className)}>
```
mariechatfield, line 143: Why does this section need to have a width of 100%? Do we do this for all our SDK-driven forms?

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
```diff
@@ -0,0 +1,290 @@
...
+        }}
+        formHookResult={compensationForm}
+      />
+
+      <CompFields.PaymentUnit
+        label={t('management.wageFrequencyLabel')}
+        description={t('paymentUnitDescription')}
+        validationMessages={{ REQUIRED: t('validations.paymentUnit') }}
+        getOptionLabel={(unit: PaymentUnit) =>
+          t(`management.wageFrequencyOptions.${unit}` as const)
```
mariechatfield, line 227: Will this silently fail if we change the `PaymentUnit` types but not the intl keys? Super nit pick over here, since it seems unlikely that changes very often if ever, but I'm skeptical of all type casting in non-testing code (there too, tbh, just a little less so)

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
```diff
@@ -0,0 +1,290 @@
...
+        <JobFields.StateWcCovered
+          label={t('stateWcCoveredLabel')}
+          description={
+            <Trans
+              t={t}
+              i18nKey="stateWcCoveredDescription"
+              components={{
+                wcLink: (
+                  <Components.Link
+                    href="https://www.lni.wa.gov/insurance/rates-risk-classes/risk-classes-for-workers-compensation/risk-class-lookup#/"
```
mariechatfield, line 286: That's hilarious, is Washington state the only one that has this specific worker's compensation risk class? Or do other states just rely on Washington's existing codes so they don't have to maintain their own

src/components/Employee/Compensation/shared/useJobForm/useJobForm.tsx
serikjensen: Yeah you're not hallucinating, "rides" was nonsense from me. Rewrote the JSDoc in 9c6b42e1 to drop the metaphor and just say what `withTitleField: false` does and the steady-state context it's intended for.

src/i18n/en/Employee.Compensation.json
serikjensen: Quick survey of `src/i18n/en/*.json`: 179 straight apostrophes vs 27 smart, with straight as the dominant pattern in every namespace including `Employee.Compensation` itself (the other validations / labels on this file all use `'`). The two existing smart apostrophes here are in `stateWcCoveredDescription` / `stateWcClassCodeDescription` from an earlier add — they're the outliers, not the convention.

Leaning toward leaving the new key as straight `'` to match the predominant style. Happy to normalize the two existing smart ones in a separate cleanup PR if we want a real convention enforced; didn't want to mix that into the steady-state work.

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.test.tsx
serikjensen: Good call — added a `does not PUT compensations if the job PUT fails` test in 9c6b42e1 that stubs the job PUT to 500 and asserts that:

- `updateCompensationResolver` is never invoked
- neither `EMPLOYEE_JOB_UPDATED` nor `EMPLOYEE_COMPENSATION_UPDATED` fire

That pins the `onAllValid` short-circuit (`if (!jobResult) return`) so a refactor can't accidentally start running them in parallel or in the wrong order. The existing ordering test stays as the happy-path bookend.

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
serikjensen: Inherited the pattern from the onboarding `EditCompensation`, which got the same `.container { width: 100% }` in f7c2ea37:

> The migrated EditCompensation rendered without a width applied to its section, so embedding it in a flex/grid parent collapsed the form to its content's intrinsic width and wrapped every label aggressively.

`<section>` is a block element so it stretches to fill a block parent, but a `<Flex flexDirection="column">` parent treats it as a flex item and shrinks it to `min-content`, which is what was wrapping every label. Same fix is on `Employee/Profile` and onboarding `EditCompensation`; not something we do for *every* SDK form, just the ones we ended up embedding in flex/grid layouts where the bug showed up. Open to factoring this into a shared utility if the pattern keeps repeating.

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
serikjensen: Not silent — it would fail typecheck. `npm run i18n:generate` produces `src/types/i18next.d.ts` which augments i18next's `CustomTypeOptions.resources`, so `t()` is typed against the literal key union from the JSON. The `as const` on the template string is exactly what preserves the literal-union type from `\`management.wageFrequencyOptions.${PaymentUnit}\``; without it TS widens to `string` and we'd lose the check entirely.

So if `PaymentUnit` gains e.g. `'Day'` upstream but `wageFrequencyOptions` doesn't, the inferred key union includes `'management.wageFrequencyOptions.Day'`, which isn't in the resources map → `tsc` errors on the call site. Same story in reverse if we delete a key. Agreed on the general skepticism of casts in non-test code; this one happens to be load-bearing for the type narrowing rather than papering over it. Open to switching to an explicit `Record<PaymentUnit, string>` lookup if you'd rather have the exhaustiveness check be even more obvious, just a bit more verbose.

src/i18n/en/Employee.Compensation.json
mariechatfield: works for me, I wasn't sure since there were two smart quotes below but straight quotes works as well if that's the norm

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
mariechatfield: Good enough for me, was mostly checking this wasn't arbitrary CSS generated by an overeager LLM agent

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
mariechatfield: As long as it'll fail, no need to be verbose! This reads a lot nicer I just wasn't certain about how the `as const` worked with an interpolated string, but your explanation makes a lot of sense since the typescript compiler know the exact enum values and it's not just working with an unsafe string to begin with

src/components/Employee/Compensation/management/EditCompensation/EditCompensation.tsx
serikjensen: i'm actually not positive, but it's a recurring theme that washington is a very particular case lol

krisxcrash (approved):

PR #1863, opened 2026-05-18
fix: validate waiting period as integer in time off policy settings

krisxcrash:
## Summary
- The API requires `accrualWaitingPeriodDays` to be an integer (`z.number().int()`), but the Waiting period `NumberInputField` on the policy settings page allowed decimal input. Entering a decimal (e.g. `0.1`) caused an unhandled Zod validation error to surface raw to the user.
- Added `maximumFractionDigits={0}` to prevent decimal entry at the input level via react-aria's `NumberField` formatting.
- Added a `rules.validate` function on the field as a safety net that shows a clear inline error message ("Waiting period must be a whole number of days") if a non-integer value reaches the form through any path (e.g., programmatic defaults).
- No user input is silently modified — the user always sees what they entered and gets clear feedback if it's invalid.

## Test plan
- [x] Added unit tests in `PolicySettingsPresentation.test.tsx` verifying:
  - Decimal waiting period value blocks submission and shows validation error
  - Integer waiting period value allows normal submission
- [x] All existing PolicySettings tests pass (34 tests across container + presentation)

### DIFF
--- a/src/components/TimeOff/PolicySettings/PolicySettingsPresentation.tsx
+++ b/src/components/TimeOff/PolicySettings/PolicySettingsPresentation.tsx
@@ -1,12 +1,26 @@
 import { useId } from 'react'
 import { FormProvider, useForm, useWatch } from 'react-hook-form'
 import { useTranslation } from 'react-i18next'
+import { z } from 'zod'
+import { zodResolver } from '@hookform/resolvers/zod'
 import type { PolicySettingsFormData, PolicySettingsPresentationProps } from './PolicySettingsTypes'
 import styles from './PolicySettings.module.scss'
 import { Flex, ActionsLayout, NumberInputField, SwitchField } from '@/components/Common'
 import { Form as HtmlForm } from '@/components/Common/Form'
 import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentContext'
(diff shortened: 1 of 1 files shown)


src/components/TimeOff/PolicySettings/PolicySettingsPresentation.tsx
```diff
@@ -158,6 +158,15 @@ export function PolicySettingsPresentation({
...
                       max={20000}
+                      maximumFractionDigits={0}
+                      rules={{
+                        validate: (value: number) => {
+                          if (waitingPeriodEnabled && !isNaN(value) && !Number.isInteger(value)) {
+                            return t('policySettings.errors.waitingPeriodMustBeWholeNumber')
+                          }
+                          return true
+                        },
+                      }}
```
serikjensen: Is there any particular reason we are using the `rules` prop? If there's not, my rec would be to update this to live in the schema with the other validation rules instead

jeffredodd (approved):

serikjensen (changes requested):
actually going to request changes on this one and recommend we use the schema instead of rules unless there's a reason not to (see previous comment)

serikjensen (approved):
LGTM!

PR #1874, opened 2026-05-19
feat(prototype): employee compensation review prototype

aaronlee777:
## Summary
- Add a prototype screen for the employee compensation page under `sdk-app/src/design/prototypes/employee-compensation/`
- Includes a Review modal listing pending changes with a simulated cancellation flow (loading state -> remove from list -> success Alert)
- Update `Alert` to render element-only children inline with the label/header row (e.g. action buttons next to the label) while keeping the existing content row for text children

## Test plan
- [ ] Open the prototype in the SDK dev app and confirm the Compensation page renders
- [ ] Click Review on the warning Alert -> Modal opens with the two pending changes
- [ ] Click Cancel change -> button shows "Cancelling" loading, ~1.5s later the row is removed and a success Alert appears at the top of the list
- [ ] Confirm Alert with only a Button child renders the button inline with the label

### DIFF
--- a/sdk-app/src/design/prototypes/employee-compensation/employeeCompensation.module.scss
+++ b/sdk-app/src/design/prototypes/employee-compensation/employeeCompensation.module.scss
@@ -1,0 +1,8 @@
+.alertContainer {
+  padding: toRem(20);
+  border-bottom: 1px solid var(--g-colorBorderSecondary);
+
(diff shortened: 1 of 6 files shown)


serikjensen (approved):

serikjensen:
This was a reference during implementation, but we already shipped it so opting to close this one.

PR #1875, opened 2026-05-19
test(SDK-726): add tests and stories for DocumentsView

mariechatfield:
## Summary

Add unit and integration tests for the `DocumentsView` component, and update the stories for the presentation-only components in the employee dashboard.

## Demo

Screenshot 2026-05-19 at 12 47 59 PM
Screenshot 2026-05-19 at 12 48 03 PM
Screenshot 2026-05-19 at 12 48 08 PM

## Related

- Jira ticket: [SDK-726](https://gustohq.atlassian.net/browse/SDK-726)

## Testing

- `npm run storybook`
  + http://localhost:6006/?path=/story/domain-employee-dashboard-documentsview
  + http://localhost:6006/?path=/story/domain-employee-dashboard-basicdetailsview
  + http://localhost:6006/?path=/story/domain-employee-dashboard-taxesview

- `npm run Dashboard`

[SDK-726]: https://gustohq.atlassian.net/browse/SDK-726?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Employee/Dashboard/Dashboard.stories.tsx
mariechatfield, line 1: Image

These were all broken, so removed them and replaced them with presentation-only stories. Only exception is `JobAndPay` since there's not a view-only component we can use

src/components/Employee/Dashboard/DocumentsView.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export default {
+  title: 'Domain/Employee/Dashboard/DocumentsView',
+  decorators: [
+    (Story: React.ComponentType) => (
+      <Suspense fallback={<div>Loading translations...</div>}>
+        <I18nLoader>
+          <Story />
+        </I18nLoader>
+      </Suspense>
```
serikjensen: Nit: wondering why we end up needing to include these here? If it's because DocumentsView isn't using BaseComponent we might consider adding that here instead? or even just wrapping that component in BaseComponent if it is meant to be used in isolation

src/components/Employee/Dashboard/TaxesView.stories.tsx
```diff
@@ -0,0 +1,101 @@
...
+  useI18n('Employee.Dashboard')
+  return <>{children}</>
+}
+
+export default {
+  title: 'Domain/Employee/Dashboard/TaxesView',
+  decorators: [
+    (Story: React.ComponentType) => (
+      <Suspense fallback={<div>Loading translations...</div>}>
+        <I18nLoader>
```
serikjensen: Nit: Same here

src/test-utils/assertions.ts
```diff
@@ -1,15 +1,42 @@
...
+  expect(result.isLoading).toBe(false)
+}
+
+/** Expects a hook result has finished loading, and returns it narrowed to the ready variant. */
+export function requireNotLoading<T extends { isLoading: boolean }>(
+  result: T,
+): Extract<T, { isLoading: false }> {
+  assertNotLoading(result)
+  return result
+}
```
serikjensen, line 42: Looks like some useful additions!

serikjensen (approved):
Looks good!

src/components/Employee/Dashboard/DocumentsView.stories.tsx
mariechatfield: I _think_ this is because the views aren't using `BaseComponent` and are being rendered within `Dashboard.tsx` like

```tsx
<BaseLayout error={allErrors}>
  {selectedTab === 'basicDetails' && <BasicDetailsView... />}
  {selectedTab === 'documents' && <DocumentsView .../>}
</BaseLayout>
```

I'm a bit hesitant to change the prod code in a test/stories PR so it sounds like the fix would be to update the stories to wrap each view in base component _instead of_ suspense and i18n directly?

PR #1878, opened 2026-05-19
feat(Dashboard): multi-job table for nonexempt Compensation card

serikjensen:
## Summary

- When the primary job is **nonexempt** and the employee has **2+ jobs**, the Compensation card on the Dashboard's **Job and pay** tab now renders a table (job title with rate, pay type, start date) instead of the single-job detail view. Each row has a hamburger menu with **Edit** (always; emits `EMPLOYEE_COMPENSATION_CREATE` with the row's job) and **Delete** (non-primary only; opens a confirmation dialog mirroring the bank-account delete pattern, and emits `EMPLOYEE_JOB_DELETED` on confirm). The top-level Edit CTA is removed in the table state.
- An **Add another job** CTA renders in the card footer whenever the primary job is nonexempt and the employee has at least one job — so it shows in both the single-nonexempt detail view (per Figma) and the multi-job table view. It emits `EMPLOYEE_JOB_ADD`.
- Renamed `EMPLOYEE_COMPENSATION_UPDATE` → `EMPLOYEE_COMPENSATION_CREATE` to better describe the underlying API behavior (creates a new compensation record with an effective date rather than mutating in place). Not a breaking change — this event was added in the unreleased empty-state PR (#1877) and is not yet shipped.
- Added a new `EMPLOYEE_JOB_DELETED` event, fired after a successful delete.
- Extracted a tiny `derivePrimaryFlsaStatus` helper (with tests) and reused it in both `useEmployeeCompensation` (Dashboard) and `JobsList` (onboarding) to avoid duplicating the "find primary job's current compensation FLSA status" logic. Deliberately a pure helper, not a shared hook — the surrounding UX and data sources between Dashboard and onboarding are different enough that a shared hook would be more abstraction than payoff.
- Fixed a dead-branch bug in `JobAndPayView`: `Job.rate` is typed `string` in `@gusto/embedded-api`, so the existing `typeof rate === 'number'` guards were always false and the **Wage** row never rendered. Replaced with a local `parseJobRate` helper that uses `parseFloat` + `Number.isFinite` (matching the pattern already in `Payroll/helpers.ts`). The two existing call sites in this view now actually render.
- Updated partner-facing docs (`docs/workflows-overview/employee-dashboard.md`, `docs/reference/jobs-and-compensations.md`) to reflect the new event name, the new delete event, the table behavior, and the footer-CTA rule.

## Test plan

- [x] `npm run test -- --run src/components/Employee/Dashboard/Dashboard.test.tsx` — 15/15. Added 6 new cases: single-job header CTA emits `EMPLOYEE_COMPENSATION_CREATE` with the primary job; exempt single-job has no footer button; nonexempt single-job shows the detail view with both Edit and Add-another-job; multi-job table renders rows + suppresses the header Edit + shows Add-another-job; hamburger menu hides Delete on primary and shows Edit+Delete on non-primary; per-row Edit emits `EMPLOYEE_COMPENSATION_CREATE` with the right job; delete dialog opens and fires the mutation + emits `EMPLOYEE_JOB_DELETED` on confirm.
- [x] `npm run test -- --run src/components/Employee/Compensation/shared/derivePrimaryFlsaStatus.test.ts` — covers empty/undefined input, no-primary, primary with no current compensation, primary with current compensation, and multiple-jobs cases.
- [x] `npm run lint:check` — 0 errors (25 pre-existing warnings unrelated to this change).
- [x] `npm run format:check` — clean.
- [x] `npm run i18n:generate` — types regenerated.

### When there are multiple nonexempt jobs

https://github.com/user-attachments/assets/e25fe583-716c-44e5-8cae-bf2bd5759f9e

### Deleting a non primary job

https://github.com/user-attachments/assets/d62af61a-c3f7-41af-b885-1924c5a06d16


src/components/Employee/Dashboard/JobAndPayView.tsx
```diff
@@ -233,13 +333,18 @@ export function JobAndPayView({
...
         <Components.Box
-          withPadding={!!job}
+          withPadding={!hasMultipleJobs}
           header={
             <Components.BoxHeader
               title={t('jobAndPay.compensation.title')}
               action={
-                job ? (
-                  <Components.Button variant="secondary" onClick={onEditCompensation}>
+                hasMultipleJobs ? null : singleJob ? (
```
jeffredodd, line 393: Light readability nit here with the nested ternary

jeffredodd (approved):

krisxcrash (approved):

PR #1834, opened 2026-05-15
feat(e2e): time-off domain — scenario + spec rewrite

jeffredodd:
## Summary
- Add time-off-management scenario
- Rewrite time-off.spec.ts to use scenario fixture
- Move to e2e/tests/time-off/ directory

## Videos
https://github.com/user-attachments/assets/cdbca03f-5528-4d17-808e-f6491d5be2f2.webm

https://github.com/user-attachments/assets/f6171833-fa0d-44c3-a809-9241e0c30bd2.webm

https://github.com/user-attachments/assets/aef98ca4-f15d-4d68-87a1-2cd900c41078.webm

https://github.com/user-attachments/assets/9dc773f4-5668-4ec1-8d20-0d198a821878.webm

https://github.com/user-attachments/assets/449de71d-fb4b-46bf-bf8d-56535f39fb86.webm

Base: PR #1829 (e2e/infrastructure)

Notion tasks: #080-085

Made with [Cursor](https://cursor.com)

PR #1879, opened 2026-05-19
fix(TimeOff): validate waiting period as integer in policy settings

krisxcrash:
## Summary

The API requires \`accrualWaitingPeriodDays\` to be an integer, but the form allowed decimal input, causing an unhandled Zod validation error. This change:

- Sets \`maximumFractionDigits={0}\` on the waiting-period \`NumberInputField\` to prevent decimal entry at the input level.
- Adds a form-level \`validate\` rule as a safety net that surfaces a clear error message when a decimal sneaks through.
- Adds the i18n key \`policySettings.errors.waitingPeriodMustBeWholeNumber\`.

## Test plan

- [x] \`npm run test -- --run src/components/TimeOff/PolicySettings\` — 34 tests pass (2 new under \`describe('waiting period integer validation')\`)
- [x] \`npm run i18n:generate\` updates \`src/types/i18next.d.ts\`

## History

This PR was originally a larger bundle that also contained ~73 e2e tests. After triage:

- The e2e tests that catch real regressions (12 of them) were extracted, rewritten on top of the scenario-driven infrastructure used by the rest of the time-off domain, and merged into the \`e2e/timeoff-domain\` PR (#1834). Kristine credited as \`Co-authored-by\` on each of those commits.
- Brittle CSS/pixel/DOM-depth assertions and tests that duplicate the existing lifecycle suite were dropped.
- This PR was reduced to just the standalone product fix it always was.

### DIFF
--- a/src/components/TimeOff/PolicySettings/PolicySettingsPresentation.test.tsx
+++ b/src/components/TimeOff/PolicySettings/PolicySettingsPresentation.test.tsx
@@ -246,6 +246,57 @@ describe('PolicySettingsPresentation', () => {
     })
   })
 
+  describe('waiting period integer validation', () => {
+    it('blocks submission and shows an error when waiting period is a decimal', async () => {
+      const user = userEvent.setup()
+      renderHoursWorked({
+        defaultValues: {
+          waitingPeriodEnabled: true,
+          waitingPeriod: 1.5,
+        },
+      })
+
+      await waitFor(() => {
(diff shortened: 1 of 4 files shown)


jeffredodd (changes requested):
Yay this is great that you did this for validating changes. Throwing a request changes on here so I can merge it into the e2e effort after I get the foundation merged.

jeffredodd (approved):

PR #1885, opened 2026-05-20
feat: add useClientPagination hook for in-memory list pagination

jeffredodd:
## Summary

Adds a reusable `useClientPagination(items, options)` hook that owns in-memory pagination plus an optional search predicate over a `T[]`, returning a `PaginationControlProps`-shaped object that drops straight into `useDataView` and `PaginationControl`.

Pairs with the existing server-side `usePagination`:

| Driver | Source | Use when |
|---|---|---|
| `usePagination` | `x-total-pages` / `x-total-count` headers | Paging through an API endpoint |
| `useClientPagination` | In-memory `T[]` | Caller already has the full list |

Both return the same shape, so swapping a list from server-paginated to client-paginated (or vice versa) is a one-line change at the call site.

### First adopter: `HolidayPolicyDetail`

Today this component hand-rolls a `searchValue` + `useMemo` filter over the policy's enrolled employees, **with no pagination at all**. This PR swaps it for `useClientPagination` at 10 per page, so rosters with more than 5 enrolled employees now render a pager.

Net code change in that file: drops the local `searchValue` state and the filter `useMemo`, adds one `useClientPagination(...)` call, wires `pagination` into the existing `employees` prop that `PolicyDetailLayout` already exposed.

### Behavior

- Whitespace-only queries (`'   '`) are treated as no-search; the predicate is never called.
- The predicate is invoked with the **trimmed, deferred** value, so partner predicates do not need to defensively trim.
- `searchValue` from the hook is the **raw** input the user typed (controlled-input semantics), even when the trimmed version is empty.
- Identity of `data` / `pagination` is stable across no-op rerenders so downstream `useMemo`s in `useDataView` are not invalidated.

## Test plan

- [x] 26 unit tests on the hook: initial state, page nav and clamps, items-per-page reset, items-per-page no-op when unchanged, exact-page totals (10 items / 5 per page == 2 pages, not 3), all three `PaginationItemsPerPage` values, large dataset (10k items: count, last-page slice 9950-9999, middle-page slice), zero matches, query-refine clamp via `safeCurrentPage`, successive queries, whitespace-only no-op, predicate receives trimmed deferred value, search + items-per-page interaction, predicate identity change re-filters, referential stability across no-op rerenders, new `allItems` reference re-derives.
- [x] 3 integration tests on `HolidayPolicyDetail`: pager hidden when roster <= 5, paginates at 10 per page and navigates between pages with 12 enrolled employees, search resets to page 1 (asserts page-2 content disappears).
- [x] Pre-existing `HolidayPolicyDetail` search test still passes against the new hook-backed implementation.
- [x] `npm run test -- --run src/components/TimeOff src/hooks` -> 433/433 pass.
- [x] `npm run lint:check` -> 0 errors (25 pre-existing warnings, none in touched files).
- [x] `npm run format:check` clean.

## Follow-ups (not in this PR)

- `useSelectEmployeesData` ([#1843](https://github.com/Gusto/embedded-react-sdk/pull/1843)) hand-rolls the same state machine. Once that PR merges main, it can adopt the hook in a small follow-up.
- `TimeOffPolicyDetail` has the same client-side-search-without-pagination pattern as `HolidayPolicyDetail` did; left for a follow-up PR to keep this diff scoped.

### DIFF
--- a/src/hooks/useClientPagination/useClientPagination.ts
+++ b/src/hooks/useClientPagination/useClientPagination.ts
@@ -1,0 +1,114 @@
+import { useCallback, useDeferredValue, useMemo, useState } from 'react'
+import type {
+  PaginationControlProps,
+  PaginationItemsPerPage,
+} from '@/components/Common/PaginationControl/PaginationControlTypes'
+
+interface UseClientPaginationOptions<TItem> {
+  /**
+   * Optional substring/fuzzy/etc. predicate run against the deferred,
+   * trimmed search value. The hook owns search state and resets to page 1
+   * whenever the value changes. A trimmed-empty value is treated as "no
+   * search" and the predicate is not called. Omit to skip search entirely.
+   */
+  searchPredicate?: (item: TItem, query: string) => boolean
+  /** Display rows-per-page. Default 5. */
+  defaultItemsPerPage?: PaginationItemsPerPage
+}
+
+export interface UseClientPaginationResult<TItem> {
+  /** Current-page slice. */
+  data: TItem[]
+  /** Drop-in `PaginationControlProps` for `useDataView` / `PaginationControl`. */
+  pagination: PaginationControlProps
+  /** Current search value (controlled). */
+  searchValue: string
+  actions: {
+    onSearchChange: (value: string) => void
+    onSearchClear: () => void
+  }
+}
+
+/**
+ * Client-side pagination + optional search over an in-memory array.
+ *
+ * Returns a `PaginationControlProps`-shaped pagination object that drops
+ * straight into `useDataView` / `PaginationControl`. Use this when the
+ * caller already has the full list of items and wants to paginate/search
+ * it locally. Pair with `usePagination` if you want to page through an
+ * API endpoint instead.
+ */
+export function useClientPagination<TItem>(
+  allItems: TItem[],
+  options: UseClientPaginationOptions<TItem> = {},
+): UseClientPaginationResult<TItem> {
+  const { searchPredicate, defaultItemsPerPage = 5 } = options
+
+  const [searchValue, setSearchValue] = useState('')
+  const deferredSearchValue = useDeferredValue(searchValue)
+  const [currentPage, setCurrentPage] = useState(1)
+  const [itemsPerPage, setItemsPerPage] = useState<PaginationItemsPerPage>(defaultItemsPerPage)
+
+  const searchFilteredItems = useMemo<TItem[]>(() => {
+    if (!searchPredicate) return allItems
+    const trimmed = deferredSearchValue.trim()
+    if (!trimmed) return allItems
+    return allItems.filter(item => searchPredicate(item, trimmed))
+  }, [allItems, deferredSearchValue, searchPredicate])
+
+  const totalCount = searchFilteredItems.length
+  const totalPages = Math.max(1, Math.ceil(totalCount / itemsPerPage))
+  // Clamp `currentPage` against the latest `totalPages` so a shrinking list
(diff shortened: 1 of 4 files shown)


src/hooks/useClientPagination/useClientPagination.ts
```diff
@@ -0,0 +1,114 @@
...
+        }
+      },
+    }),
+    [safeCurrentPage, totalPages, totalCount, itemsPerPage],
+  )
+
+  const onSearchChange = useCallback((value: string) => {
+    setSearchValue(value)
+    setCurrentPage(1)
+  }, [])
```
serikjensen, line 131: Nit: any need to debounce? i've been looking at the candidate take home exercise for too long 😛

serikjensen (approved):
LGTM!

src/hooks/useClientPagination/useClientPagination.ts
```diff
@@ -0,0 +1,114 @@
...
+} from '@/components/Common/PaginationControl/PaginationControlTypes'
+
+interface UseClientPaginationOptions<TItem> {
+  /**
+   * Optional substring/fuzzy/etc. predicate run against the deferred,
+   * trimmed search value. The hook owns search state and resets to page 1
+   * whenever the value changes. A trimmed-empty value is treated as "no
+   * search" and the predicate is not called. Omit to skip search entirely.
+   */
+  searchPredicate?: (item: TItem, query: string) => boolean
```
krisxcrash, line 17: q: Will this be extendable if we were to eventually add filtering?

krisxcrash: i know with the api, there are limitations in terms of filtering + search + pagination and they have to happen in a specific order so I am just curious.

jeffredodd: 100%! I started to think about filtering but I think two things are true right now
- Our APIs don't have a consistent way to filter API wide
- Our dataview hasn't worked too much on sorting filtering

So my suggestion is when we need to improve those things we come in and add filtering/sorting.

krisxcrash (approved):

src/hooks/useClientPagination/useClientPagination.ts
jeffredodd: Bwahahahaha what a good call out. This is why we don't do client side search :-P

src/hooks/useClientPagination/useClientPagination.ts
krisxcrash: makes total sense!

src/hooks/useClientPagination/useClientPagination.ts
jeffredodd: Good call -- I went back and forth on this. Initial reasoning was "this is purely client-side, no requests to debounce, and `useDeferredValue` already deprioritizes the filter pass." But for big lists or a fancier predicate the filter itself can still chew CPU on every keystroke, and `useDeferredValue` only helps when React has work it can interrupt -- it won't actually skip the synchronous filter call.

So I added a built-in debounce (default 120ms, configurable via `searchDebounceMs`, pass 0 to opt out). `searchValue` returned from the hook is still synchronous so the input stays controlled; the predicate + page reset are gated by the debounce, then further deferred by `useDeferredValue`.

Pushed in a6e0b00. Coverage:
- searchValue updates synchronously even mid-debounce (controlled-input feel)
- rapid typing collapses to a single predicate pass with the final value
- searchDebounceMs: 0 runs the predicate synchronously
- custom windows respected exactly

PR #1890, opened 2026-05-21
perf(Dashboard): lazy-load tab data, cut first-paint requests

dmortal:
## Summary

Employee Dashboard fired every tab's data fetches on mount (Suspense), blocking render until the slowest of 7+ requests resolved. A user who only opened Basic Details still paid for taxes, forms, paystubs, and garnishments bandwidth.

This makes data fetching lazy per tab — each tab's data hook only mounts when its tab is selected, wrapped in its own `<Suspense>` boundary so tabs paint independently.

## Changes

- Each tab now owns its data. `BasicDetailsView`, `TaxesView`, and `DocumentsView` get a co-located `…WithData` container that calls the corresponding hook; presentational views stay pure so existing tests/stories keep passing props.
- `JobAndPayView` (already monolithic) absorbs `useEmployeeCompensation`, the paystub download, and the cancel-pending-change handler that used to live in `Dashboard.tsx`. Compensation errors join the local `composeErrorHandler`.
- `Dashboard.tsx` drops the four data hooks and the all-or-nothing loading gate; each tab mounts inside its own `<Suspense fallback={<BaseLayout isLoading />}>` (323 → 178 lines).
- `useEmployeeCompensation` exposes `employeeFirstName` (free off the shared employee query) so JobAndPayView no longer needs it threaded as a prop.

The `EMPLOYEE_FEDERAL_TAXES_EDIT` event payload (`{ employeeId, federalTaxes }`) is preserved — the federal-tax record is threaded through the `onEditFederalTaxes` callback so the public event contract is unchanged.

## Testing

Verified via Playwright against `/employee/DashboardFlow` (gws-flows local):

| Tab                   | Fires when           |
|-----------------------|----------------------|
| Basic details (mount) | `employee?include=all_compensations`, `home_addresses`, `work_addresses` |
| Job and pay (click)   | `pay_stubs`, `payment_method`, `bank_accounts`, `garnishments` |
| Taxes (click)         | `federal_taxes`, `state_taxes` |
| Documents (click)     | `forms` |

**3 requests on first paint instead of 7+.**

Commands run, all green:
- \`npx tsc --noEmit\`
- \`npm run lint\` — 0 errors (25 pre-existing warnings in unrelated files)
- \`npm run test -- --run src/components/Employee/Dashboard src/components/Employee/Deductions src/components/Employee/PaymentMethod\` — **170/170 pass**

### DIFF
--- a/src/components/Employee/Dashboard/hooks/useEmployeeCompensation.tsx
+++ b/src/components/Employee/Dashboard/hooks/useEmployeeCompensation.tsx
@@ -31,6 +31,10 @@ interface UseEmployeeCompensationReady extends BaseHookReady<
     hasMultipleJobs: boolean
     pendingChanges: PendingCompensationChange[]
     payStubs: EmployeePayStub[]
+    /** First name from the shared employee fetch; useful for cosmetic copy
+     *  in alerts (e.g. "Heads up, Jane has pending changes"). Optional
+     *  because the employee record can omit it. */
+    employeeFirstName?: string
   },
   { isPending: boolean; cancellingCompensationUuid: string | null }
 > {
@@ -135,6 +139,7 @@ export function useEmployeeCompensation({
       hasMultipleJobs,
       pendingChanges,
       payStubs,
+      employeeFirstName: employee?.firstName ?? undefined,
     },
     status: {
       isPending,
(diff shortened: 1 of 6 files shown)


src/components/Employee/Dashboard/hooks/useEmployeeCompensation.tsx
```diff
@@ -135,6 +139,7 @@ export function useEmployeeCompensation({
       hasMultipleJobs,
       pendingChanges,
       payStubs,
+      employeeFirstName: employee?.firstName ?? undefined,
```
serikjensen, line 142: My recommendation here would be to just return the full fetched employee which should be the convention with the `data` attribute, we're basically just returning the raw data that we fetch as part of the hook in case the consumer needs it (exactly what's happening here)

serikjensen (approved):
LGTM!

PR #1901, opened 2026-05-21
fix: show employee names in ineligible-employee add errors

krisxcrash:
## Summary
- Catches 422 errors from the add-employees API and resolves error UUIDs to human-readable employee names.
- Includes the API-provided reason for each ineligible employee in the error message.
- Adds i18n keys for the formatted error messages.

Fixes bug: adding an ineligible employee to a policy shows UUID only, no reason.

## Test plan
- [x] Existing `SelectEmployeesTimeOff.test.tsx` passes (25/25)
- Manually verify: attempt to add an ineligible employee → error shows their name and reason

### DIFF
--- a/src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
+++ b/src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
@@ -190,13 +190,38 @@ function SelectEmployeesTimeOffInner({
       await baseSubmitHandler({}, async () => {
         let policyResult: unknown
         if (toAdd.length > 0) {
-          const response = await addEmployees({
-            request: {
-              timeOffPolicyUuid: policyId,
-              requestBody: { employees: buildAddPayload(toAdd) },
-            },
-          })
-          policyResult = response.timeOffPolicy
+          try {
+            const response = await addEmployees({
+              request: {
+                timeOffPolicyUuid: policyId,
+                requestBody: { employees: buildAddPayload(toAdd) },
+              },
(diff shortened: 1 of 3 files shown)


src/components/TimeOff/TimeOffManagement/SelectEmployees/SelectEmployeesTimeOff.tsx
```diff
@@ -190,13 +190,38 @@ function SelectEmployeesTimeOffInner({
...
+                .map(e => {
+                  const uuid = e.errorKey
+                  const employee = selectedEmployeesRef.current.get(uuid)
+                  const name = employee
+                    ? `${employee.firstName ?? ''} ${employee.lastName ?? ''}`.trim()
+                    : null
+                  const reason = e.message ?? t('errors.ineligibleUnknownReason')
+                  return name ? `${name}: ${reason}` : reason
+                })
+                .filter(Boolean)
```
mariechatfield, line 213: Image

I had to comment out Aaron's fix in `useSelectEmployeeData.ts` to even get to this error, and I realize this is an edge case without a very good error message. But this looks a bit odd

```ts
      showEmployees
        // .filter(e => isStartedByToday(e.jobs?.find(job => job.primary)?.hireDate))
```

mariechatfield: Do you have a test case that you would recommend instead to see what this would look like this with a per-UUID error message the way we expect?

krisxcrash:
this was fixed with @aaronlee777 pr that was merged post testfest

PR #1941, opened 2026-05-26
ci: run e2e selectively per domain and add e2e-success aggregator

jeffredodd:
## Summary

- **`e2e-success` aggregator** — single required check that branch protection can gate on. Aggregates `scenarios + e2e-changes + e2e-setup + e2e` (the matrix's `result` rolls up across every shard). Treats `skipped` as success so PRs that scope down to a subset of shards still satisfy the gate. Uses `if: !cancelled()` so a failed upstream actually fails this check instead of being skipped (which branch protection treats as passing).
- **Re-enabled e2e on `main`** — main always runs every domain unconditionally so it stays the canonical "all green" signal.
- **Selective shards on PRs** — only run the domain shards whose surface was actually touched.

## How selection works

Two-stage filter in the new `e2e-changes` job:

1. **Extension allowlist** — drop files whose extension can't affect runtime/test behavior (`.md`, `LICENSE`, `.gitignore`, images, etc.). A docs-only PR produces an empty filtered diff and skips e2e entirely.
2. **Domain ownership** — of the surviving files, if any sit *outside* every known domain's surface (`src/components/<Pascal>/` or `e2e/tests/<kebab>/`), treat the change as cross-cutting and run every shard. Otherwise run only the domains whose surface was touched.

The domain list is discovered dynamically from `e2e/tests/<domain>/`. The naming contract is `e2e/tests/<kebab>` ↔ `src/components/<Pascal>` (e.g. `time-off` ↔ `TimeOff`). Adding a new domain folder is sufficient — no CI edit required.

Stage 2 is safe by default: new code in unfamiliar locations (e.g. a fresh top-level dir like `src/middleware/`) fans out to all shards rather than silently producing an empty matrix.

## Behavior matrix

| Change | Result |
|---|---|
| `src/components/Payroll/*.tsx` | `payroll` only |
| `src/components/Payroll/*` + `src/components/Employee/*` | `employee + payroll` |
| `src/components/Common/*` / `src/hooks/*` / `package.json` / `playwright.demo.config.ts` | all shards |
| New top-level dir like `src/middleware/auth.ts` | all shards (safe default) |
| Push to `main` / `workflow_dispatch` | all shards (always) |
| `README.md` / `docs/**.md` / `LICENSE` / images | skip e2e entirely |
| `payroll source + README.md` | `payroll` only |

## Follow-up (not in this PR)

After merge, set `e2e-success` as a required status check in branch protection. That's a manual GitHub UI step.

## Test plan

- [ ] CI on this PR runs only the e2e shards relevant to `.github/workflows/ci.yaml` — which is itself cross-cutting, so all shards should run
- [ ] `e2e-success` job appears in checks and passes
- [ ] After merge, verify a follow-up PR touching only one domain runs only that domain's shard
- [ ] After merge, verify a docs-only PR skips e2e entirely
- [ ] After merge, verify a push to `main` runs every shard

### DIFF
--- a/.github/scripts/e2e-discover-domains.sh
+++ b/.github/scripts/e2e-discover-domains.sh
@@ -29,20 +29,26 @@
 #   e2e/tests/<kebab>  ↔  src/components/<Pascal>     (e.g. time-off ↔ TimeOff)
 # The list is discovered from disk on every run; no maintenance needed.
 #
-# Portability: depends on bash 4+ (`mapfile`) and GNU find (`-printf`).
-# Both are present on the GitHub-hosted ubuntu-latest runner this is
-# intended for. Running on macOS for local debugging will fail the
-# `mapfile` line — install `bash` and `findutils` from Homebrew or just
-# run the workflow.
+# Portability: written for POSIX-ish bash 3.2+ and either GNU or BSD find,
+# so it runs natively on the GitHub-hosted ubuntu-latest runner AND on
+# macOS for local debugging. Required tools: bash, find, sed, awk, jq, git.
 
 set -euo pipefail
 
 if [[ -z "${GITHUB_OUTPUT:-}" ]]; then
-  echo "ERROR: GITHUB_OUTPUT is not set — this script must run inside a GitHub Actions step" >&2
+  echo "ERROR: GITHUB_OUTPUT is not set — set it to a writable file path" >&2
+  echo "       (GitHub Actions sets this automatically; locally use a tempfile)" >&2
   exit 2
 fi
 
-mapfile -t domains < <(find e2e/tests -mindepth 1 -maxdepth 1 -type d -printf '%f\n' | sort)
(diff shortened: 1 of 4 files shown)


mariechatfield (approved):
I trust you and Cursor but it would be super radical if this was rewritten as a script that we could run locally to verify what outputs it thinks are updated, and that CI runs. GitHub actions are so devilishly annoying to test and reading bash is hard enough without it being inlined in YAML

jeffredodd:
> I trust you and Cursor but it would be super radical if this was rewritten as a script that we could run locally to verify what outputs it thinks are updated, and that CI runs. GitHub actions are so devilishly annoying to test and reading bash is hard enough without it being inlined in YAML

I totally agree! Right now the big elephant 🐘 we're dealing with is getting Demo to behave. 

I do believe you should be able to run `npm run e2e` which runs the suite fully locally off your code.

PR #1948, opened 2026-05-27
fix: add indeterminate state to select-all checkbox in data views (SDK-890)

aaronlee777:
## Summary

- The select-all checkbox in `DataTable` and `DataCards` now shows a dash icon and enters the native indeterminate state when some (but not all) rows are selected
- Previously it was always either fully checked or unchecked, with no partial-selection indicator

## Changes

- Added `isIndeterminate` prop to `Checkbox` — sets `input.indeterminate` via a ref (required since indeterminate is a DOM property, not an HTML attribute) and renders the dash icon from the existing `checkbox_indeterminate.svg` asset
- Updated `useSelectionState` to compute and return `isIndeterminate` alongside `allSelected`
- `DataTable` and `DataCards` both consume the new value and pass it to their select-all checkbox

## Related

- Jira ticket: [SDK-890](https://gustohq.atlassian.net/browse/SDK-890)

## Testing

- New tests cover the indeterminate DOM property on `Checkbox`, and the partial-selection state on the header checkbox in both `DataTable` and `DataCards`
- Run: `npm run test -- --run src/components/Common/UI/Checkbox src/components/Common/DataView/DataTable src/components/Common/DataView/DataCards`
- Storybook: `DataViewSelectableWithPagination` story — select a subset of rows to verify the dash icon appears in the header checkbox

[SDK-890]: https://gustohq.atlassian.net/browse/SDK-890?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/component-adapter/component-inventory.md
+++ b/docs/component-adapter/component-inventory.md
@@ -231,22 +231,23 @@
 
 ## CheckboxProps
 
-| Prop                        | Type                            | Required | Description                                                            |
-| --------------------------- | ------------------------------- | -------- | ---------------------------------------------------------------------- |
-| **value**                   | `boolean`                       | No       | Current checked state of the checkbox                                  |
-| **onChange**                | `(value: boolean) => void`      | No       | Callback when checkbox state changes                                   |
-| **inputRef**                | `Ref<HTMLInputElement \| null>` | No       | React ref for the checkbox input element                               |
-| **isInvalid**               | `boolean`                       | No       | Indicates if the checkbox is in an invalid state                       |
-| **isDisabled**              | `boolean`                       | No       | Disables the checkbox and prevents interaction                         |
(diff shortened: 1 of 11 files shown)


serikjensen (changes requested):
I think we need to have a chat on this one. For accessibility i think we will need to set aria-checked to mixed https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-checked#mixed

but this is also going to have implications for how we navigate the value/onChange ecosystem since we don't want a situation where you can set the checkbox value to true/false and also indeterminate at the same time

Could you find some time to go over this together? would also be good to include @jeffredodd

serikjensen:
We had a follow up discussion around UX. Going to close this one as we will look into a different approach that is better for component adapters

PR #1819, opened 2026-05-14
fix: constrain starting balance input width in employee table (SDK-893)

aaronlee777:
## Summary

Constrains the starting balance input in the Select Employees table to 120px so the column doesn't stretch to fill available space, and right-aligns it within the cell (when in table mode) so the input sits flush with the column's right-aligned header.

## Changes

- New `SelectEmployeesPresentation.module.scss` with a `.balanceInput` class: `width: toRem(120)` plus `margin-left: auto` scoped to `.react-aria-Cell` ancestors only.
- Applied the class via `className` on the balance `TextInput` in `SelectEmployeesPresentation.tsx`.

## Related

- Jira ticket: [SDK-893](https://gustohq.atlassian.net/browse/SDK-893)

## Testing

- Storybook: `npm run storybook` → Company / TimeOff / SelectEmployees stories
- Verify in table mode the balance input is 120px wide and right-aligned within its cell
- Verify in responsive DataCard mode the input remains left-aligned (no `margin-left: auto` because there's no `.react-aria-Cell` ancestor)

[SDK-893]: https://gustohq.atlassian.net/browse/SDK-893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1949, opened 2026-05-27
fix: use max-width to constrain starting balance input width (SDK-893)

aaronlee777:
## Summary

Regression fix for SDK-893. The starting balance input width fix from #1819 was regressed by #1843.

**Root cause:** The original fix used `width: toRem(120)` on `.balanceInput`, which shares an element with `FieldLayout`'s `.root { width: 100% }`. Both are single-class selectors (equal specificity), so the rule appearing *later* in the CSS bundle wins. #1843's large module graph refactor flipped the Vite bundle order such that `FieldLayout.module.scss` now loads after `SelectEmployeesPresentation.module.scss`, letting `width: 100%` win again.

**Fix:** Add `max-width: toRem(120)` alongside `width`. `max-width` is not set by `FieldLayout` or `DataCards`, so it has no specificity conflict and always caps the input at 120px regardless of bundle order.

## Test plan

- [ ] `npm run storybook` → Company / TimeOff / SelectEmployees stories — verify balance input is ~120px wide in table mode and card mode
- [ ] Verify right-alignment within the table cell is preserved

## Related

- Original fix: #1819 (SDK-893)
- Regressing PR: #1843

### DIFF
--- a/src/components/Common/DataView/DataTable/DataTable.module.scss
+++ b/src/components/Common/DataView/DataTable/DataTable.module.scss
@@ -1,0 +1,4 @@
+.cellEnd {
+  display: flex;
+  justify-content: flex-end;
+}
--- a/src/components/Common/DataView/DataTable/DataTable.test.tsx
+++ b/src/components/Common/DataView/DataTable/DataTable.test.tsx
@@ -196,6 +196,17 @@ describe('DataTable Component', () => {
     })
   })
 
+  describe('column justify', () => {
+    test('wraps header and cell content in a flex-end container when justify is end', () => {
+      renderTable({
+        columns: [{ key: 'name', title: 'Name', justify: 'end', render: item => item.name }],
+      })
+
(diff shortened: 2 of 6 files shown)


serikjensen (changes requested):
Reading the description, it looks like the real issue here is a specificity conflict where the core input css is conflicting with the component styles?

We want to avoid having classes like this conflict with out component styles. A good rule of thumb would be to avoid having css classes set on the UI component from adapters.

The way that you can get around this is to place inputs inside of containers that have widths set.

ex. instead of doing

```

...

<Components.TextInput className={styles.maxWidth} ... />
```

do

```

...

<div className={styles.maxWidth}>
  <Components.TextInput ... />
</div>
```

all components should fill the width of their containing elements. That allows us to define the layout without having to conflict with partner styles. Partners just define the adapters as full width

serikjensen (approved):

PR #1950, opened 2026-05-27
fix(SDK-933): track agency-required garnishment fields dynamically

dmortal:
## Summary

In the child-support garnishment form (Add deduction → Garnishment → Child Support → pick an agency), fields that the agency requires (e.g. CSE Case Number for South Carolina) were rendered with an "(optional)" label and could be submitted blank — the API would then reject with "Case number is required". The label was wrong AND the form skipped client-side validation, so the user saw a server error instead of inline guidance.

Root cause: \`createChildSupportGarnishmentFormSchema\` was called once at hook init with no agency context. The three agency-attribute fields (\`caseNumber\`, \`orderNumber\`, \`remittanceNumber\`) defaulted to \`'never'\` required, so \`isRequired\` stayed \`false\` for the whole form lifetime. The view rendered them conditionally on \`requiredAttrKeys\`, but their labels still said "(optional)".

Fix: add an \`agencyList\` option to the schema factory. When provided, requiredness for the three fields is computed via predicates that look up the agency by the form's \`state\` value at validation and metadata-derivation time. No schema rebuild is needed when the user switches states — \`useDeriveFieldsMetadata\` already watches predicate dependencies, and \`superRefine\` evaluates the predicates at submit. The hook passes the agency list it already loads, so picking an agency that requires a field immediately flips its \`isRequired\` to \`true\` (dropping the "(optional)" suffix) and an empty submit is rejected client-side before hitting the API.

The legacy \`selectedAgency\` option still produces static rules so external callers of the public factory aren't affected.

## Changes

- \`childSupportGarnishmentFormSchema\`: new \`agencyList\` option. When set, \`requiredFieldsConfig\` is predicate-based against \`data.state\`. Otherwise falls back to the existing static \`selectedAgency\` behavior.
- \`useChildSupportGarnishmentForm\`: pass \`agencyList\` into the factory.
- New test: \`marks agency-required attribute fields as required in metadata after selecting that agency\`. Verified to fail on \`main\` and pass with the fix.

## Out of scope

- The ticket also notes "There was a problem with your submission" rendering below the Deduction type radio instead of at the top. That's a separate concern (the \`BaseLayout\` error banner sits inside the child-support view, so the agency picker renders before it). Tracking as follow-up.

## Related

- Jira: [SDK-933](https://gustohq.atlassian.net/browse/SDK-933)

## Testing

- \`npm run test -- --run src/components/Employee/Deductions\` → 33/33 passed (32 prior + 1 new)
- Stashed only the schema/hook changes and re-ran the new test → fails on main (\`isRequired\` returns \`false\`), confirming the test guards the regression.

Manual: \`npm run storybook\` → ChildSupportGarnishment story → pick "South Carolina Integrated Child Support Services" → confirm CSE Case Number no longer says "(optional)" and submitting blank surfaces an inline required error rather than an API error.

[SDK-933]: https://gustohq.atlassian.net/browse/SDK-933?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Employee/Deductions/shared/useChildSupportGarnishmentForm/childSupportGarnishmentFormSchema.ts
+++ b/src/components/Employee/Deductions/shared/useChildSupportGarnishmentForm/childSupportGarnishmentFormSchema.ts
@@ -101,25 +101,47 @@ export type ChildSupportGarnishmentFormOutputs = ChildSupportGarnishmentFormData
 interface ChildSupportGarnishmentFormSchemaOptions {
   mode?: 'create' | 'update'
   /**
-   * The agency record matching the currently selected `state`. The agency's
-   * `requiredAttributes` determine which of `caseNumber` / `orderNumber` /
-   * `remittanceNumber` are required. When omitted (no agency selected yet),
-   * all three are optional.
+   * The agency record matching the currently selected `state`. Used only when
+   * `agencyList` is not provided — `requiredAttributes` are pinned to this
+   * single agency, so the schema must be rebuilt whenever the user picks a
+   * different state. Prefer passing `agencyList` instead so requiredness
+   * tracks the form's `state` value dynamically.
    */
   selectedAgency?: Agencies | null
+  /**
+   * Full list of agencies. When provided, the schema's requiredness for
+   * `caseNumber` / `orderNumber` / `remittanceNumber` is computed at validation
+   * time by looking up the agency whose `state` matches the form's `state`
+   * value — so a single schema instance stays correct as the user changes
+   * states. Takes precedence over `selectedAgency`.
+   */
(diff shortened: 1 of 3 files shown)


serikjensen (approved):

dmortal:
Addressed in f0dcb80: read `data.state` into a local `const` before the `agencyList.find` so `buildFormSchema`'s Proxy-based predicate-dep detection always observes the `state` access, regardless of whether `agencyList` is empty at schema-build time. Tests still pass (33/33).

PR #1955, opened 2026-05-27
fix(SDK-946): remount DocumentViewer embed when the PDF url changes

dmortal:
## Summary

After completing an e-signature, the embedded PDF viewer transitioned to read-only but the PDF itself didn't render. Manually reloading the tab or switching tabs back made the embed appear. The "download this document" link worked the whole time — proving the URL itself updated; the iframe just didn't reload.

Root cause: the global \`@gusto/embedded-api\` mutation auto-invalidation refetches the signed PDF URL after \`signForm\` succeeds, and \`DocumentViewer\` rerenders with a new \`src\`. React was reusing the existing \`<embed>\` DOM node across that rerender, so the browser's PDF plugin never reloaded — the viewer stayed on the previous (unsigned) PDF or went blank.

Fix: add \`key={url}\` to both \`<embed>\` elements in \`DocumentViewer\`. A URL change now forces the DOM node to unmount and remount, which the plugin treats as a fresh document load.

## Changes

- \`src/components/Common/DocumentViewer/DocumentViewer.tsx\` — \`key={url}\` on the two \`<embed>\` render paths (wide + narrow container), plus a short comment explaining the why.
- \`src/components/Common/DocumentViewer/DocumentViewer.test.tsx\` — new test renders DocumentViewer with one URL, rerenders with another, asserts the \`<embed>\` reference changed (verified to fail on \`main\`).

## Related

- Jira: [SDK-946](https://gustohq.atlassian.net/browse/SDK-946)

## Testing

- \`npm run test -- --run src/components/Common/DocumentViewer src/components/Employee/Documents\` → 94/94 passed (new + sibling)
- Stashed the source change and re-ran → new test fails on \`main\`, confirming it guards the regression.

Manual: \`npm run sdk-app\` → onboarded employee with unsigned W-4 → Documents tab → View → sign → confirm the embedded PDF renders immediately instead of going blank.

[SDK-946]: https://gustohq.atlassian.net/browse/SDK-946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Common/DocumentViewer/DocumentViewer.test.tsx
+++ b/src/components/Common/DocumentViewer/DocumentViewer.test.tsx
@@ -1,0 +1,43 @@
+import { describe, it, expect } from 'vitest'
+import { DocumentViewer } from './DocumentViewer'
+import { renderWithProviders } from '@/test-utils/renderWithProviders'
+
+const getEmbed = (container: HTMLElement) => container.querySelector('embed')
+
+describe('DocumentViewer', () => {
+  it('remounts the embed when the document URL changes so the browser PDF plugin reloads', () => {
+    const { container, rerender } = renderWithProviders(
+      <DocumentViewer
+        url="https://example.com/unsigned.pdf"
+        title="W-4"
+        viewDocumentLabel="Download"
+      />,
+    )
+
+    const firstEmbed = getEmbed(container)
+    expect(firstEmbed).not.toBeNull()
+    expect(firstEmbed?.getAttribute('src')).toContain('unsigned.pdf')
+
+    rerender(
+      <DocumentViewer
+        url="https://example.com/signed.pdf"
+        title="W-4"
+        viewDocumentLabel="Download"
+      />,
+    )
+
+    const secondEmbed = getEmbed(container)
+    expect(secondEmbed).not.toBeNull()
+    expect(secondEmbed?.getAttribute('src')).toContain('signed.pdf')
+    // A new DOM node — the `key={url}` on <embed> forces React to unmount the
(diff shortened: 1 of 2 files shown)


src/components/Common/DocumentViewer/DocumentViewer.test.tsx
```diff
@@ -0,0 +1,43 @@
...
+      />,
+    )
+
+    const secondEmbed = getEmbed(container)
+    expect(secondEmbed).not.toBeNull()
+    expect(secondEmbed?.getAttribute('src')).toContain('signed.pdf')
+    // A new DOM node — the `key={url}` on <embed> forces React to unmount the
+    // previous element so the browser's PDF plugin picks up the new src.
+    expect(secondEmbed).not.toBe(firstEmbed)
+  })
```
serikjensen, line 35: Nit: might be worth verifying that this test fails without this change, this is the sort of test that claude implements without verifying that

serikjensen (approved):

src/components/Common/DocumentViewer/DocumentViewer.test.tsx
dmortal: Verified before opening — I stashed only `DocumentViewer.tsx` (the `key={url}` change), reran `npm run test -- --run src/components/Common/DocumentViewer/DocumentViewer.test.tsx -t "remounts the embed"`, watched it fail, then restored. PR description includes the same note under Testing. Happy to add the verification step to the test description if useful.

PR #1962, opened 2026-05-28
fix: remove Text wrapper from time off balance cell

aaronlee777:
## Summary

The balance cell in the time off policy detail employee table wrapped its value in `<Text as="span">`, which applied default Text styles inside a table cell where the DataView already controls typography.

## Changes

- Render the raw `balance` value (or `'-'`) directly in the cell instead of wrapping it in `<Text as="span">`
- Drop the now-unused `Text` from `useComponentContext()` and from the `useMemo` deps

## Testing

- Open Storybook: `npm run storybook` → Company/TimeOff/TimeOffPolicyDetail
- Confirm the Balance column renders values (and `-` for empty) with the table's default text styling


krisxcrash (approved):
🚀

PR #1965, opened 2026-05-28
fix: clear stale compensation field errors when FLSA change disables them

serikjensen:
## Summary

- When `useCompensationForm` forces a field value in response to an FLSA change and marks the field disabled (`rate`, `paymentUnit`) or hides it (`adjustForMinimumWage`, `minimumWageId`), the prior validation error on that field persisted in form state. This left stale messages visible on disabled fields (e.g. "FLSA Exempt employees must meet salary threshold" lingering on the disabled Wage field after switching to Commission Only) or invisibly blocked the next submit.
- Call `clearErrors` alongside the existing `setValue` calls in the two FLSA-driven `useEffect` blocks so errors do not outlive the condition that produced them. No public API change.

## Repro

1. Open an existing compensation in Edit mode with FLSA = Exempt.
2. Enter a wage below the FLSA salary threshold ($35,568/yr) and Save → validation fires `RATE_EXEMPT_THRESHOLD` on the Wage field.
3. Change the FLSA dropdown to "Commission Only/No Overtime". The hook forces wage to $0 and disables the field, but the red threshold error remains stuck under the now-disabled input.

## Test plan

- [x] New unit test: stale `RATE_EXEMPT_THRESHOLD` clears when FLSA changes to Commission Only
- [x] New unit test: stale `PAYMENT_UNIT_OWNER` clears when FLSA changes to a value that hook-forces `paymentUnit`
- [x] New unit test: stale `REQUIRED` on `minimumWageId` clears when FLSA leaves Nonexempt (min-wage gate closes)
- [x] All 138 tests in `src/components/Employee/Compensation/` pass with no regressions


src/components/Employee/Compensation/shared/useCompensationForm/useCompensationForm.tsx
```diff
@@ -408,26 +408,33 @@ export function useCompensationForm({
...
+      clearErrors('paymentUnit')
     } else if (
       watchedFlsaStatus === FlsaStatus.COMMISSION_ONLY_NONEXEMPT ||
       watchedFlsaStatus === FlsaStatus.COMMISSION_ONLY_EXEMPT
     ) {
       setValue('paymentUnit', PAY_PERIODS.YEAR)
       setValue('rate', 0)
+      clearErrors(['paymentUnit', 'rate'])
     } else {
       setValue('paymentUnit', resolvedDefaults.paymentUnit)
```
mariechatfield, line 434: Why should this branch _not_ clear errors?

serikjensen: Good catch! updated and added test coverage for the case this wasn't capturing

mariechatfield (approved):
Thanks for making that change, it fixed the only error I found (switching from salary/no overtime to salary/overtime now clears the error on the wage if it's too low)

PR #1967, opened 2026-05-28
chore(SDK-970): lint tsdoc coverage and quality

mariechatfield:
## Summary

Add custom eslint rules to enforce TSDoc coverage and quality within our source code.

## Changes

- Three custom lint rules enforce coverage: 
    - `tsdoc-coverage/require-comment`: every exported symbol needs a TSDoc block
    - `tsdoc-coverage/require-release-tag`: every TSDoc block needs `@public`, `@beta`, `@alpha`, or `@internal`
    - `tsdoc-coverage/sort-tags`: autofixes tag groups into canonical order: description → `@remarks` → other tags →
  `@example`

- `eslint-plugin-jsdoc` was not used, despite the reference in the RFC.
    - Its `require-jsdoc` rule operates on JSDoc conventions and needed overrides to validate types and interfaces.
    - The `export { foo }` re-export pattern also requires resolving the local declaration to check the comment, which `jsdoc/require-jsdoc` doesn't do.
    - It also didn't have support for any of our more custom rules like "require at least one release tag" or "enforce a sort order." One other plugin (`eslint-plugin-tsdoc-require-2`) looked promising but it's at less than 200 downloads a week and doesn't have the "at least one release tag" functionality either, so we'd end up writing custom anyway.

- Rules are currently configured at warn severity with an empty allowlist. Files are promoted to error severity by adding them to the allowlist — the path for SDK-971 migration work.
- Tests for all three rules using `@typescript-eslint/rule-tester`.

## Related

- Jira ticket: [SDK-970](https://gustohq.atlassian.net/browse/SDK-970)

[SDK-970]: https://gustohq.atlassian.net/browse/SDK-970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


eslint-rules/no-ui-component-defaults.js
mariechatfield, line 1: This rule was defined but never used; just delete it to avoid confusion.

src/helpers/federalEin.ts
```diff
@@ -2,6 +2,7 @@ import { useTranslation } from 'react-i18next'
 
 /**
  * Normalizes string input to a 9 digit EIN format with two leading digits, a dash, and 7 trailing digits removing any non-numeric characters.
+ *
```
mariechatfield, line 5: New line added by the sort tags rule, to break out the description from the tags

src/i18n/I18n.ts
mariechatfield, line 1: The `src/i18n/` directory is the only source code directory with all the rules being enforced. You can see where I had to add tsdoc comments with a release tag to get this file to pass lint. The other files only had the sorting rules applied, as part of the general "all comments must have valid syntax" pass

src/i18n/I18n.ts
```diff
@@ -4,6 +4,7 @@ import type { CustomTypeOptions } from 'i18next'
 import { LRUCache } from '@/helpers/LRUCache'
 import type { ResourceDictionary, Resources } from '@/types/Helpers'
 
+/** @internal */
 export const defaultNS = 'common'
 
 const translationLoaders = import.meta.glob<Record<string, string>>('./*/*.json', {
```
mariechatfield, line 10: Note that this did _not_ require a comment, since it is never exported. If you add `export { translationLoaders }` below, you will see an eslint error here at the declaration

Also note that `loadResource` below formatted its comment but did _not_ require adding an `@internal` release tag. That's because we only require a tag for any exported values. If you export `loadResource`, you'll see the rule fail until you add a release tag

eslint-rules/tsdoc-require-comment.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Requires a TSDoc comment on every exported symbol.
+ *
+ * Covers:
+ * - export function/class/interface/type/enum/const declarations
+ * - export { foo } re-exports of locally-declared symbols
+ * - export default declarations and expressions
+ * - function overload deduplication (comment on first signature suffices)
+ *
+ * Use alongside tsdoc/syntax to validate comment syntax.
+ */
```
serikjensen, line 11: So basically if it is exported we require a comment? assume that works for

function MyComponent ({ someProp }) {...}

export MyComponent

serikjensen (approved):
Looking good! small question

eslint-rules/tsdoc-require-comment.ts
mariechatfield: Exactly! Here's the test for that case:

https://github.com/Gusto/embedded-react-sdk/pull/1967/changes#diff-e902ba4248148d7f9c0a0aacb8c64311946794798973aef8ff1b4fc4ff1e7d4aR80-R84
```ts
    // export { foo } with no comment on the declaration
    {
      code: `function helper(): void {}\nexport { helper }`,
      errors: [{ messageId: 'missingTSDoc' }],
    },
```

PR #1973, opened 2026-05-29
fix(SDK-937): switch Amount helper text when toggling Percentage / Fixed

dmortal:
## Summary

In the Custom Deduction form, the helper text under "Amount to withhold" stayed on the percentage copy ("Enter the percentage of your employee's wages to withhold.") even after the user selected **Fixed dollar amount**. The actual submission was correct (the zod schema's \`coerceStringBoolean\` preprocesses on validate), so this was a UI-only confusion.

Root cause: the radio's options are wired through \`withOptions\` with string \`'true'\` / \`'false'\` values. \`useStringifyGenericFieldValue\`'s lookup map keys by the stringified option value, so when the user clicks Fixed-dollar, the form state ends up holding the **string** \`'false'\`. The conditional \`watchedDeductAsPercentage ? percentText : fixedText\` then evaluates the non-empty string as truthy and sticks on percentage copy.

Coerce the watched value explicitly: treat it as \`boolean | 'true' | 'false' | undefined\` and only branch into the percentage copy when the raw value is literally \`true\` or \`'true'\`.

## Changes

- \`StandardDeductionForm.tsx\` — narrow the typed shape of \`useWatch\`'s return value and compute \`watchedDeductAsPercentage\` as a strict boolean. Added a comment explaining why this is necessary (the option-value/form-state round-trip).
- \`DeductionsForm.test.tsx\` (new) — regression test rendering the full form, toggling between Percentage / Fixed dollar amount, and asserting the helper text flips each way. Verified to fail on \`main\` (description sticks on percentage copy).

Verified live via Playwright on \`sdk-app:demo\`: the helper text and the \`$\` / \`%\` adornment both flip correctly.

## Out of scope

The wider asymmetry (form-state holds the option-value string, schema output is boolean) is a real design seam — \`FieldMetadataWithOptions.options\` is typed \`value: string\` and the radio chain stringifies through. Genericizing \`withOptions\` / \`FieldMetadataWithOptions\` to carry the actual entry type would fix this everywhere but touches many call sites. Tracking as a separate cleanup; this PR keeps the change surgical.

## Related

- Jira: [SDK-937](https://gustohq.atlassian.net/browse/SDK-937)

## Testing

- \`npm run test -- --run src/components/Employee/Deductions\` → 34/34 passed (33 prior + 1 new)
- Stashed only \`StandardDeductionForm.tsx\` and re-ran the new test → fails on \`main\`, confirming the test guards the regression.
- Manual via Playwright on \`sdk-app:demo\`: Custom deduction → Fixed dollar amount → helper text flips + \`$\` adornment appears; back to Percentage → percentage copy restored + \`%\` adornment appears.

[SDK-937]: https://gustohq.atlassian.net/browse/SDK-937?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Employee/Deductions/DeductionsForm/StandardDeductionForm.tsx
```diff
@@ -89,10 +89,17 @@ function ReadyForm({
...
+  // The RadioGroup's options carry string values (`'true'`/`'false'`), which
+  // round-trip into form state as strings — only `coerceStringBoolean` in the
+  // zod preprocessor turns them into actual booleans at validation time. So
+  // here we explicitly compare against both shapes; `Boolean('false')` would
+  // be truthy and surface the wrong copy under the Amount field.
+  const watchedDeductAsPercentageRaw = useWatch({
     control: form.form.hookFormInternals.formMethods.control as Control<DeductionFormData>,
     name: 'deductAsPercentage',
-  })
+  }) as boolean | 'true' | 'false' | undefined
```
serikjensen, line 100: Seems like some TS anti patterns happening here. Would be good to get this straightened out properly. Looks like the check below resolves it anyway but i'm still confused why the cast to boolean | 'true' | 'false' | undefined

serikjensen (approved):

PR #1975, opened 2026-05-29
fix(dashboard): show compensation.title (not job.title) on job & pay row

serikjensen:
## Summary

Renaming a secondary or pending job through the future-comp edit form persisted the new title on `compensation.title` but left `job.title` — a denormalized snapshot of the primary comp's title — stale until the comp's effective date. The dashboard Job & Pay row, the delete confirmation, the pending-change alert, the review modal, and the onboarding jobs list all read `job.title`, so a renamed job appeared with its original title until the future comp activated.

Reading title from the comp pointed to by `currentCompensationUuid` makes the rename surface immediately across every display surface, regardless of whether the job is in effect today or still pending.

## What changed

- **Display surfaces** read `currentComp.title` instead of `job.title`:
  - `JobAndPayView.tsx` — Job & Pay row and the delete confirmation modal.
  - `getPendingCompensationChanges.ts` — `change.jobTitle` (used by the pending-change alert label and the `PendingChangesReviewModal` heading). Internal title-diff helpers drop the `job.title` fallback; the `futureTitle &&` guard still suppresses spurious title changes on pay-only edits.
  - `JobsListPresentation.tsx` — onboarding jobs list row.
- **Form prefills** drop the `job.title` fallback so an in-place edit can't silently re-introduce a stale title:
  - `EditCompensation.tsx` `defaultValues.title` and `useCompensationForm` `resolvedDefaults.title` seed directly from the loaded compensation.
- **`EditPendingCompensation.tsx`** reverts an earlier exploration that routed title edits to PUT /v1/jobs for not-yet-started jobs. Title is owned by `useCompensationForm` on this surface (consistent with `EditCompensation`), so PUT /v1/compensations remains the single write path. The now-dead `withTitleField` knob on `useCompensationForm` is removed.
- **Test fixtures and the jobs/compensations factory** set `title` on compensations, matching how the API stores it.

`useJobForm`'s `withTitleField` knob is unchanged — the job form still owns title on standalone job-creation flows, where PUT /v1/jobs is the only endpoint hit.

## Test plan

- [x] `npm run test -- --run` — 2881 passed + 1 pre-existing expected fail.
- [x] `npm run lint:check` — 0 errors (26 pre-existing warnings).
- [x] `npm run format:check` — clean.
- [ ] Manual: rename a future-dated secondary job's title from the dashboard's edit form and confirm the row updates immediately without a refresh.
- [ ] Manual: schedule a future title change on an active primary job and confirm the alert/review modal heading reflects the current title.
- [ ] Manual: delete a renamed secondary and confirm the confirmation modal references the current title.


jeffredodd (approved):
LGTM

PR #1984, opened 2026-06-01
feat!: flsa-aware alerts and aligned wage labels in compensation forms

aaronlee777:
## Summary

When a user selects a particular FLSA status on the compensation edit / add screens, the form now shows context-appropriate alerts and hides inputs that aren't meaningful for that status. The visibility logic lives entirely in `useCompensationForm`, so any partner consuming the hook (not just our screens) gets the same UX automatically. Also aligns the wage / wage-frequency input labels across the Add and Management screens.

### FLSA-driven hook surface

`useCompensationForm` exposes three new reactive boolean flags on `status` (mirroring the existing `status.willDeleteSecondaryJobs` precedent):

- `showCommissionFederalMinimumPayAlert` — `true` when FLSA is `Commission Only Exempt`. Render the federal-minimum-pay warning.
- `showCommissionMinimumWageAlert` — `true` when FLSA is `Commission Only Nonexempt`. Render the local-minimum-wage warning (with support-article link).
- `showOwnerSalaryAlert` — `true` when FLSA is `Owner`. Render the IRS reasonable-salary reminder.

`Fields.Rate` and `Fields.PaymentUnit` are now exposed as `Field | undefined`, with both set to `undefined` when FLSA is either commission-only status. `fieldsMetadata.rate.isDisabled` / `paymentUnit.isDisabled` are left untouched so partners reading metadata directly still get the correct interaction signal.

### Screens

`ManagementCompensationFormBody` and `AddCompensationFormBody` both:

- Render the three new alerts gated on the corresponding `status.show*Alert` flags.
- Use the standard `{CompFields.X && <CompFields.X ... />}` pattern for `Rate` / `PaymentUnit`.
- No longer reach into the form's `useWatch` directly — they read the hook's published flags instead.

### Label alignment (breaking i18n change)

Consolidates the wage / wage-frequency input labels under top-level `wageLabel: "Wage"` and `wageFrequencyLabel: "Wage frequency"` i18n keys. The Add form previously rendered "Compensation amount"; it now matches the Management form's "Wage".

### Docs

`docs/hooks/useCompensationForm.md` updated to document the three new status flags, the conditional availability of `Fields.Rate` / `Fields.PaymentUnit`, and the realigned label keys in the usage example.

## BREAKING CHANGE

The following i18n keys in the `Employee.Compensation` namespace have been removed. Partners who override these via the `dictionary` prop on `GustoProvider` must migrate to the new keys:

| Removed key                       | Use instead                       |
| --------------------------------- | --------------------------------- |
| `amount`                          | `wageLabel`                       |
| `paymentUnitLabel`                | `wageFrequencyLabel`              |
| `management.wageLabel`            | `wageLabel`                       |
| `management.wageFrequencyLabel`   | `wageFrequencyLabel`              |

If you have not overridden any of these keys, no action is required — the SDK ships with the new keys populated.

## Testing

- Open `EditCompensation` / `EditPendingCompensation` / `AddAnotherJob` / onboarding `Compensation`.
- Cycle FLSA through each status:
  - `Commission Only/No Overtime`: federal-minimum-pay warning; wage and wage-frequency inputs hidden.
  - `Commission Only/Eligible for overtime`: minimum-wage warning with support link; wage and wage-frequency inputs hidden.
  - `Owner's draw`: IRS reasonable-salary info alert; wage-frequency input remains rendered but disabled and locked to `Paycheck`.
  - Other statuses: no contextual alerts; wage and wage-frequency render normally.
- Submit each commission-only state — request body carries `rate: "0"`, `payment_unit: "Year"` (hook-forced values).
- `npm run test -- --run src/components/Employee/Compensation` — 147 / 147 passing.

### DIFF
--- a/docs/hooks/useCompensationForm.md
+++ b/docs/hooks/useCompensationForm.md
@@ -122,6 +122,7 @@ The hook returns a discriminated union on `isLoading`.
     willDeleteSecondaryJobs: boolean              // see "Derived helpers" below
     showCommissionFederalMinimumPayAlert: boolean // see "Derived helpers" below
     showCommissionMinimumWageAlert: boolean       // see "Derived helpers" below
+    showOwnerSalaryAlert: boolean                 // see "Derived helpers" below
   }
   actions: {
     onSubmit: (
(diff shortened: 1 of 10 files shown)


serikjensen (changes requested):
This one will need to be solved at the deeper hooks level so that it can be generalizable across all forms consuming hooks, and usable by partners who are consuming the hooks

I'm going to do a follow up with that update

serikjensen (approved):

PR #1997, opened 2026-06-03
fix(SDK-944): show deduction cap fields consistently across create and edit

dmortal:
## Summary

The custom deduction form showed a different set of optional cap fields ("Total amount owed" / "Annual maximum") when **editing** vs **adding** a deduction, and toggling frequency changed which fields appeared inconsistently.

**Root cause:** the cap fields are gated on the watched `recurring` value in `useDeductionForm`. The frequency RadioGroup round-trips its value as the **strings** `'true'`/`'false'` (only the zod preprocessor coerces to a real boolean at validation time), and the string `'false'` is **truthy**. Reading the raw watched value left `isRecurring` wrongly truthy after any frequency toggle, so the caps leaked into one-time deductions — and appeared inconsistently between create (default `recurring: true`) and edit (loads the saved boolean `false`).

## Changes

- `useDeductionForm.tsx`: coerce the watched `recurring` value against both the boolean and string shapes (`=== true || === 'true'`), mirroring the existing `deductAsPercentage` handling in `StandardDeductionForm`. Caps now show iff the deduction is genuinely recurring — the same fields for a given frequency in both create and edit.
- `useDeductionForm.test.tsx`: regression test for the RadioGroup string-`'false'` shape (the boolean case was already covered), plus an edit-mode one-time test asserting caps are hidden on load (matching create-mode visibility).

## Testing

- `npm run test -- --run src/components/Employee/Deductions/shared/useDeductionForm/useDeductionForm.test.tsx` — 9 passed.
- `npm run tsc`, eslint, prettier — clean.

## Related

- Jira: SDK-944
- Parent: SDK-517 — Employee Steady State Management Implementation - Milestone 1
- Source: Test Fest — Employee Dashboard (2026-05-21)


src/components/Employee/Deductions/shared/useDeductionForm/useDeductionForm.tsx
```diff
@@ -194,8 +194,18 @@ export function useDeductionForm({
...
+  // validation time). The string `'false'` is truthy, so reading the raw
+  // watched value here would leave the cap fields visible for one-time
+  // deductions after a frequency toggle. Compare against both shapes — mirrors
+  // the `deductAsPercentage` handling in StandardDeductionForm.
+  const watchedRecurring = useWatch({ control: formMethods.control, name: 'recurring' }) as
+    | boolean
+    | 'true'
+    | 'false'
+    | undefined
+  const isRecurring = watchedRecurring === true || watchedRecurring === 'true'
```
serikjensen, line 208: Do we know if we could fix something to get the boolean to properly work as the value rather than needing to provide this workaround? Wondering if this is actually the symptom of another issu

serikjensen (approved):

PR #2019, opened 2026-06-04
build: add eslint rule to require comments on interfaces

mariechatfield:
## Summary

`@public` (or alpha/beta) interfaces that are exported and documented need to require a short description on every member.

I had to keep re-instructing the agents and skills to document interface members, so I added an eslint rule to enforce it for me.


src/partner-hook-utils/form/composeSubmitHandler.ts
mariechatfield, line 1: Only error that slipped past my manual review

jeffredodd (approved):

serikjensen (approved):

PR #2040, opened 2026-06-05
fix(SDK-971): extend require-member-comment rule to cover object type aliases

mariechatfield:
## Summary

The `tsdoc-coverage/require-member-comment` ESLint rule previously only checked members of `interface` declarations. Type aliases with plain object bodies (`type Foo = { ... }`) expose the same public API surface but were silently skipped, allowing undocumented members to pass lint undetected.

This PR extends the rule to also check `TSTypeAliasDeclaration` nodes whose type annotation is a `TSTypeLiteral` (i.e. a plain object literal — not unions, intersections, or mapped types).

### Changes

- **`eslint-rules/tsdoc-require-member-comment.ts`** — `checkMembers` now handles `type Foo = { ... }` in addition to `interface Foo { ... }`; updated message from "interface member" to "type member"
- **`eslint-rules/tsdoc-require-member-comment.test.ts`** — 5 new test cases covering the type alias path (all documented, `@internal`, non-object alias, empty object, re-export)
- **`src/partner-hook-utils/composeErrorHandler.ts`** — member comments added to `SubmitStateForErrorHandling` (was undetected before this fix)
- **`src/partner-hook-utils/types.ts`** — member comments added to `FormHookResult` (was undetected before this fix)

## Related

- Jira ticket: [SDK-971](https://gustohq.atlassian.net/browse/SDK-971)

[SDK-971]: https://gustohq.atlassian.net/browse/SDK-971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jeffredodd (approved):
LGTM 🚢

PR #2047, opened 2026-06-05
test: default container-breakpoints mock to desktop layout

serikjensen:
## Summary

- Flips the global `useContainerBreakpoints` test mock in [src/test/setup.ts](src/test/setup.ts) to return the full desktop breakpoint set (`['base','small','medium','large']`) by default, and re-applies it in the global `beforeEach`. This means any test rendering a `DataView`/table now gets the desktop table layout automatically — no per-test mock boilerplate.
- Re-applying the default in `beforeEach` (rather than relying on `mockClear()`, which doesn't reset a `mockReturnValue`) guarantees a test that opts into the mobile layout via `mockUseContainerBreakpoints.mockReturnValue(['base'])` can't leak that override into later tests in the same file.
- Removes the ~16 duplicated per-file `vi.mock('@/hooks/useContainerBreakpoints/...')` blocks and ~10 redundant desktop `mockReturnValue`/`vi.mocked(...)` overrides (including bare auto-mocks) that previously had to be copy-pasted into every test rendering a data view.
- Restores the mobile/card layout explicitly in the 8 files that genuinely depend on it (query `data-card`/`data-cards`/`role=list`/tab dropdown), via a single `mockUseContainerBreakpoints.mockReturnValue(['base'])` opt-in.

Net effect: write a test with a data view and it just works against the table layout; only reach for the `['base']` opt-in when you specifically want the mobile card view.

## Test plan

- [x] `npm run test -- --run` — 255 files / 3075 passed, 1 expected fail
- [x] `npm run lint:check` — 0 errors (pre-existing warnings only)


jeffredodd (approved):
Love it! Thanks!

PR #2048, opened 2026-06-05
refactor: standardize management card loading and normalize useCompensationManagement return shape

serikjensen:
## Summary

Track 1 of the dashboard-block cleanup: make standalone management card loading states consistent and normalize a data-hook return shape.

- **Standardize card loading to the partial pattern.** `PaymentMethodCard`, `PaystubsCard`, and `DocumentsCard` previously short-circuited to a full-screen `<BaseLayout isLoading />`. They now render the card chrome (`Components.Box` + `Components.BoxHeader`) with an inline `<Loading />` in the body while data is in flight, and disable the header action while loading — matching the 7 cards that already do partial loading (FederalTaxes, StateTaxes, Profile, Deductions, etc.). Error handling stays on `<BaseLayout error={...}>`.
- **Normalize `useCompensationManagement`.** It now returns the standard `HookLoadingResult | BaseHookReady<...>` discriminated union (`src/partner-hook-utils/types.ts`), returning `{ isLoading: true, errorHandling }` while the jobs query is pending instead of always `isLoading: false` with a `status.isCompensationLoading` flag. This matches the reference hooks `useEmployeeList` and `usePaymentMethodList`. `CompensationCard` now branches on the discriminant (still rendering partial loading).
- **Verified (no changes needed)** that the thin summary hooks already return the standard union: `useEmployeeProfileSummary`, `useHomeAddressSummary`, `useEmployeeWorkAddressSummary`, `useFederalTaxesSummary`, `useStateTaxesSummary`, `useDocumentsList`.

No event names, i18n namespaces, docs, or other features were touched.

## Test plan

- [x] `npm run test -- --run` for the three card tests + compensation hook/card tests (26 passed). Each card gains a test asserting the chrome (title) + inline loading render while the request is pending (held open with `delay('infinite')`); the compensation hook tests branch on the `isLoading` discriminant via an `assertReady` helper.
- [x] `npm run tsc` (clean)
- [x] `npm run lint:check` (no new errors/warnings on changed files)
- [x] `npm run format:check` (clean)


serikjensen (commented):
@cursoragent update this one to remove these tests. Asserting on loading states like this is brittle

krisxcrash (approved):

PR #1996, opened 2026-06-03
docs(SDK-971): tsdoc models and internal hooks

mariechatfield:
### Phase 1 — Internal easy wins (all parallel)

Mark everything `@internal`. No reference docs needed.

| Directory | Violations | Notes |
|-----------|-----------|-------|
| `src/contexts/ApiProvider/` | 3 | React Query + auth wiring |
| `src/contexts/LocaleProvider/` | 2 | i18n context |
| `src/contexts/LoadingIndicatorProvider/` | 2 | loading indicator context |
| `src/components/Flow/` | 3 | flow orchestration base |
| `src/components/Base/` | 4 | base component abstractions |
| `src/models/` | 6 | internal data models |
| `src/contexts/ObservabilityProvider/` | 5 | wraps partner ObservabilityHook; the hook interface itself is public and documented in Phase 3 (src/types/) |
| `src/hooks/` | 12 | useClientPagination, useContainerBreakpoints, useDateRangeFilter, useDebounce, useForkRef, useOverflowDetection, usePagination |

**Total: 37 violations, all parallel.**

## Changes

Each commit represents running `/tsdoc-directory <directory> and commit` in claude, using the new agent skills. This represents the first round of backfilling docs according to the docs audit and plan in [Notion](https://app.notion.com/p/374ad673c6c280dd8437d204f11e8470#84f7d23ae8fd4c8c8991b023f95bf348).

## Related

- Jira ticket: [SDK-971](https://gustohq.atlassian.net/browse/SDK-971)

[SDK-971]: https://gustohq.atlassian.net/browse/SDK-971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2064, opened 2026-06-08
docs(SDK-1011): backfill phase 6

mariechatfield:
### Directories

| Directory | Reference Docs | Notes |
| --- | --- | --- |
| `src/contexts/GustoProvider/` | `docs/integration-guide/request-interceptors.md` | GustoProviderProps, APIConfig, GustoProviderCustomUIAdapterProps. SDKHooks types come from `src/types/` (Phase 3) — document re-exports only. |
| `src/contexts/ThemeProvider/` | `docs/theming/theme-variables.md`, `docs/theming/theming-guide.md` | GustoSDKTheme, GustoSDKThemeColors. |

### Changes

**Type refactoring (pre-req for ThemeProvider docs):**
- Introduced explicit `GustoSDKThemeColors` and `GustoSDKTheme` interfaces in `src/contexts/ThemeProvider/types.ts` — fully required, with per-token TSDoc and `@defaultValue` tags. Previously both were inferred (`Partial<typeof const>` / `Partial<ReturnType<fn>>`).
- `createTheme` now merges with defaults internally so computed tokens are never `undefined`.
- Call sites that accept partner overrides now explicitly use `Partial<GustoSDKTheme>`.

**Documentation:**
- `src/contexts/GustoProvider/` — 26 violations resolved across 3 files
- `src/contexts/ThemeProvider/` — 13 violations resolved across 4 files; removed dead `render` re-export from barrel

Reference PR: https://github.com/Gusto/embedded-react-sdk/pull/1996

### DIFF
--- a/src/contexts/ThemeProvider/theme.ts
+++ b/src/contexts/ThemeProvider/theme.ts
@@ -44,6 +44,7 @@ const baseColors = {
   },
 }
 
+/** @internal */
 export const transitionDuration = 200
 
 const defaultThemeColors: GustoSDKThemeColors = {
@@ -74,6 +75,7 @@ const defaultThemeColors: GustoSDKThemeColors = {
   colorButtonIcon: baseColors.neutral[400],
 }
 
+/** @internal */
 export const createTheme = (colors: Partial<GustoSDKThemeColors> = {}): GustoSDKTheme => {
   const mergedColors = { ...defaultThemeColors, ...colors }
 
@@ -126,6 +128,7 @@ export const createTheme = (colors: Partial<GustoSDKThemeColors> = {}): GustoSDK
   }
 }
 
+/** @internal */
 export const mergePartnerTheme = (partnerTheme: Partial<GustoSDKTheme>): GustoSDKTheme => {
   const colors = Object.entries(defaultThemeColors).reduce(
     (acc, [key, value]) => {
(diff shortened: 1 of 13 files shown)


src/contexts/ThemeProvider/theme.ts
```diff
@@ -120,34 +119,28 @@ export const createTheme = (colors: GustoSDKThemeColors = {}) => {
...
-    (acc: GustoSDKThemeColors, [key, value]) => {
-      acc[key as keyof typeof defaultThemeColors] =
-        partnerTheme[key as keyof typeof defaultThemeColors] || value
+    (acc, [key, value]) => {
+      acc[key as keyof GustoSDKThemeColors] =
+        partnerTheme[key as keyof GustoSDKThemeColors] || value
       return acc
     },
-    {},
+    { ...defaultThemeColors },
```
mariechatfield, line 139: I did actually modify the code here a little bit, to make the types more robust and reliable. I believe the final computation for what colors are chosen is the same. We still only pass the color values from `partnerTheme` to `createTheme` with overrides to the default theme colors built in, and `createTheme` does one more pass of merging the default with the provided colors in case it's called directly

jeffredodd (approved):
LGTM and thanks for the type fixes.

PR #2076, opened 2026-06-09
docs: add versioned documentation support (SDK-766)

jeffredodd:
## What partners get

Docs versioning. Once the first snapshot is bootstrapped in `Gusto/embedded-sdk-docs`, `sdk.gusto.com` serves two flavors of every page:

| URL | Content |
|---|---|
| `/docs/<page>` | The latest released minor — `0.47` after first release post-merge |
| `/docs/next/<page>` | Live in-development content (the `docs/` directory at `main`) |

A version dropdown lives in the navbar between Docs and GitHub.

JIRA: [SDK-766](https://gustohq.atlassian.net/browse/SDK-766) · Epic: [SDK-758](https://gustohq.atlassian.net/browse/SDK-758) (Modernize SDK Documentation, Phase 3c)

## Where versioned content lives

**Not in this repo.** This was the key shift from earlier iterations of this PR, driven by [Marie's review](https://github.com/Gusto/embedded-react-sdk/pull/2076#discussion_r3406382132): keeping `versioned_docs/` here would (a) flood the source tree with stale copies of docs LLMs would happily index alongside current ones, and (b) make every patch release add 20k+ lines of "old docs" churn to PRs. So:

- **This repo** carries only `docs/` (live content). No `versioned_docs/`, no `versions.json`, no `versioned_sidebars/`, no snapshot script.
- **`Gusto/embedded-sdk-docs`** accumulates `versioned_docs/version-X.Y/`, `versions.json`, and `versioned_sidebars/` as releases land. The Sync Docs Source workflow creates and maintains them there.

### What docs in this repo look like

```
embedded-react-sdk/
├── docs/                      ← live, in-development docs (only)
│   ├── component-adapter/
│   ├── getting-started/
│   ├── hooks/
│   └── ...
├── docs-site/                 ← Docusaurus source: config, sidebars, theme
│   ├── docusaurus.config.ts   ← versioning conditional on versions.json
│   ├── sidebars.ts
│   ├── src/
│   └── package.json           ← NO versioned_docs/, NO versions.json
└── .github/workflows/publish-docs.yaml  ← does all versioning work downstream
```

No `versioned_docs/`, no snapshot script, no `docs:snapshot` / `docs:version` npm scripts. Locally:
- `npm run docs` → dev server, single version, no dropdown
- `npm run docs:build` → builds clean, single version

### What docs in `embedded-sdk-docs` look like

```
embedded-sdk-docs/
├── docs/                      ← synced from source on every workflow run
├── docs-site/                 ← synced from source EXCEPT for:
│   ├── versioned_docs/        ← accumulates downstream over releases
│   │   ├── version-0.47/      ← (created on first release post-merge)
│   │   ├── version-0.48/      ← (created when 0.48.0 ships)
│   │   └── ...
│   ├── versioned_sidebars/    ← accumulates downstream
│   └── versions.json          ← ["0.48", "0.47", ...] — accumulates
├── .buildkite/                ← downstream's own pipeline
└── .nvmrc                     ← synced from source
```

### Why the same `docusaurus.config.ts` works in both places

It reads `versions.json` at config-load time:
- **Absent** (this repo) → versioning disabled, no dropdown, no `/next/` route
- **Present** (downstream) → versioning active, `lastVersion = versions[0]` (so the unprefixed `/docs/` automatically points at the freshest minor), dropdown shown

No fork, no config drift, no manual `lastVersion` bumps.

### Path 1 — docs change between releases (push trigger)

```
PR with docs/** or docs-site/** changes merges to main on this repo
  → Sync Docs Source fires (push trigger)
  → workflow:
       - runs typedoc → docs/api/
       - clones embedded-sdk-docs
       - preserves downstream's versioned_docs/, versioned_sidebars/, versions.json
       - wipes + copies fresh docs/ + docs-site/ + .nvmrc in
       - snapshot action = NONE
  → commits + pushes to embedded-sdk-docs/main
  → downstream Buildkite builds + publishes to gh-pages
  → sdk.gusto.com /docs/next/* updated; /docs/0.47/* unchanged
```

### Path 2 — patch release (e.g. 0.47.2)

```
chore: release 0.47.2 merges → Publish to NPM publishes
  → Sync Docs Source fires (workflow_run trigger)
  → workflow:
       - runs typedoc; clones downstream; preserves versioned_*;
         copies fresh source in
       - versions.json has "0.47" → snapshot action = REFRESH
       - walks versioned_docs/version-0.47/, refreshes content of each file
         from the matching path in the freshly-synced docs/
       - files in the snapshot with no live counterpart left as-is
       - files in live docs/ that aren't in the snapshot are NOT added
  → commits + pushes
  → /docs/0.47/* now reflects 0.47.2 content; new docs landed on main since
    0.47.0 (e.g. useEmployeeList) do not propagate to the 0.47 snapshot
```

### Path 3 — minor release (e.g. 0.48.0)

```
chore: release 0.48.0 merges → Publish to NPM publishes
  → Sync Docs Source fires
  → workflow:
       - runs typedoc; clones downstream; preserves any existing versioned_*
         (none for 0.48 yet); copies fresh source in
       - versions.json doesn't have "0.48" → snapshot action = CREATE
       - runs `npx docusaurus docs:version 0.48`:
           - copies current docs/ → versioned_docs/version-0.48/
           - captures sidebar to versioned_sidebars/version-0.48-sidebars.json
           - prepends "0.48" to versions.json
       - lastVersion = versions[0] = "0.48" (config-derived, automatic)
  → commits + pushes
  → /docs/ now resolves to 0.48; /docs/0.47/ still works;
    dropdown shows 0.48, 0.47, Next
```

### Path 4 — bootstrap (first time after this PR merges)

```
This PR merges; downstream's versions.json is empty / missing
Either: the next release fires the workflow,
Or: a maintainer manually triggers Sync Docs Source via workflow_dispatch
  → workflow: same prep as Path 1
       - versions.json doesn't have current minor → snapshot action = CREATE
       - runs `npx docusaurus docs:version <current-minor>`
       - downstream now has versioned_docs/version-X.Y/, versions.json=[X.Y]
  → /docs/ → X.Y; /docs/next/ → main; dropdown active
```

## Workflow safety properties

| Trigger | Snapshot policy |
|---|---|
| `workflow_run` after Publish to NPM | Create if minor is new; refresh content if patch |
| `push` to `docs/**` or `docs-site/**` on `main` | Never touch versioned files; only sync live source |
| `workflow_dispatch` | Create if minor missing in downstream (bootstrap / recover); otherwise none |

- **Doc fixes don't wait for a release.** Push trigger redeploys `/docs/next/` automatically.
- **Manual dispatch is safe.** Only creates a snapshot when one is missing; never overwrites.
- **Defense-in-depth allowlist.** The workflow refuses to push anything outside `^(docs/|docs-site/|\.nvmrc$)` to downstream — guards against future edits accidentally widening the touch radius.

## Editing docs in a released version

- **Typo / clarification that should reach a previous minor**: edit `docs/` here. The next patch release of that minor refreshes the file's content in the downstream snapshot.
- **Add / remove / rename a doc in an already-released minor**: edit `versioned_docs/version-X.Y/` directly in `Gusto/embedded-sdk-docs`. Patch refresh is content-only — never adds new files to a versioned snapshot, never removes orphaned ones. Update `versioned_sidebars/version-X.Y-sidebars.json` downstream too if navigation changes.

This split keeps released-minor structure stable: a new live doc for a future feature can't accidentally appear in a partner's `0.47` view.

## Files to focus on while reviewing

**The actual versioning wiring:**
- `docs-site/docusaurus.config.ts` — conditional on `versions.json` presence. Same file works in both repos.
- `.github/workflows/publish-docs.yaml` — adds `push` trigger; rewrites the sync step to preserve downstream's versioned files and handle snapshot create/refresh in shell.

**Cleanup:**
- `package.json` — removed `docs:snapshot` and `docs:version` scripts (workflow does this in downstream now).
- `.gitignore`, `eslint.config.ts` — removed entries that referenced source-side versioned dirs / snapshot script.
- `.claude/commands/release.md` — release flow no longer touches docs; workflow handles everything post-merge.

**Process docs:**
- `CONTRIBUTING.md` — new "Documentation versioning" section.

## Design choices worth scrutiny

**Why conditional config instead of two files.** Same `docusaurus.config.ts` deployed to both repos avoids drift. `existsSync('./versions.json')` is a one-line gate at config-load time; `lastVersion` derived from `versions[0]` is auto-correct.

**Why content-only patch refresh.** Adding / removing / renaming a doc in a released minor is an intentional, version-scoped decision; it shouldn't be a side effect of editing live `docs/`. Content-only keeps released-minor structure stable while still letting fixes and regenerated reference data flow into older versions.

**Why workflow-driven instead of release-skill-driven.** Earlier iterations of this PR ran `npm run docs:snapshot` locally as part of `chore: release`. That created two failure modes: (a) the snapshot could go stale or be forgotten, and (b) the source tree carried committed copies of every released minor's docs. Moving the work into the workflow eliminates both.

**Bootstrap on dispatch.** A manual `workflow_dispatch` will create the snapshot if it's missing for the current minor. We can bootstrap the first 0.47 snapshot in downstream without waiting for the next release.

## Verified locally

- [x] `npm run docs:build` — clean (single-version build since `versions.json` is absent in this repo)
- [x] `npm run lint:check`, `npm run format:check`, `npm run tsc` — all clean
- [x] Workflow YAML parses; trigger semantics, snapshot-decision branching, and allowlist defense traced through manually for every event type
- [x] `versions.json` conditional in `docusaurus.config.ts` — verified the disabled path renders correctly (no dropdown, no `/next/` route, no `lastVersion`)

Two items can only be confirmed end-to-end on the next release or via manual dispatch post-merge:
- [ ] Sync Docs Source bootstraps the first snapshot in `embedded-sdk-docs`
- [ ] Subsequent patch releases refresh the snapshot content correctly downstream

## Heads-up for reviewers

`npm run lint:check` shows 26 pre-existing warnings — all `react-hooks/exhaustive-deps` warnings in `src/components/Payroll/*`, `src/hooks/*`, etc. These exist on `main` and are not introduced by this PR. CI doesn't gate on warnings.

## Acceptance criteria (SDK-766)

- [x] Versioned snapshots exist alongside "Next" — bootstrapped downstream by the workflow on first release post-merge or manual dispatch
- [x] Version switcher dropdown works in the navbar — wired in `docusaurus.config.ts`, active when `versions.json` is present
- [x] Team documentation exists — `CONTRIBUTING.md` "Documentation versioning" section
- [x] Integration with release CI — Sync Docs Source workflow handles snapshot creation and content refresh automatically

[SDK-766]: https://gustohq.atlassian.net/browse/SDK-766?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SDK-758]: https://gustohq.atlassian.net/browse/SDK-758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/publish-docs.yaml
+++ b/.github/workflows/publish-docs.yaml
@@ -1,11 +1,42 @@
 name: Sync Docs Source
 
-# Fires after a successful NPM publish on main. Runs TypeDoc to generate
-# docs/api/ markdown, then pushes the Docusaurus source
-# (docs/ + docs-site/ + .nvmrc, including the generated docs/api/ files)
-# to the main branch of Gusto/embedded-sdk-docs. That repo's Buildkite
-# pipeline runs the Docusaurus build and publishes the result to its
-# gh-pages branch.
+# Builds docs source on this repo and pushes it to Gusto/embedded-sdk-docs.
+# The downstream repo's Buildkite pipeline then runs the Docusaurus build
+# and publishes the result to its gh-pages branch.
+#
+# Versioning lives entirely in the downstream repo, not here. This repo
+# only ever carries the live (current) docs in `docs/`. Versioned snapshots
+# (`docs-site/versioned_docs/`, `versions.json`, `versioned_sidebars/`)
+# accumulate in `embedded-sdk-docs` over time as releases land.
+#
+# Three triggers, all funnel through the same job:
+#   1. workflow_run after Publish to NPM — the release path. If the released
+#      minor isn't in downstream's versions.json, creates a new snapshot.
+#      If it is, refreshes the existing snapshot's content from the freshly
+#      synced docs (content-only; structural changes to versioned docs are
+#      explicit edits in the downstream repo).
+#   2. push to main on docs/** or docs-site/** — the live-content path.
+#      Syncs the latest docs source down so /docs/ refreshes. Does not touch
+#      versioned snapshots.
+#   3. workflow_dispatch — manual run. Always allowed; useful for recovery,
+#      end-to-end testing, or bootstrapping the first snapshot on a clean
+#      downstream (it will create the snapshot if missing for the current
+#      package.json minor).
+#
+# Snapshot mechanics:
+#   New minor / bootstrap: `npx docusaurus docs:version <X.Y>` runs inside the
+#   freshly-synced downstream working tree. Docusaurus copies the current
+#   docs/ into versioned_docs/version-<X.Y>/, captures the sidebar, and adds
+#   <X.Y> to versions.json. The conditional config in docs-site/docusaurus.config.ts
+#   then derives lastVersion from versions.json[0], so the unprefixed /docs/
+#   automatically points at the newest snapshot — no manual config change.
+#
+#   Patch refresh: walks versioned_docs/version-<X.Y>/ and refreshes the
+#   content of each file from the matching path in the freshly-synced docs/.
+#   Files in the snapshot that have no live counterpart are left as-is.
+#   Files in live docs/ that aren't already in the snapshot are NOT added —
+#   adding a doc to a released minor is an intentional change made directly
+#   in the downstream repo.
 #
 # Auth: a GitHub App installed on Gusto/embedded-sdk-docs with
 # Contents: write. Credentials live in this repo's `publish-docs`
@@ -18,17 +49,22 @@ name: Sync Docs Source
 #      DOCS_APP_ID and DOCS_APP_PRIVATE_KEY (paste the entire .pem
 #      contents for the private key).
 #   3. Once a manual workflow_dispatch run is green, set repo variable
-#      DOCS_PUBLISH_ENABLED=true to turn on auto-publish on release.
+#      DOCS_PUBLISH_ENABLED=true to turn on the workflow_run + push paths.
 #
 # Manual `workflow_dispatch` triggers fire any time (the gate below only
-# guards the auto-on-release path), so the pipeline can be exercised
-# end-to-end before flipping the variable.
+# guards the auto paths), so the pipeline can be exercised end-to-end
+# before flipping the variable.
 
 on:
   workflow_run:
     workflows: [Publish to NPM]
     types: [completed]
     branches: [main]
+  push:
+    branches: [main]
+    paths:
+      - 'docs/**'
+      - 'docs-site/**'
   workflow_dispatch:
 
 concurrency:
@@ -38,12 +74,15 @@ concurrency:
 jobs:
   sync:
     # workflow_dispatch always allowed (for manual testing). workflow_run
-    # only fires when DOCS_PUBLISH_ENABLED=true, so the next NPM release
-    # won't auto-publish docs until that variable is explicitly set.
+    # and push only fire when DOCS_PUBLISH_ENABLED=true, so neither an NPM
+    # release nor a docs push will auto-publish until that variable is
+    # explicitly set.
     if: |
       github.event_name == 'workflow_dispatch' ||
-      (vars.DOCS_PUBLISH_ENABLED == 'true' &&
-       github.event.workflow_run.conclusion == 'success')
+      (vars.DOCS_PUBLISH_ENABLED == 'true' && (
+        github.event_name == 'push' ||
+        (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success')
+      ))
     runs-on:
       group: gusto-ubuntu-default
     environment: publish-docs
@@ -60,7 +99,13 @@ jobs:
 
       - name: Read SDK version
         id: version
-        run: echo "version=$(node -p "require('./package.json').version")" >> "$GITHUB_OUTPUT"
+        run: |
+          VERSION=$(node -p "require('./package.json').version")
+          MINOR=$(echo "$VERSION" | cut -d. -f1,2)
+          {
+            echo "version=${VERSION}"
+            echo "minor=${MINOR}"
+          } >> "$GITHUB_OUTPUT"
 
       - name: Install SDK dependencies for TypeDoc
         run: npm ci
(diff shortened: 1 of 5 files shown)


docs-site/versioned_docs/version-0.47/component-adapter/component-adapter-faq.md
mariechatfield, line 1: Does this mean that each new minor release we cut will copy the entire directory into source code in this repo? I'm worried that's going to create a lot of surface area for LLMs to use old docs

this feels like something I'd rather keep in the docs repo, so that the docs that are live in our repo are always the latest and greatest

.gitignore
```diff
@@ -145,6 +145,7 @@ docs-site/.docusaurus/
 # Typedoc generated files (run by docusaurus)
 
 docs/api/
+docs-site/versioned_docs/*/api/
```
mariechatfield: Bonus if we save the complete markdown source in the docs directory, we'll also keep the full API reference. if we also git ignore it here, we'll lose the ability to show past versions of typedoc files tagged to the version, right?

docs-site/versioned_docs/version-0.47/component-adapter/component-adapter-faq.md
jeffredodd: Yeah, the goal is going to be that our repo stays 'normal' and that when we do the copy the verisoning is in the docs. Still tweaking bits of it i'll confirm without a doubt thats how this is going to work because I totally agree. We don't want the versions living in our SDK repo, but we wnat that in the docs repo. :)

jeffredodd: Great catch — that concern reshaped the architecture. Versioned content has been moved out of this repo entirely in `a2a6a9e9`.

`versioned_docs/`, `versions.json`, and `versioned_sidebars/` now live exclusively in `Gusto/embedded-sdk-docs`. The Sync Docs Source workflow creates and refreshes them there: on a new minor it runs `npx docusaurus docs:version <X.Y>` inside the downstream working tree; on a patch it walks the existing `versioned_docs/version-<X.Y>/` and content-refreshes from the freshly-synced `docs/`. This repo only ever carries the live (current) docs — so LLMs scanning the source tree never index old copies, and patch releases stop adding 20k+ lines of churn to PRs.

`docs-site/docusaurus.config.ts` is the same file in both repos: it reads `versions.json` at config-load time and disables versioning (no dropdown, no `/next/`) when the file is absent. Local `npm run docs:build` here continues to work as a single-version build.

Full architecture + flow paths are in the rewritten PR description.

.gitignore
jeffredodd: Good point, and moot now since `versioned_docs/*/api/` is no longer a path in this repo — that gitignore line has been removed in `a2a6a9e9`.

The versioned API reference IS captured downstream: the Sync Docs Source workflow runs `npm run typedoc` (generating `docs/api/`) BEFORE the snapshot step, so when `docusaurus docs:version <X.Y>` runs inside the downstream working tree, it copies `docs/api/` into `versioned_docs/version-<X.Y>/api/` along with the rest. Past versions retain their TypeDoc reference frozen at the release tag.

docs-site/versioned_docs/version-0.47/component-adapter/component-adapter-faq.md
mariechatfield: oh hell yeah that was so stinking fast! thanks for revisiting the plan so quickly

.github/workflows/publish-docs.yaml
```diff
@@ -18,17 +49,22 @@ name: Sync Docs Source
...
 on:
   workflow_run:
     workflows: [Publish to NPM]
     types: [completed]
     branches: [main]
+  push:
+    branches: [main]
+    paths:
+      - 'docs/**'
+      - 'docs-site/**'
```
mariechatfield, line 67: 🙏🏻 bless thank you

mariechatfield: this does make me think it might be worth auto-generating the typedoc and committing to the repo, partially so we have an easier document to review and partly so tsdoc changes will also auto-trigger this

jeffredodd: Thats not a bad idea, we might also be able to then see the changes per PR that occur from code which might be kinda neat too. I'm up for whatever we think is the path forward here. :)

mariechatfield (approved):
love it, thank you so much for setting this up!

jeffredodd:
Thanks @mariechatfield, as an FYI I'm going to wait to merge this until Monday in case something goes wrong :)

.github/workflows/publish-docs.yaml
mariechatfield: if we can set up the endpoint inventory script with the new github app tokens, then I'm in favor of doing the same for the typedoc step (on push to a branch, CI pushes a commit with any changes to auto-generated docs no matter the job that runs the docgen)

jeffredodd: Lets sync Monday and see if we can get this working 🎉

jeffredodd:
@claude Can you review this PR and tell me if you see anything that needs addressing? Please use our PR review skill if we have one that works for this task.

jeffredodd:
@cursoragent can you do a code review?

jeffredodd:
Thanks for the review! Addressing the feedback above:

**Critical — unsafe `as string[]` cast** ✅ Fixed in `docs-site/docusaurus.config.ts`. Replaced the cast with an `isStringArray` type guard that throws on shape mismatch. Went with an inline guard rather than Zod since `docs-site/` doesn't currently take a Zod dependency and the file we're parsing is one we generate at build time — proportional to scope.

```ts
function isStringArray(val: unknown): val is string[] {
  return Array.isArray(val) && val.every((v) => typeof v === 'string')
}

const rawVersions: unknown = existsSync(versionsPath)
  ? JSON.parse(readFileSync(versionsPath, 'utf8'))
  : []
if (!isStringArray(rawVersions)) {
  throw new Error('docs-site/versions.json must be a string[]')
}
const versions: string[] = rawVersions
```

Verified the failure path: dropped a stub `versions.json` of `{}` and `npm run build` throws the expected error.

**Important — dependency downgrades & removed prop-type exports** ⚠️ These appear to have been cross-posted from a different PR (the review header references PR #2078, branch `cursor/docs-versioning-bd58`). On this branch the actual diff vs. the merge base only touches 5 files:

- `.claude/commands/release.md`
- `.github/workflows/publish-docs.yaml`
- `CONTRIBUTING.md`
- `docs-site/docusaurus.config.ts`
- `docs-site/package.json`

No changes to `package.json` / `package-lock.json` (the downgrades only appeared because the branch was behind `main`'s dep bumps), and no changes to `src/components/Employee/exports/employeeManagement.ts`, `src/components/Payroll/Dismissal/index.ts`, `src/components/Payroll/index.ts`, or `src/components/TimeOff/index.ts`. I rebased onto current `main` so the stale-branch noise is cleared from the diff.

**Suggestion — `CONTRIBUTING.md` Markdown links example** ✅ Added a good/bad fenced example beneath the existing section.

**Verified locally:** `npm run tsc` clean; `cd docs-site && npm run build` clean; type-guard failure path throws as expected.

PR #2067, opened 2026-06-09
feat(SDK-767): wire SEO metadata, OG defaults, sitemap URL, per-page descriptions

jeffredodd:
## Summary

First of three subject-focused PRs for [SDK-767](https://gustohq.atlassian.net/browse/SDK-767). This PR delivers the SEO surface: production URL, OG/metadata defaults, the sitemap fix, the OG image asset, and the per-page `description:` sweep across the docs.

**Stacked PRs (review/merge in order):**
1. **This PR** — SEO, sitemap, OG, descriptions, test-fests exclusion
2. [PR B](#) — Edit-this-page links _(link added after open)_
3. [PR C](#) — a11y fixes _(link added after open)_

(Replaces #2066, which was the same scope as one omnibus PR.)

## What changed

| File / area | Change |
| --- | --- |
| `docs-site/docusaurus.config.ts` | `url` → `https://sdk.gusto.com` (was localhost); added `themeConfig.image` + `metadata` block; `exclude: ['test-fests/**']` on the docs preset and matching `ignoreFiles` on the search plugin |
| `docs-site/static/img/og-default.png` | New 1200×630 OG image (programmatically-rasterized placeholder, see follow-ups) |
| `docs/**/*.md` (63 files) | Added `description:` front matter (1 sentence each, 120–160 chars, partner-facing voice per CLAUDE.md) |

## Why each change

- **`url` was `http://localhost`** → Docusaurus bakes that into the sitemap, canonical links, and every `og:url` meta tag. Fixing it is the single highest-leverage SEO change in this PR.
- **`themeConfig.metadata` + `image`** → site-wide defaults so any page without a per-page override still gets proper meta tags. Per-page overrides happen via front matter `image:` if a page wants one.
- **Per-page `description:`** → previously every page rendered the same site-tagline as its meta description. Now each page tells crawlers and link-previewers what it specifically covers.
- **`test-fests/` exclusion** → `docs/test-fests/compensation-steady-state-sdk.md` (added in #2036) is internal test-fest notes written in third person about internal QA scope. It shouldn't be on the public docs site.

## Verification

- `cd docs-site && npm run build` — clean (CI's `docs-build` job also enforces no broken links/anchors)
- `build/sitemap.xml` — 64 URLs, all `https://sdk.gusto.com/...` prefix, zero `test-fests` entries
- `build/img/og-default.png` — present
- View source on `/docs/integration-guide`:
  ```
  <meta name="description" content="Wire the SDK into a React app — versioning, event handling, ...">
[...]

PR #2081, opened 2026-06-09
ci(SDK-767): lint docs/**/*.md for title + description front matter

jeffredodd:
## Summary

Stacked on top of #2067. Adds CI enforcement so the `description:` field that PR #2067 just added to every `docs/**/*.md` can't silently regress.

- New: `build/lintDocsFrontmatter.ts` — walks `docs/**/*.md` and fails if any file is missing a non-empty `title:` or `description:` in its YAML front matter.
- New: `npm run docs:lint` — invokes the script via `npx tsx` (same pattern as `endpoints:verify`).
- CI: added a `Validate docs front matter` step to the existing `docs-build` job, ahead of the Docusaurus build, so front-matter mistakes surface cleanly instead of as opaque build failures.

No new runtime deps — uses Node built-ins only.

## Why

Docusaurus silently falls back to the site-wide default `<meta name="description">` if a page doesn't define one. The author of #2067 verified with a one-off grep, but nothing stops a future edit (or new page) from dropping the field. This wires the same check into CI.

## Stacked

Base: `sdk-767-seo-and-og` (PR #2067). Will retarget to `main` once #2067 merges.

## Test plan

- [x] Happy path: `npm run docs:lint` → `✓ Front matter valid: 64 files checked`
- [x] Missing `description:` → exit 1, names the file + field
- [x] Missing `title:` → exit 1, names the file + field
- [x] Empty value (`description:   `) → exit 1
- [x] No front matter block at all → exit 1
- [x] `npm run lint:check` clean (0 errors)
- [x] `npx prettier --check` on changed files clean
- [ ] CI's `docs-build` job runs the new step before `docs:build`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -181,6 +181,15 @@ jobs:
         with:
           node-version-file: '.nvmrc'
 
+      - name: Restore node_modules cache
+        uses: actions/cache/restore@v5
+        with:
+          path: node_modules
+          key: ${{ needs.setup.outputs.cache-key }}
+
+      - name: Validate docs front matter
+        run: npm run docs:lint
+
       - name: Install docs dependencies
         working-directory: docs-site
         run: npm ci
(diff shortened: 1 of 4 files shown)


serikjensen (approved):

mariechatfield (approved):
thank you for adding the yaml parser!! regexes are dead to me

PR #2083, opened 2026-06-09
refactor(SDK-1012): simplify form types and exports

mariechatfield:
## Summary

- **FormData-first schema pattern**: `{Domain}FormData` is now declared as an explicit interface *before* `fieldValidators`, which uses `satisfies ValidatorsFor<FormData>` to constrain validators to match it. Previously `FormData` was a mapped type derived from `fieldValidators` — the new direction reverses this, making the data interface the source of truth.
- **`requiredFieldsConfig` typed against `FormData`**: `RequiredFieldConfig<{Domain}FormData>` instead of `RequiredFieldConfig<typeof fieldValidators>`, and `excludeFields` uses `keyof {Domain}FormData`.
- **`createFormSchema` removed from public API**: These schema factory functions are internal SDK infrastructure; they were removed from `src/index.ts` and from all hook barrel exports.
- **`useDeriveFieldsMetadata` reparametrized**: Now typed over `TFormData` directly rather than `Record<string, z.ZodType>`.
- **`EmployeeStateTaxesMetadataConfig`**: Extended `FieldsMetadataConfig<Record<string, unknown>>` instead of `Record<string, z.ZodType>`.
- **`FLSA_STATUS` renamed** from `FlsaStatus` const (type alias preserved); `FlsaStatus`, `PayPeriod`, and `UsedPreparer` exported as types from the public barrel.
- **`EmployeeDocuments`** moved from `EmployeeManagement` to `EmployeeOnboarding` namespace export.
- Internal authoring docs (`.claude/`) updated to reflect the new 5-part schema structure and removed `createFormSchema` from public export templates and checklists.

## Test plan

- [x] `npm run build` passes (pre-commit hook verified on each commit)
- [x] `npm run test -- --run` passes
- [x] API report diff reviewed (`.reports/embedded-react-sdk.api.md`) — `createFormSchema` entries removed, `useDeriveFieldsMetadata` signature updated


.reports/embedded-react-sdk.api.md
```diff
@@ -223,13 +223,19 @@ export interface BadgeProps extends Pick<HTMLAttributes<HTMLSpanElement>, 'class
...
+export interface BankFormData {
+    // (undocumented)
+    accountNumber: string;
+    // (undocumented)
+    accountType: AccountType;
+    // (undocumented)
+    name: string;
+    // (undocumented)
+    routingNumber: string;
+}
```
mariechatfield, line 238: Wow so much more readable beautiful (this is what's getting documented next)

.reports/embedded-react-sdk.api.md
```diff
@@ -1283,261 +1319,18 @@ export type CountyEntry = {
 // @public (undocumented)
 export type CourtesyWithholdingFieldProps = HookFieldProps<CheckboxHookFieldProps<HomeAddressRequiredValidation>>;
 
-// Warning: (ae-forgotten-export) The symbol "BankFormSchemaOptions" needs to be exported by the entry point index.d.ts
-// Warning: (ae-forgotten-export) The symbol "BuildFormSchemaResult" needs to be exported by the entry point index.d.ts
-// Warning: (ae-missing-release-tag) "createBankFormSchema" is part of the package's API, but it is missing a release tag (@alpha, @beta, @public, or @internal)
-//
-// @public (undocumented)
-export function createBankFormSchema(options?: BankFormSchemaOptions): BuildFormSchemaResult<    {
```
mariechatfield, line 1291: Removed all the actual `createDomainFormSchema` outputs which means a lot of zod types just disappear

.reports/embedded-react-sdk.api.md
```diff
@@ -2518,6 +2336,12 @@ export type FipsCodeFieldProps = HookFieldProps<SelectHookFieldProps<ChildSuppor
 // @public (undocumented)
 export type FirstNameFieldProps = HookFieldProps<TextInputHookFieldProps<NameValidation>>;
 
+// Warning: (ae-forgotten-export) The symbol "FLSA_STATUS" needs to be exported by the entry point index.d.ts
+// Warning: (ae-internal-missing-underscore) The name "FlsaStatus" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export type FlsaStatus = (typeof FLSA_STATUS)[keyof typeof FLSA_STATUS];
+
```
mariechatfield, line 2344: This is a bit tricky. In order to update the `FormData` types I needed to reference types based on constants. Do we actually want to export the `FLSA_STATUS` constant and `PAY_PERIODS` constant etc? if not, if we just want the `FlsaStatus` type to resolve to 

```ts
type FlsaStatus = "Exempt" | "Salaried Nonexempt" | "Nonexempt" |
  "Owner" | "Commission Only Exempt" | "Commission Only Nonexempt"`
```

 we may need to declare the types first _then_ build the constants instead of deriving the type from the constant

src/partner-hook-utils/form/buildFormSchema.ts
```diff
@@ -171,7 +174,9 @@ export function buildFormSchema<
       if (typeof effectiveRule === 'function') {
         dynamicRequired.push({
           name,
-          predicate: isPartnerRequired ? () => true : effectiveRule,
+          predicate: isPartnerRequired
+            ? () => true
+            : (effectiveRule as (data: Record<string, unknown>, mode: FormMode) => boolean),
```
mariechatfield, line 179:
```ts
 const dynamicRequired: Array<{
    name: string
    predicate: (data: Record<string, unknown>, mode: FormMode) => boolean
  }> = []
```

This is how `dynamicRequired` is typed and I tried to figure out how to make it work with `TFormData` and it got very weird very fast so we're just casting this slightly more strongly typed `effectiveRule` off the config to match the `Record<string, unknown>`

jeffredodd (approved):

serikjensen (changes requested):
I think we need to pair on this one. one of the biggest things with the fieldValidators pattern was being able to have a single source of truth. We're now ending up declaring the TS interface and then redeclaring it for zod. It's a lot to try and keep in sync and i worry about the maintenance overhead

My rec would potentially be to proceed with hooks docs where we can to stay unblocked, but ticket the challenge of deriving the data from zod separately so we can give it a dedicated effort. I think we need to put some dedicated thought into how to keep the field validators as the single source of truth.

mariechatfield:
Closing this after syncing with Steve, we don't need to radically change our types to get good docs, we have `@interface` tags and now I know how to use them

PR #2137, opened 2026-06-15
fix(docs): inline navbar logo SVG to prevent progressive-paint flash

jeffredodd:
## Summary

On slow networks (e.g. 3G throttling) the docs site at sdk.gusto.com briefly renders the orange "gusto" portion of the navbar logo before the "embedded" text appears — looking like a flash from a plain "gusto" logo to the full "gustoembedded" wordmark.

**Cause:** `gdev-logo-light.svg` (~8KB) contains two paths: the orange "gusto" path at bytes 0–4237, then the "embedded" text path at bytes 4237–8087. Chromium-family browsers progressively paint SVGs loaded via `` as bytes arrive, so on a slow connection the first path paints before the second has downloaded.

**Fix:** Swizzle `@theme/Navbar/Logo` to render the SVG inline inside the page HTML. There is no separate file fetch and the entire logo paints atomically. Light/dark switching is handled by `currentColor` plus a CSS rule on `.navbar__logo` that swaps black ⇄ white based on `[data-theme="dark"]`.

## Test plan

- [x] `npm run start` in `docs-site/` — homepage and docs pages render correctly
- [x] Light mode: orange "gusto" + black "embedded"
- [x] Dark mode: orange "gusto" + white "embedded"
- [x] No network request for `/img/gdev-logo-*.svg` from the navbar (footer still uses `` — below the fold, no visible flash)
- [ ] Confirm fix on production after deploy by throttling Network to 3G and reloading sdk.gusto.com


aaronlee777 (approved):

serikjensen (approved):

krisxcrash (approved):

docs-site/src/theme/Navbar/Logo/index.tsx
```diff
@@ -0,0 +1,22 @@
+import React, { type ReactNode } from 'react'
+import Link from '@docusaurus/Link'
+import useBaseUrl from '@docusaurus/useBaseUrl'
+
+export default function NavbarLogo(): ReactNode {
```
hukid, line 5: Do we have an official logo?

jeffredodd: @aaronlee777 would be the one to ask here. This is the logo from https://docs.gusto.com/embedded-payroll/docs/introduction

PR #2143, opened 2026-06-15
ci: generate TypeDoc API docs in derive job

mariechatfield:
## Summary

Extends the CI derive job to auto-generate and commit `docs/api/` alongside the endpoint inventory and API report on every push. Pairs with #2142, which checks in the initial files and adds the sdk-router frontmatter logic.

## Changes

- Derive job now installs `docs-site` deps and runs TypeDoc, committing `docs/api/` if stale
- Switched to the `ci-tasks` GitHub App token so derive-job auto-commits re-enter the CI pipeline (default `GITHUB_TOKEN` commits are blocked from triggering new runs)
- Removed `endpoints:derive` from the build script — it now runs exclusively in the dedicated derive job

## Testing

- Green CI build on this branch (stacked on #2142) confirms the full derive → commit → CI re-trigger loop works end-to-end
- Verify derive step locally: `npm run endpoints:derive && npm run api-report:derive && npm run docs:api:generate`


.github/workflows/ci.yaml
```diff
@@ -128,25 +128,71 @@ jobs:
...
+  # Derive job: Re-run endpoint inventory, API report, and TypeDoc, then commit any changes
+  # back to the branch. Runs after build so api-extractor has dist/index.d.ts available.
+  # On non-main branches: auto-commits and pushes derived files if stale.
+  # On main: fails if derived files are stale (they should always be up-to-date after PRs).
+  derive:
+    needs: [build]
+    # Must use the self-hosted runner group — GitHub-hosted runners are blocked
+    # by the org IP allow list and cannot push back to the repository.
+    runs-on:
+      group: gusto-ubuntu-default
```
jeffredodd, line 140: I think..... we need to just be ubuntu default. If we do gusto-ubuntu it goes to our internal runner versus using the free public runner.

mariechatfield: We were chatting in Slack but the takeaway is we may need to figure out an alternative way to get around the IP allow list issue without using this runner, _or_ only use this runner if we do a check and confirm we _need_ to actually commit and push a change

jeffredodd (approved):

mariechatfield:
I just removed `endpoints` as a required action to merge to main. Once this lands I'll add `derive` so future PRs will be required to have that passing before they can merge

PR #2167, opened 2026-06-16
feat(docs-site): move version switcher into the docs sidebar

aaronlee777:
## Summary
- Replaces the navbar version dropdown with a custom **Version selector** at the top of the docs sidebar (desktop + mobile primary menu) so the active version is always visible alongside the sidebar tree.
- `global-docs-sidebar` plugin now exposes per-version sidebar items. On non-docs routes (e.g. landing), mobile menu links route into the user's preferred version instead of always pointing at latest.
- Mobile sidebar footer is now a flex row with **GitHub on the left** and the **color-mode toggle on the right**; the navbar-only hide rule for the GitHub pill is scoped to `.navbar__items` so it doesn't hit the footer.
- Mobile menu scrollbar picks up the same thin styling as the desktop sidebar/TOC.

### About `SidebarVersionSelect`
Custom button + popover (not a native `<select>`) so the trigger is clickable across its full width and the menu can be styled. Closes on outside click / Escape, saves the preferred version via Docusaurus's `useDocsPreferredVersion`, and routes to the equivalent doc in the chosen version (falls back to the version's main doc).

### Why the `.gitignore` change
Versioning snapshots (`versions.json`, `versioned_docs/`, `versioned_sidebars/`) are owned by `Gusto/embedded-sdk-docs`, not this repo. To develop the version-dropdown UI you need a couple of cut versions locally — gitignoring those paths lets contributors do that without accidentally committing snapshots.

## Test plan
- [ ] Locally cut at least two doc versions (`npm --prefix docs-site run docusaurus -- docs:version 0.47.0`, then bump and repeat) so the selector has something to show
- [ ] Desktop: open `/docs/` and confirm the Version selector renders above "Getting Started"; switch versions; verify URL + persisted preference
- [ ] Mobile: open hamburger on a docs page; selector at top, footer has GitHub + theme toggle; switching versions navigates to the equivalent doc
- [ ] Landing page on mobile: switch version, confirm the docs sidebar links in the menu now point at `/docs/<version>/...`
- [ ] Light + dark mode visual check for trigger, popover, and scrollbar

### DIFF
--- a/docs-site/.gitignore
+++ b/docs-site/.gitignore
@@ -2,3 +2,9 @@ node_modules/
 .docusaurus/
 build/
 .cache-loader/
+
+# Version snapshots are owned by Gusto/embedded-sdk-docs; only created locally
+# to develop the version-dropdown UI.
+versions.json
+versioned_docs/
+versioned_sidebars/
--- a/docs-site/docusaurus.config.ts
+++ b/docs-site/docusaurus.config.ts
@@ -127,15 +127,6 @@ const config: Config = {
           position: 'left',
           label: 'Docs',
         },
-        ...(hasVersions
-          ? [
-              {
-                type: 'docsVersionDropdown' as const,
-                position: 'right' as const,
-                dropdownActiveClassDisabled: true,
-              },
-            ]
-          : []),
         {
           href: 'https://github.com/Gusto/embedded-react-sdk',
           label: 'GitHub',
--- a/docs-site/plugins/global-docs-sidebar/index.ts
+++ b/docs-site/plugins/global-docs-sidebar/index.ts
@@ -1,10 +1,15 @@
 import type { LoadContext, Plugin } from '@docusaurus/types'
-import type { LoadedContent, LoadedVersion } from '@docusaurus/plugin-content-docs'
+import type { LoadedContent, PropSidebarItem } from '@docusaurus/plugin-content-docs'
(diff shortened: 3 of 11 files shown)


docs-site/docusaurus.config.ts
```diff
@@ -127,15 +127,6 @@ const config: Config = {
           position: 'left',
           label: 'Docs',
         },
-        ...(hasVersions
```
jeffredodd, line 130: I think you'll want to restore this... unless i'm missing something here

jeffredodd (changes requested):

jeffredodd:
Aside from that it looks good!

jeffredodd (approved):

docs-site/.gitignore
```diff
@@ -2,3 +2,9 @@ node_modules/
 .docusaurus/
 build/
 .cache-loader/
+
+# Version snapshots are owned by Gusto/embedded-sdk-docs; only created locally
+# to develop the version-dropdown UI.
+versions.json
+versioned_docs/
+versioned_sidebars/
```
mariechatfield, line 10: @aaronlee777 and @jeffredodd I _think_ we needed to update the build script to force add / not delete versions when we're syncing sources? I believe this is why https://github.com/Gusto/embedded-sdk-docs/commit/cbb5e90d441696e9ae20cf5c269eb7676f30da05 now ignores those directories and we've lost them on the docs site

PR #2168, opened 2026-06-16
fix(Form): disable native browser validation by default

serikjensen:
## Summary

The SDK validates forms exclusively through react-hook-form + Zod, yet the shared `Form` component did not set `noValidate`. As a result, native HTML constraint validation (`type="email"`, `min`/`max`, `maxLength`, etc.) can fire browser-native validation bubbles that:

- bypass the SDK design system and look inconsistent across browsers, and
- can block `onSubmit` from firing, interfering with the SDK's own validation/submission.

This sets `noValidate` on the shared `Form` component (`src/components/Common/Form/Form.tsx`), which backs 50+ forms across the SDK. The existing `noValidate` prop still lets a caller opt back into native validation (`noValidate={false}`).

This does **not** remove validation — react-hook-form + Zod still enforce every constraint. Native `required` was never wired up (`isRequired` only drives the visual label + aria), so no form depends on native enforcement.

## Why it doesn't break forms

- All SDK forms route submission/validation through react-hook-form + Zod resolvers; native constraint validation was redundant.
- `isRequired` is not mapped to the native `required` attribute, so nothing relied on native blocking.
- jsdom doesn't run constraint validation, so existing tests are unaffected; targeted `Form` tests were added.

## Test plan

- [x] `npm run test -- --run src/components/Common/Form/Form.test.tsx` (3 passing: default `noValidate`, opt-out via prop, preventDefault + onSubmit forwarding)
- [x] ESLint clean on changed files
- [x] Pre-commit build passes
- [ ] CI green

### DIFF
--- a/src/components/Common/Form/Form.tsx
+++ b/src/components/Common/Form/Form.tsx
@@ -5,10 +5,11 @@ import styles from './Form.module.scss'
 export type FormProps = React.FormHTMLAttributes<HTMLFormElement>
 
 /** @internal */
-export const Form = ({ children, className, onSubmit, ...props }: FormProps) => {
+export const Form = ({ children, className, onSubmit, noValidate = true, ...props }: FormProps) => {
   return (
     <form
       className={classNames(styles.form, className)}
+      noValidate={noValidate}
       onSubmit={e => {
         e.preventDefault()
         onSubmit?.(e)
--- a/src/components/Common/Form/Form.test.tsx
+++ b/src/components/Common/Form/Form.test.tsx
@@ -1,0 +1,38 @@
+import { describe, expect, it, vi } from 'vitest'
+import { fireEvent, screen } from '@testing-library/react'
+import { Form } from './Form'
+import { renderWithProviders } from '@/test-utils/renderWithProviders'
+
+describe('Form', () => {
+  it('disables native browser validation by default', () => {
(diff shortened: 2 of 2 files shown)


src/components/Common/Form/Form.tsx
```diff
@@ -5,10 +5,11 @@ import styles from './Form.module.scss'
 export type FormProps = React.FormHTMLAttributes<HTMLFormElement>
 
 /** @internal */
-export const Form = ({ children, className, onSubmit, ...props }: FormProps) => {
+export const Form = ({ children, className, onSubmit, noValidate = true, ...props }: FormProps) => {
   return (
     <form
       className={classNames(styles.form, className)}
+      noValidate={noValidate}
```
jeffredodd, line 12: Going to hit approve, but curious if you have an idea of how this might affect A11y?

jeffredodd (approved):

src/components/Common/Form/Form.tsx
serikjensen: i'm not sure if there's a screenreader announcement on form related to it? but beyond that we already have been doing non native validation for our forms. We just have react-aria opting out of it at the component level. This makes it tough for component adapters to wire it up correctly

if there's partner signal around this we may end up needing to expose this as a global config

PR #2169, opened 2026-06-16
fix(EmptyData): drop magnifying glass fallback illustration

aaronlee777:
## Summary
- Remove the default magnifying glass `` fallback from `EmptyData`; the icon slot now renders only when a caller passes `icon`.
- Tighten the text stack — medium-weight title, supporting-variant description, and explicit 16/4 gaps via `Flex` — and shrink the icon slot to 36×36.
- Delete the unused `magnifyingGlass.png` asset and the `icons.magnifyingGlass` i18n key (types regenerated).

`EmptyData` is `@internal` and not part of the `ComponentsContext` adapter, so there is no partner-facing API change.

## Test plan
- [ ] `npm run test -- --run`
- [ ] Storybook spot-check of consumers that previously relied on the default illustration (PayrollHistory, DocumentList, InformationRequestList, ContractorList, payments lists, StateTaxes, Locations) — confirm they render cleanly without a graphic.
- [ ] Storybook spot-check of consumers that pass their own `icon` (WorkAddressView, PolicyDetailLayout, etc.) — confirm icon still renders at the new 36×36 size.


mariechatfield (approved):
Love this! Looks great locally. I think I did find a bug with the HTML adapter _in_ the sdk demo where the empty state just doesn't render on wider screens, but that's on main too so your changes don't actually cause that

Screenshot 2026-06-16 at 1 07 49 PM
Screenshot 2026-06-16 at 1 08 10 PM

mariechatfield (approved):
Love this! Looks great locally. I think I did find a bug with the HTML adapter _in_ the sdk demo where the empty state just doesn't render on wider screens, but that's on main too so your changes don't actually cause that

Image
Image

PR #2171, opened 2026-06-16
ci(derive): split derive into parallel check + conditional commit jobs

mariechatfield:
## Summary

- Adds `npm run derive` — a single standalone command that builds then regenerates all derived files (endpoint inventory, API report, TypeDoc). No more running three commands manually.
- Replaces the single `derive` CI job (always on expensive self-hosted runner) with four jobs:
  - **`derive-check-src`** (`ubuntu-latest`, parallel with `build`): checks endpoint inventory + TypeDoc staleness; runs both checks in parallel and doesn't block on the build job
  - **`derive-check-dist`** (`ubuntu-latest`, after `build`): checks API report staleness; requires `dist/` from the build job
  - **`derive-commit`** (`gusto-ubuntu-default` + `ci-tasks`): only runs when a check job reports stale — the expensive runner and environment secrets are never used on branches with up-to-date derived files
  - **`derive`** aggregator (`ubuntu-latest`): single required status check for branch protection; gates on both check jobs passing so the branch protection rule stays simple
- Dependabot now runs the check jobs (previously skipped entirely); staleness is an error rather than a commit trigger, matching main's behavior — no action needed since derived files don't change with dep bumps
- Minor CLAUDE.md cleanup: adds `derive` to the commands table and fixes code-fence language hints (`text`, `bash`)

## Test plan

- [ ] Push a branch with no derived-file changes → `derive-check-src` and `derive-check-dist` pass, `derive-commit` is skipped, `derive` aggregator passes
- [ ] Push a branch with a stale derived file → check job detects staleness, `derive-commit` runs and auto-commits, `derive` aggregator passes
- [ ] Verify `derive-commit` does not appear on Dependabot PRs (check jobs run; commit job is unreachable because checks fail on staleness rather than setting `stale=true`)
- [ ] `npm run derive` runs locally end-to-end (requires `cd docs-site && npm ci` first)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -285,12 +285,15 @@ jobs:
 
       - name: Commit and push derived files
         run: |
-          git add -u
+          git add docs/reference
           git add docs/api
+          git add .reports
           git config user.name "github-actions[bot]"
           git config user.email "github-actions[bot]@users.noreply.github.com"
-          git commit -m "chore: update derived files"
-          git push
+          if ! git diff --staged --quiet; then
+            git commit -m "chore: update derived files"
+            git push
+          fi
 
   # derive: Single required check that branch protection can gate on.
   # Aggregates derive-check-src and derive-check-dist. derive-commit is excluded
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -240,6 +240,7 @@ jobs:
         with:
           name: derived-dist
           path: .reports
+          include-hidden-files: true
           retention-days: 1
           if-no-files-found: error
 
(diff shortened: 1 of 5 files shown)


mariechatfield (commented):
Self-review — two issues found and fixed in the same pass:

**`git add -u` → explicit adds** (`derive-commit`, commit step): replaced `git add -u` with `git add docs/reference`, `git add docs/api`, `git add .reports` to match `derive-check-src`'s staging logic. `git add -u` only stages modifications to tracked files; explicit paths also catch new untracked files, preventing a silent omission if `endpoints:derive` ever creates new files in `docs/reference`.

**Restored empty-commit guard** (`derive-commit`, commit step): added back `if ! git diff --staged --quiet` around the commit+push. The original code had this guard; removing it created a race where two concurrent `derive-commit` jobs could both detect staleness, the second one re-derive the same files, and then fail on `git commit` with nothing to commit. Guard restored.

jeffredodd (approved):

PR #2144, opened 2026-06-15
docs: improve TypeDoc output and consolidate legacy generated docs

mariechatfield:
## Summary

Builds on the TypeDoc pipeline established in #2133 with a set of output quality improvements, new standalone pages, and removal of three legacy doc generation scripts now superseded by TypeDoc:

- Adds `STANDALONE_PAGES` routing to `sdk-router` — routes TypeDoc-generated reflections to dedicated pages by source path or `@group` tag, with tests
- Adds group-based routing so the new `Events` standalone page can pull `componentEvents` and `EventType` by `@group` tag rather than source path, keeping unrelated constants (e.g. `PAY_PERIODS`, `EmployeeOnboardingStatus`) on the project index
- Exports nested utility types referenced by public component props (`Breadcrumb`, `DescriptionListItem`, `TabProps`, etc.) and converts several type aliases to interfaces so TypeDoc tracks extension relationships correctly
- Groups component adapter prop types under a `Component Props` TSDoc group, renames the section from "Component Adapters" to "Component Inventory"
- Sorts interface properties required-first then alphabetical in TypeDoc output
- Suppresses `undefined` in the Default value column of property tables
- Removes `generateAdapterPropDocs`, `generateThemeDocs`, and `eventTypeDocsEmitter` build scripts — their output is now covered by TypeDoc-generated pages under `docs/api/`
- Consolidates `docs/component-adapter/component-inventory.md`, `docs/theming/theme-variables.md`, and `docs/integration-guide/event-types.md` into the corresponding TypeDoc pages; updates all cross-links
- Adds `@example` block to `GustoSDKTheme` so TypeDoc surfaces the usage example
- Improves `event-handling.md` with cross-links to the Events API reference and component docs
- Regenerates `docs/api/` with all improvements applied

Screenshot 2026-06-16 at 11 43 59 AM
Screenshot 2026-06-16 at 11 44 15 AM
Screenshot 2026-06-16 at 11 44 25 AM
Screenshot 2026-06-16 at 11 44 33 AM

## Test plan

- [ ] `npm run test -- --run docs-site/plugins/sdk-router.test.ts` passes
- [ ] `npm --prefix docs-site run start` opens the docs site locally and each of the new pages looks correct
  - [ ] http://localhost:3000/docs/api/component-inventory
  - [ ] http://localhost:3000/docs/api/theme-variables
  - [ ] http://localhost:3000/docs/api/utilities
  - [ ] http://localhost:3000/docs/api/events
[...]

PR #2173, opened 2026-06-16
docs: export utility types, add TSDoc @group tags (2/4)

mariechatfield:
## Summary

Split from #2144 (2 of 4). Stacked on #2172 — review after that merges (or switch base to `main` at that point).

Exports previously-internal types referenced by public component props, converts type aliases to interfaces for correct TypeDoc extension tracking, and adds `@group` TSDoc tags for consistent categorization:

- Adds `COMPONENT_PROP_GROUPS` (`Component Props`, `Utility Types`) to `typedoc-utils.mjs` and wires it into the `groupOrder` in `typedoc.config.ts`; `@group` values validated by `eslint-rules/tsdoc-valid-group.ts`
- Groups all `*Props` interfaces under a `Component Props` TSDoc group on the component adapter standalone page
- Exports `Breadcrumb`, `DescriptionListItem`, `TabProps`, `BaseListProps`, and other nested utility types so TypeDoc documents them from their definition rather than inlining them as anonymous shapes
- Converts several `type` aliases (`SharedHorizontalFieldLayoutProps`, `OrderedListProps`, `UnorderedListProps`, etc.) to `interface` so TypeDoc correctly tracks their extension relationships
- Adds `HorizontalFieldLayout` to the standalone pages routing and `Utility Types` to the component prop groups
- Switches component adapter context types to use `React.FunctionComponent` for consistent type representation
- Updates the API surface report (`.reports/embedded-react-sdk.api.md`)

Regenerates `docs/api/` with all changes applied.

## Depends on / stack

Stacked on #2172. Full split from #2144:

| # | Branch | Changes |
|---|--------|---------|
| 1 | `docs/marie/pr1-routing` (#2172) | Routing + TypeDoc infrastructure |
| **2** | `docs/marie/pr2-types` (this PR) | Public type / TSDoc metadata |
| 3 | `docs/marie/pr3-formatting` | TypeDoc output formatting |
| 4 | `docs/marie/pr4-consolidate` (#2176) | Legacy docs consolidation |

## Test plan

- [ ] `npm --prefix docs-site run test` passes (109 tests)
- [ ] `npm run docs:api:generate` produces no new errors
- [ ] `npm --prefix docs-site run start` and verify the component adapter page shows `Component Props` and `Utility Types` groups


docs/api/index.md
mariechatfield, line 1: The deleted diff here shows why we need `SharedHorizontalFieldProps` to be an interface instead of a type alias. As a type alias, it doesn't include the beautiful `extended by` links that `HorizontalFieldProps` has. As an interface, we can see (in the `component-inventory.md`) that it's extended by switch, checkbox, and radio props

src/components/Common/UI/Link/LinkTypes.ts
```diff
@@ -5,8 +5,9 @@ import type { AnchorHTMLAttributes, ReactNode } from 'react'
  * Renders an HTML anchor (`<a>`) for inline navigation.
  *
  * @public
+ * @group Component Props
  */
-export type LinkProps = Pick<
+export interface LinkProps extends Pick<
```
mariechatfield, line 10: All of these props work better as interfaces. The documentation is cleaner and it's also a better representation of the contract

src/contexts/ComponentAdapter/useComponentContext.ts
```diff
@@ -95,84 +95,84 @@ import type { PayrollLoadingProps } from '@/components/Common/PayrollLoading/Pay
  */
 export interface ComponentsContextType {
   /** Status message with an optional dismiss action; used for errors, warnings, success, and info. */
-  Alert: (props: AlertProps) => JSX.Element | null
+  Alert: FunctionComponent<AlertProps>
```
mariechatfield, line 98: `FunctionComponent` from React here does most of the heavy lifting _and_ it also means we get a much cleaner documentation that links directly to the props. This is technically a contract change but not much of one

docs/component-adapter/component-inventory.md
mariechatfield, line 1: Ignore this file, this is the auto-generated old component inventory I'm deleting in the 4/4 PR

src/components/Common/UI/Link/LinkTypes.ts
mariechatfield: after pairing with @serikjensen this morning, I think I should undo the source code changes and just use `@interface` in the comments to get the same result

mariechatfield: just kidding, only `SharedHorizontalFieldLayoutProps` makes sense as a type alias that's documented as `@interface` the other ones all either should have been interfaces or end up being really weird as aliases (the `Ordered` and `Unordered` list props)

serikjensen (approved):

PR #2191, opened 2026-06-17
feat: add CSP nonce support to GustoProvider

jeffredodd:
## Summary

- Adds an optional `nonce` prop to `GustoProvider` / `GustoProviderCustomUIAdapter`. The nonce flows through a new `NonceContext` and is applied to both runtime `<style>` elements the SDK creates: the theme-variables element in `ThemeProvider`, and the loading-spinner element injected into the paystub PDF download window.
- Exposes a new public `useNonce` hook so partners using custom UI can plumb the same nonce into any `<style>` or `<script>` they inject themselves.
- Extracts three SDK-authored static inline `style={{…}}` attributes (`Card`, `Toast`, `DocumentSigner/List`) to CSS modules, so they no longer require `style-src-attr 'unsafe-inline'`.
- Documents the resulting minimum CSP and the inline-style surface that remains (`Flex`/`Grid`/`ProgressBar`/`ReorderableList` runtime CSS custom properties, plus `react-aria-components` overlay positioning) under `docs/getting-started/proxy-security-partner-guidance.md`.

## Why

Partners running a strict nonce-based CSP can't use the SDK today — the theme `<style>` and paystub-popup `<style>` tags are dropped by the browser without `'unsafe-inline'`. After this change, the only `style-src` relaxation a partner needs is `style-src-attr 'unsafe-inline'`, which is forced by the CSP spec (no per-attribute nonces) and by `react-aria-components`.

## Test plan

- [x] `npm run test -- --run` — 3110 pass, 1 expected fail (pre-existing).
- [x] `npm run build` — clean build, types emit.
- [x] New unit coverage: `useNonce` returns the value / undefined; `ThemeProvider` sets `.nonce` on the injected `<style>`; `PaystubsCard` sets `.nonce` on the popup `<style>` when one is provided.
- [ ] Manual smoke under a nonce-only `style-src` in the SDK dev app (reviewer please verify locally if you can).


jeffredodd:
@cursoragent Can you resolve conflicts?

krisxcrash (approved):

PR #2199, opened 2026-06-17
docs: rewrite landing and add Quick Start guide

elena-sharma-g:
## Summary

Revamp of landing / getting started documentation pages

## Changes

- Adds docs/getting-started/quick-start.md — developer-focused guide with install callout, 3-step overview, "Try it with the demo application" section, and configuration subsections.
- Rewrites docs/getting-started/getting-started.md from scratch as the landing page — replaces the old install/setup content with a capabilities table, theming and composition overview, and a callout pointing developers to Quick Start.
- Removes the deciding-to-build-with-the-sdk page and moves component-types into getting-started/ section.
- Updates docs-site/sidebars.ts and link references across docs pages to match the new structure.


larchai (approved):
I think this is a good starting point. Whether in the quick start page itself or elsewhere though, I think we should actually suggest a short set of both A) user stories (which can also evolve) for the developer to play out with the reference app in order to understand how basic payroll workflows look, and B) specific things to try in order to better understand our build approaches (e.g. build your own version of X subcomponent using these hooks). Of course it's still up to the developers if they want to engage, but think of it as a "tutorial" type of thing similar to when you open a new software/app for the first time - a guided way to see what's going on.

elena-sharma-g:
Thanks, Charlie!! For the review and the feedback, that's super valid. FWIW I think a lot of the examples and "tutorial" level content is tackled in https://github.com/Gusto/embedded-react-sdk/pull/2209 so I'll look into what a good follow-up would be after that gets merged.

docs/getting-started/quick-start.md
```diff
@@ -0,0 +1,75 @@
...
+
+- **Authentication.** Your proxy acquires OAuth2 tokens from the Gusto Embedded API on behalf of the authenticated user. See [Authentication](./authentication.mdx) for the full setup.
+- **User IP address.** Some workflows require users to sign forms, which needs the user's IP. Your proxy must add the `x-gusto-client-ip` header to forwarded requests so the IP can't be spoofed client-side.
+- **Authorization.** User-level authorization is your responsibility. At minimum, your proxy should authenticate every request, allowlist endpoints, validate resource ownership, and log proxied traffic. See [Proxy Security](./proxy-security-partner-guidance.md) for detailed practices.
+
+The SDK mirrors the [Gusto Embedded API Reference](https://docs.gusto.com/embedded-payroll/reference/whats-new-in-v2025-06-15) with a 1:1 endpoint mapping, so your proxy doesn't need to transform requests—it just forwards them with authentication attached.
+
```
serikjensen, line 44: @cursoragent could you create a separate follow up PR and explain that there is an example of a working proxy in the SDK demo application? You can link to the proxy file here https://github.com/Gusto/embedded-react-sdk-demo-app/blob/main/backend/src/proxy.ts

PR #2202, opened 2026-06-17
fix: clear server-side field errors on edit (SDK-923)

jeffredodd:
## Summary

Fixes [SDK-923](https://gustohq.atlassian.net/browse/SDK-923) — server-side validation errors on form fields blocked resubmission until a page refresh (reported on Pay Schedule date fields, but the bug is SDK-wide).

- `useSyncFieldErrors` (`SDKFormProvider`) maps server `fieldErrors` onto RHF via `setError(name, { type: 'custom', message })`.
- `useField.handleChange` never called `clearErrors`, and forms run `mode: 'onSubmit'`, so RHF didn't reclear on change either. The custom error sat there forever.
- Fix: in `handleChange`, when the field already has a `type: 'custom'` error, call `clearErrors(name)` before forwarding the new value. Client-side validation errors (`required`, `min`, zod-driven, etc.) are untouched — RHF's normal validation loop is unaffected.

Despite the ticket framing, this was **not** a state-machine problem — `PaySchedule` already uses a robot3 machine. The bug was in the shared form layer, so the fix lives there and benefits every form, not just Pay Schedule.

## Test plan

- [x] `npm run test -- --run src/components/Common/Fields/hooks/useField.test.tsx` (22 pass; 3 new)
- [x] `npm run test -- --run src/partner-hook-utils/form/ src/components/Company/PaySchedule/` (152 pass)
- [x] `npm run test -- --run src/components/Common/Fields/` (51 pass)
- [x] Verified the new positive tests fail without the fix and pass with it (red → green); the negative test (client-side `required` error not cleared on edit) passes in both states, confirming the change is surgical.
- [ ] Manual repro in sdk-app: open Pay Schedule, submit a date combination that trips a server validator, confirm the date error appears, edit the date, confirm the error clears, then resubmit successfully.
- [ ] Spot-check a non-date field elsewhere (e.g., Compensation rate) to confirm the same clear-on-edit behavior across field types.

[SDK-923]: https://gustohq.atlassian.net/browse/SDK-923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Company/PaySchedule/shared/usePayScheduleForm/usePayScheduleForm.tsx
+++ b/src/components/Company/PaySchedule/shared/usePayScheduleForm/usePayScheduleForm.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo } from 'react'
+import { useCallback, useEffect, useMemo, type FocusEventHandler } from 'react'
 import { useForm, useWatch } from 'react-hook-form'
 import type { UseFormProps } from 'react-hook-form'
 import { zodResolver } from '@hookform/resolvers/zod'
@@ -132,6 +132,11 @@ export interface UsePayScheduleFormReady extends BaseFormHookReady<
   actions: {
     /** Validates the form and dispatches the create or update mutation. Returns the saved schedule, or `undefined` if validation failed. */
     onSubmit: () => Promise<HookSubmitResult<PayScheduleShow> | undefined>
+  }
+  /** Form bindings extended with a blur handler for the two anchor date fields. */
+  form: BaseFormHookReady<FieldsMetadata, PayScheduleFormData, PayScheduleFields>['form'] & {
+    /** Wire to `onBlur` on a wrapper around the anchor date field JSXs to clear stale server-side (`type: 'custom'`) errors when the user leaves the group (SDK-923). */
+    handleDateFieldsBlur: FocusEventHandler<HTMLDivElement>
   }
 }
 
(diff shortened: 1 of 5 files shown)


src/components/Common/Fields/hooks/useField.ts
```diff
@@ -81,6 +81,9 @@ export function useField<TValue = string, TRef = HTMLInputElement>({
 
   const handleChange = (updatedValue: TValue) => {
     const value = transform ? transform(updatedValue) : updatedValue
+    if (fieldState.error?.type === 'custom') {
```
jeffredodd: Confirmed errors are coming back as type custom from the server. Seems save if a user is indeed changing the input.

src/components/Company/PaySchedule/shared/usePayScheduleForm/usePayScheduleForm.tsx
```diff
@@ -309,6 +309,22 @@ export function usePayScheduleForm({
...
+      if (!name) return
+      if (formMethods.getFieldState(name).error?.type === 'custom') {
+        formMethods.clearErrors(name)
+      }
+    })
+    return () => {
+      subscription.unsubscribe()
+    }
+  }, [formMethods])
+
```
serikjensen: i'm a little worried about doing this subscription based. is this watching each field on the form?

I'm wondering if we could do an event based fix here instead where we could detect changes to the problem fields and clear them if an error is present?

if we keep this approach i think we need to include each formMethod function as an entry in the dep array individually. formMethods in the dep array used to cause infinite renders (maybe they fixed it?)

serikjensen:
This ends up being a global issue with some field responses. going to close this one and look for a universal resolution

PR #2225, opened 2026-06-22
fix(ContractorPayments): align numeric columns and table footer

aaronlee777:
## Summary
- Right-align numeric columns (hours, wage, bonus, reimbursement, total) across contractor payment tables (CreatePayment, Preview, PaymentsList, PaymentHistory, PaymentSummary)
- Fix `DataTable` footer cells to honor each column's `justify` setting so totals line up under their column values
- Update DataTable tests to cover footer justify behavior

## Test plan
- [ ] `npm run test -- --run src/components/Common/DataView/DataTable/DataTable.test.tsx`
- [ ] Verify each contractor payments table in Storybook / sdk-app: numeric columns and footer totals are right-aligned

### DIFF
--- a/src/components/Common/DataView/DataTable/DataTable.test.tsx
+++ b/src/components/Common/DataView/DataTable/DataTable.test.tsx
@@ -205,6 +205,19 @@ describe('DataTable Component', () => {
       expect(screen.getByText('Name').closest('div')?.className).toMatch(/cellEnd/)
       expect(screen.getByText('Alice').closest('div')?.className).toMatch(/cellEnd/)
     })
+
+    test('applies column justify to footer cells', () => {
+      renderTable({
+        columns: [
+          { key: 'name', title: 'Name', render: item => item.name },
+          { key: 'age', title: 'Age', justify: 'end', render: item => item.age.toString() },
+        ],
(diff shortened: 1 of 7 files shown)


src/components/Common/DataView/DataTable/DataTable.test.tsx
```diff
@@ -205,6 +205,19 @@ describe('DataTable Component', () => {
...
+        columns: [
+          { key: 'name', title: 'Name', render: item => item.name },
+          { key: 'age', title: 'Age', justify: 'end', render: item => item.age.toString() },
+        ],
+        footer: () => ({ name: 'Total', age: '55' }),
+      })
+
+      expect(screen.getByText('Total').closest('div')?.className).not.toMatch(/cellEnd/)
+      expect(screen.getByText('55').closest('div')?.className).toMatch(/cellEnd/)
+    })
```
serikjensen, line 220: nit: i think layout stuff like this can result in pretty brittle tests. i'd actually recommend not including it. if we do include it we'd want to just assert that the classname gets set rather than trying to verify like this

serikjensen (approved):
see comment above on potentially omitting the included test

PR #2227, opened 2026-06-22
fix(Locations): render address lines as stacked flex items

aaronlee777:
## Summary
- Wrap street and city/state/zip in span elements so the address flex column produces two separate items instead of a merged text run.
- Add line-specific font weights for visual hierarchy and tighten the gap to `toRem(2)`.

## Test plan
- [ ] Storybook: verify Locations list renders each address on two lines with the medium/regular weight contrast.

### DIFF
--- a/src/components/Company/Locations/LocationsList/List.module.scss
+++ b/src/components/Company/Locations/LocationsList/List.module.scss
@@ -1,7 +1,15 @@
 .addressCell {
   display: flex;
   flex-direction: column;
-  gap: toRem(4);
+  gap: toRem(2);
   container-type: normal;
   font-style: normal;
+}
+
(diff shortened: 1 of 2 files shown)


serikjensen (commented):
Why did we remove the usage of the Text component here? wouldn't we still want adapters to apply?

aaronlee777:
> Why did we remove the usage of the Text component here? wouldn't we still want adapters to apply?

@serikjensen DataViews should be getting their styling from that UI component without the addition of text components in an instance. It doesn't expect the the address tag causing styling issues. You are correct that in the case we should make an exception and render this with Text components. I'll adjust.

krisxcrash (approved):

PR #2241, opened 2026-06-23
refactor(ContractorPayments): replace total tally with live hourly pay hint

aaronlee777:
## Summary
- Remove the running "Total pay" line from the edit-contractor-payment modal — with only 2–3 inputs in play, the tally was noise.
- For hourly contractors, show a live `{rate}/hr × hours = {total}` hint directly under the Hours input.
- Update the hint imperatively via a native `input` event listener (not React state) so we don't re-render react-aria's NumberField mid-keystroke.

## Test plan
- [ ] Open the edit-payment modal for an **hourly** contractor — hint appears under Hours and updates live as you type.
- [ ] Open the edit-payment modal for a **fixed-wage** contractor — no hint, no "Total pay" line.
- [ ] Typing in the Hours input works normally (no character drops, no resets).
- [ ] Arrow keys still increment/decrement and the hint updates accordingly.
- [ ] Hint hides when no hourly rate is set on the contractor.


serikjensen (changes requested):
## Review — requesting changes 🔧

Nice simplification — dropping the running tally and the extra section headings is the right direction. Three changes are required before merge:

**1. Accessibility: move the computed pay text into the field's `description`.** It's currently a free-standing `<Text><span>` sibling of the Hours input, so it isn't associated with the input or announced to assistive tech. `NumberInputField` already accepts a `description` and wires it via `aria-describedby` — render it through that.

**2. Replace the manual event listener with an `onInputChange` prop.** The hand-rolled `addEventListener` / `useCallback` ref / cleanup-ref machinery is reimplementing input handling that belongs in the component. Add an `onInputChange` to `NumberInput`/`NumberInputField` (there's precedent in `ComboBox` and `MultiSelectComboBox`) and move the hours computation onto it. Plain `onChange` won't work here — react-aria's `NumberField` only commits on blur, not per-keystroke.

**3. Track the computed text in React state, not via direct DOM manipulation.** Writing to a ref's `textContent` is against React's model and gets clobbered on any re-render. Keep the on-open seed (the field is prefilled from the API, and a programmatic reset doesn't fire an input event, so the text must be populated before interaction) — but source it correctly and let `onInputChange` be the sole live writer, so there's only one writer at a time.

Please also **stress-test live**: type fast into the Hours field, hold a key, use arrow steps, paste, and tab away/back — confirm no dropped characters, cursor jumps, or resets, and that the text is correct on open before any interaction.

---

@cursoragent — please apply the changes exactly as specified in the prompt below and push them to this branch.

**Reviewer note:** even with the agent applying these, still pull the branch down locally and manually verify the result — especially the live stress-test of the Hours input — before approving. Don't rely on the agent's changes alone.

<details>
<summary>📋 Prompt for Cursor agent</summary>

```
In the embedded-react-sdk repo, on branch
al/contractor-payment/simplify-edit-modal-with-live-hours-pay, apply this PR review feedback
for the edit-contractor-payment hourly pay text (EditContractorPaymentPresentation.tsx and the
shared NumberInput primitive). Make exactly three changes.

CHANGE 1 — Accessibility: render the computed pay text through the field's `description` prop.
Currently the text is a free-standing <Text><span ref=...> sibling of the Hours
NumberInputField, so it has no aria-describedby association. NumberInputField accepts
`description?: React.ReactNode` and the underlying NumberInput wires it to the input via
descriptionId/ariaDescribedBy. Pass the text through that prop instead of a separate sibling:

    <NumberInputField
      min={0}
      name="hours"
      isRequired
      label={t('hoursLabel')}
      adornmentEnd={t('hoursAdornment')}
      description={hourlyRate && hourlyRate > 0 ? hoursPayDescription : undefined}
      onInputChange={raw => setHoursPayDescription(computeHoursPayDescription(parseHours(raw)))}
    />

CHANGE 2 — Add an `onInputChange` prop and move the computation onto it; delete the listener.
react-aria's NumberField only fires onChange on commit (blur/Enter/step), NOT per keystroke, so
onChange cannot drive a live update — that's the whole reason. Surface the per-keystroke value
with a new prop instead of attaching a DOM listener:

  a) In src/components/Common/UI/NumberInput/NumberInputTypes.ts add:
       /** Fires on every keystroke with the raw input string (pre-commit), unlike onChange which
        * fires on blur/Enter. */
       onInputChange?: (value: string) => void

  b) In src/components/Common/UI/NumberInput/NumberInput.tsx, destructure onInputChange and wire it
     to the underlying <Input>'s native onInput (Input already spreads props onto the AriaInput):
       <Input
         /* ...existing props... */
         onInput={event => onInputChange?.(event.currentTarget.value)}
       />

  c) In src/components/Common/Fields/NumberInputField/NumberInputField.tsx, accept onInputChange and
     pass it through to the rendered component (alongside the existing numberInputProps spread).

  d) Mirror the existing precedent: ComboBox.tsx maps react-aria's onInputChange -> onChange, and
     MultiSelectComboBox.tsx uses onInputChange directly.

  e) In EditContractorPaymentPresentation.tsx, DELETE the imperative machinery entirely: the
     useCallback ref (hoursInputRef), hoursInputCleanupRef, the addEventListener/removeEventListener,
     the eslint-disable-next-line react-hooks/exhaustive-deps, and the ref'd <span>. The hours
     computation now lives only in the onInputChange handler shown in Change 1.

CHANGE 3 — Track the computed text in state with a descriptive name; keep the seed; one writer.
Replace the textContent writes with React state. Use a descriptive name (hoursPayDescription),
not a generic "hint". Define the compute helper once (it already closes over hourlyRate, t, and
currencyFormatter):

    const computeHoursPayDescription = (hours: number) => {
      if (!hourlyRate || hourlyRate <= 0) return ''
      return t('hoursPayDescription', {
        rate: currencyFormatter(hourlyRate),
        total: currencyFormatter(hours * hourlyRate),
      })
    }

    const [hoursPayDescription, setHoursPayDescription] = useState('')

    // SEED on open / contractor switch. Required because hours is prefilled from the API and a
    // programmatic reset() does NOT fire an input event, so onInputChange won't populate it.
    // Read via getValues (reliable here: reset() runs before the modal opens). Do NOT seed via a
    // useState initializer (runs once at mount when hours is still 0) and do NOT use
    // useWatch('hours') (that adds a second writer firing on blur that fights onInputChange).
    useEffect(() => {
      if (isOpen) {
        setHoursPayDescription(computeHoursPayDescription(formMethods.getValues('hours') ?? 0))
      }
      // Depend on the stable formMethods.getValues reference, NOT the whole formMethods object.
    }, [isOpen, hourlyRate, formMethods.getValues])

The onInputChange handler (Change 1) is the SOLE live writer while the modal is open; the seed
effect only fires on open / hourlyRate change, so the two never run at the same time. Keep
parseHours locale-aware (do not hard-code a "." decimal separator).

After implementing, run the app and confirm: the text is correct on open before any interaction,
and typing into the Hours field does NOT drop characters or reset the input. Run the lint and the
existing test suite, and fix any failures.
```

</details>

---

serikjensen:
@cursoragent could you implement the feedback outlined in the comment above https://github.com/Gusto/embedded-react-sdk/pull/2241#pullrequestreview-4563744670

src/components/Contractor/Payments/CreatePayment/EditContractorPaymentPresentation.tsx
```diff
@@ -37,32 +37,32 @@ export const EditContractorPaymentPresentation = ({
...
+  }
+
+  const [hoursPayDescription, setHoursPayDescription] = useState('')
+
+  useEffect(() => {
+    if (isOpen) {
+      setHoursPayDescription(computeHoursPayDescription(formMethods.getValues('hours') ?? 0))
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [isOpen, hourlyRate])
```
serikjensen: We should update this to include the formMethods as part of the dependency array, we should still include exhaustive deps when possible


```suggestion
  useEffect(() => {
    if (isOpen) {
      setHoursPayDescription(computeHoursPayDescription(formMethods.getValues('hours') ?? 0))
    }
  }, [isOpen, hourlyRate, formMethods.getValues])
```

src/components/Common/UI/NumberInput/NumberInput.tsx
```diff
@@ -105,6 +111,9 @@ export function NumberInput({
             placeholder={placeholder}
             aria-describedby={ariaDescribedBy}
             isDisabled={isDisabled}
+            onInput={(event: React.FormEvent<HTMLInputElement>) => {
+              onInputChange?.(event.currentTarget.value)
+            }}
```
serikjensen: You should be able to just use onChange here. In React, onChange should fire on every input. That will allow us to not need to whitelist this prop for the Input and keep the API surface smaller

serikjensen:
@cursoragent could you implement the latest requested changes around 
* Removing onInput from the input api
* Using onChange instead
* adding formMethods.getValues to the dependency array to avoid needing to disable exhaustive deps

serikjensen (approved):

PR #2245, opened 2026-06-23
refactor(StateTaxes): flatten form into a single evenly-spaced stack

aaronlee777:
## Summary
- Drop per-requirement-set heading and effective-date subtext from the form
- Render every input as a sibling of one outer `Flex` with `gap={20}` so spacing is uniform across the form
- Use `flatMap` to concatenate requirements across sets while preserving the set-prefixed RHF field key

## Test plan
- [ ] Open a state taxes form for a state with a single requirement set; verify inputs are stacked with even 20px gaps
- [ ] Open a state taxes form for a state with multiple requirement sets; verify inputs from different sets sit as siblings with the same spacing
- [ ] Submit the form; confirm the request body still groups requirements by set key


serikjensen:
Closing, aaron will give this another look with a fresh start down the road

PR #2248, opened 2026-06-23
feat(StateTaxes): respect applicable_if for conditional field visibility

aaronlee777:
## Summary
- Honor each `TaxRequirement.applicable_if` constraint when rendering the State Taxes form
- Add `isRequirementApplicable` helper plus unit tests
- Drop inapplicable requirements from the submit payload so stale values aren't re-asserted

## Context
The embedded API returns each `TaxRequirement` with an `applicable_if: Array<{key, value}>` field that gates visibility on the answers to other requirements in the same set. The SDK previously ignored it and rendered every requirement unconditionally — for example, Idaho's `taxrates` set returns three rate inputs that are only meant to show when the user has answered "No, my agency gave me new rates" on a parent radio, but the SDK rendered them regardless. gws-flows has honored this contract via a small JS file for years.

## Demo

Screenshot 2026-06-23 at 5 45 32 PM
Screenshot 2026-06-23 at 5 45 38 PM

## Implementation
- New helper `src/components/Company/StateTaxes/StateTaxesForm/applicableIf.ts` evaluates `applicableIf` against the live RHF form values, prefix-aware (constraints reference the raw API key; form values are stored under `setKey.toRhfKey(reqKey)`).
- `Form.tsx` adds `useWatch({ control })` and filters each requirement before rendering.
- The submit handler in `StateTaxesForm.tsx` reuses the helper to drop inapplicable requirements from the payload.
- No Zod schema changes needed — every field was already `.optional()`.

## Test plan
- [x] Unit tests for `isRequirementApplicable` (empty / matching / non-matching / AND-logic / missing set / pipe-key round-trip)
- [x] Existing State Taxes tests pass; the WA "renders tax rate fields" test was implicitly relying on the bug and is updated to assert the corrected behavior (rate input hidden until the radio toggles to "No")
- [x] In-browser sanity check on Idaho via `npm run sdk-app` against the demo company

### DIFF
--- a/src/components/Company/StateTaxes/StateTaxesForm/Form.tsx
+++ b/src/components/Company/StateTaxes/StateTaxesForm/Form.tsx
@@ -1,7 +1,9 @@
 import { useTranslation } from 'react-i18next'
 import { Fragment } from 'react/jsx-runtime'
+import { useFormContext, useWatch } from 'react-hook-form'
 import { useStateTaxesForm } from './context'
 import { toRhfKey } from './rhfKey'
+import { isRequirementApplicable, type StateTaxesFormValues } from './applicableIf'
 import { QuestionInput } from '@/components/Common/TaxInputs/TaxInputs'
 import { useLocaleDateFormatter } from '@/contexts/LocaleProvider/useLocale'
 import { useComponentContext } from '@/contexts/ComponentAdapter/useComponentContext'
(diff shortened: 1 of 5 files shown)


serikjensen (commented):
We're going to close this one and create a ticket so that we can make sure there's some dedicated eng attention

aaronlee777:
Closing — tracking this work via a dedicated JIRA ticket instead.

PR #2254, opened 2026-06-24
fix(PayrollOverview): correct loading copy and gating in the run-payroll flow

aaronlee777:
## Summary

Two related fixes to the loading states users see when entering and submitting a payroll:

- **PayrollOverview** previously surfaced "Submitting payroll..." in two wrong situations: while initially fetching any payroll (the data-fetch loader reused the submit overlay's i18n keys) and while polling an already-submitting payroll on load (the overlay was gated on \`isPolling\`, which also fires for server-side processing of an already-submitted payroll). Added a dedicated \`dataLoadingTitle\` ("Loading payroll...") for the initial fetch and gated the submit overlay on a new \`hasSubmittedInSession\` flag so it only renders for a submit the current user just initiated.
- **PayrollExecutionFlowContextual** had a Suspense boundary with no fallback, so users saw a blank screen during the upstream payroll fetch when clicking "Review and Submit". Added a basic \`<Loading />\` skeleton as the fallback. The destination component (PayrollOverview or PayrollConfiguration) still owns its own titled loading state once mounted.

## Test plan

- [ ] \`npm run test -- --run src/components/Payroll/PayrollOverview src/components/Payroll/PayrollFlow\` passes
- [ ] Storybook / sdk-app: open a calculated payroll via "Review and Submit" → expect Loading skeleton → "Loading payroll..." → content
- [ ] Click Submit → expect "Submitting payroll..." overlay during the mutation and the post-submit polling window
- [ ] Reload while a payroll is mid-processing → expect normal review UI, not the "Submitting payroll..." overlay
- [ ] Already-submitted payrolls open instantly with the right copy

### DIFF
--- a/src/components/Payroll/PayrollOverview/PayrollOverview.test.tsx
+++ b/src/components/Payroll/PayrollOverview/PayrollOverview.test.tsx
@@ -196,6 +196,33 @@ describe('PayrollOverview polling', () => {
   })
 })
 
+describe('PayrollOverview submit-in-progress overlay', () => {
+  const mockOnEvent = vi.fn()
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+    mockPayrollData = { ...basePayrollData }
+  })
+
+  it('does not show the "Submitting payroll" overlay when loading a payroll whose server-side status is already submitting', async () => {
+    mockPayrollData = {
+      ...basePayrollData,
+      processed: false,
+      processingRequest: { status: 'submitting', errors: [] },
+    }
+
+    renderWithProviders(
+      <PayrollOverview companyId="company-uuid" payrollId="payroll-uuid" onEvent={mockOnEvent} />,
+    )
+
+    expect(await screen.findByText(/Review payroll/i)).toBeInTheDocument()
+    // The "Submitting payroll" overlay is only correct when the current user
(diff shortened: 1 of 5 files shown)


src/components/Payroll/PayrollOverview/PayrollOverview.test.tsx
```diff
@@ -196,6 +196,33 @@ describe('PayrollOverview polling', () => {
...
+      <PayrollOverview companyId="company-uuid" payrollId="payroll-uuid" onEvent={mockOnEvent} />,
+    )
+
+    expect(await screen.findByText(/Review payroll/i)).toBeInTheDocument()
+    // The "Submitting payroll" overlay is only correct when the current user
+    // just clicked Submit. A page load against an already-processing payroll
+    // must not block the review UI behind it.
+    expect(screen.queryByText(/Submitting payroll/i)).not.toBeInTheDocument()
+  })
+})
```
serikjensen: I don't love that this is verifying against text that's not there. there's an infinite amount of things that could not be on the page. My rec would be to have claude update this test to assert on the correct loading behavior that this PR is introducing rather than asserting that the submission text is absent

serikjensen (approved):

PR #2270, opened 2026-06-25
fix(SDK-137): hide bank-account list and Split button on Check toggle in onboarding

jeffredodd:
## Summary

- The onboarding `ListView` was gating the bank-account `DataView` and the **Split paycheck** button on `bankAccounts.length` alone. The management `PaymentMethodCard` already gates correctly on `isDirectDeposit && length`.
- The old server behavior masked this — switching from Direct Deposit to Check deleted bank accounts server-side, so the next refetch returned `[]` and the inconsistency disappeared on its own.
- [SDK-137](https://gustohq.atlassian.net/browse/SDK-137) decouples payment type from payment methods (server-side, behind a company-level FF, gradual rollout). Bank accounts now persist when switching to Check, so the missing local gate becomes user-visible: a Check-selected user keeps seeing the bank-account table and the Split button.
- Aligns the onboarding gates with the management card and adds a test that switching to Check never issues a delete-bank-account mutation.

## Files

- `src/components/Employee/PaymentMethod/onboarding/ListView.tsx` — add `watchedType === PAYMENT_METHODS.directDeposit` to the `DataView` and Split-button gates
- `src/components/Employee/PaymentMethod/onboarding/ListView.test.tsx` — two new cases: list hides on Check toggle, Split button hides on Check toggle
- `src/components/Employee/PaymentMethod/shared/usePaymentMethodForm/usePaymentMethodForm.test.tsx` — new case: switching to Check does not call `DELETE /employees/:id/bank_accounts/:id`

## Test plan

- [x] `npm run test -- --run src/components/Employee/PaymentMethod` — 91/91 pass
- [ ] Storybook smoke: open the onboarding PaymentMethod story with bank accounts present; toggle to Check; confirm the list and Split button hide. Toggle back to Direct Deposit; confirm they reappear from the same client-side data without a network round-trip.
- [ ] Manual (optional, if a flagged-on test company is available): in `sdk-app`, switch an employee DD → Check → DD; confirm bank accounts return intact.

[SDK-137]: https://gustohq.atlassian.net/browse/SDK-137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/Employee/PaymentMethod/onboarding/ListView.test.tsx
+++ b/src/components/Employee/PaymentMethod/onboarding/ListView.test.tsx
@@ -117,6 +117,22 @@ describe('PaymentMethod onboarding ListView', () => {
     expect(screen.queryByRole('button', { name: /add.*bank account/i })).not.toBeInTheDocument()
   })
 
+  it('hides bank account list when Check is selected even if accounts exist server-side', async () => {
+    const user = userEvent.setup()
+    renderWithProviders(<PaymentMethod employeeId="employee-123" onEvent={onEvent} />)
+
+    await waitFor(
+      () => {
+        expect(screen.getByText('Chase')).toBeInTheDocument()
+      },
+      { timeout: 5000 },
+    )
+
+    await user.click(screen.getByRole('radio', { name: /check/i }))
+
+    expect(screen.queryByText('Chase')).not.toBeInTheDocument()
+  })
+
   it('navigates to BankForm when Add bank account is clicked', async () => {
     const user = userEvent.setup()
     renderWithProviders(<PaymentMethod employeeId="employee-123" onEvent={onEvent} />)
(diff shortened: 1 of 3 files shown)


jeffredodd:
Closing without merging.

Research in Glean turned up that the new server behavior **is not on for embedded/GEP companies today**. PR [Gusto/zenpayroll#327217 (PAYPROF-524)](https://github.com/Gusto/zenpayroll/pull/327217) shipped the preserve-on-pay-by-check behavior for core Gusto but explicitly kept legacy `destroy_all!` behavior on the embedded path via `EmployeePaymentMethodService`, gated behind `emb_preserve_payment_methods_on_check` (default off, per-partner migration).

That means:
- The visual bug this PR was fixing (bank accounts visible after toggling to Check) **does not manifest for embedded partners today** — the server still deletes on submit and the next refetch returns `[]`.
- The SDK as it stands today is correct against the production rollout state.
- If/when a partner opts into the new behavior, this fix becomes necessary — but at that point the right approach is more likely a full decoupling of bank-account UI from payment type (parity with the contractor experience), not a status-quo gate tightening.

SDK-137 will be closed as resolved against the current rollout state. If/when the embedded flag flips, we'll revisit with a fuller UX.

PR #2271, opened 2026-06-25
fix(ContractorPayments): use SDK DatePicker for payment date input

aaronlee777:
## Summary
- Swap the raw `TextInput type="date"` in `CreatePaymentPresentation` for the adapter's `DatePicker` so the payment date field renders the SDK's date input + calendar picker.
- Convert at the presentation boundary between the parent's `YYYY-MM-DD` string state and `Date | null` using the existing `normalizeToDate` / `normalizeDateToLocal` / `formatDateToStringDate` helpers — same pattern `DatePickerField` uses internally. Parent state and the `RFCDate(paymentDate)` API contract are unchanged.

## Test plan
- [ ] Verify the payment date field renders the SDK calendar picker in Storybook / sdk-app
- [ ] Pick a date and confirm `checkDate` in the preview/create request matches the selection
- [ ] Confirm the initial value (today + payment speed business days) still populates correctly

### DIFF
--- a/src/components/Contractor/Payments/CreatePayment/CreatePaymentPresentation.tsx
+++ b/src/components/Contractor/Payments/CreatePayment/CreatePaymentPresentation.tsx
@@ -12,6 +12,11 @@ import { HamburgerMenu } from '@/components/Common/HamburgerMenu'
 import { useI18n } from '@/i18n'
 import { formatHoursDisplay } from '@/components/Payroll/helpers'
 import useNumberFormatter from '@/hooks/useNumberFormatter'
+import {
+  formatDateToStringDate,
+  normalizeDateToLocal,
+  normalizeToDate,
+} from '@/helpers/dateFormatting'
 
 const ZERO_HOURS_DISPLAY = '0.000'
 
@@ -50,7 +55,7 @@ export const CreatePaymentPresentation = ({
   isLoading,
   paymentSpeedDays,
 }: ContractorPaymentCreatePaymentPresentationProps) => {
-  const { Button, Text, Heading, TextInput, Alert } = useComponentContext()
+  const { Button, Text, Heading, DatePicker, Alert } = useComponentContext()
   useI18n('Contractor.Payments.CreatePayment')
   const { t } = useTranslation('Contractor.Payments.CreatePayment')
   const currencyFormatter = useNumberFormatter('currency')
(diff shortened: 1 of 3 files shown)


aaronlee777:
@cursoragent  Can you fix the failing contractor e2e test?

src/components/Contractor/Payments/CreatePayment/CreatePaymentPresentation.tsx
```diff
@@ -110,10 +115,12 @@ export const CreatePaymentPresentation = ({
...
-        <TextInput
-          type="date"
-          value={paymentDate}
-          onChange={onPaymentDateChange}
+        <DatePicker
+          value={paymentDate ? normalizeToDate(paymentDate) : null}
+          onChange={date => {
+            const normalized = normalizeDateToLocal(date)
+            onPaymentDateChange(normalized ? (formatDateToStringDate(normalized) ?? '') : '')
+          }}
```
serikjensen, line 123: @mariechatfield do you know if this is the correct date normalization process here? wondering why these functions would be needed in this case and not elsewhere?

mariechatfield: Let me try to reload that context. For anyone else following along, #1764 has my recommendations for date normalization and #1767 has a bug fix for date pickers

mariechatfield: Yes, this is correct and necessary — and we can prove it's necessary by adding a unit test here mirroring the UTC timezone tests that cover `DatePickerField` (the ones validating that dates in UTC+ timezones don't roll back a day). Everything else in the codebase uses `DatePickerField`, which handles this normalization internally, so those cases are covered there.

Ideally we'd use `DatePickerField` here too — but we can't, because this isn't a form context. `DatePickerField` requires `FormProvider`, and `paymentDate` is plain `useState` in the container, not a form field.

The better version of this would be a shared adapter function (`stringToDatePickerValue` / `datePickerValueToString` or similar) that both `DatePickerField` and this component use, rather than recreating the same three-function chain. That way there's one place to update and one place to test. Worth a follow-up if this pattern shows up elsewhere.

For now, I'd recommend adding a comment here pointing future readers to `DatePickerField` as the canonical example, and adding the timezone unit test so this stays in sync.

*— via Claude*

serikjensen (approved):

PR #2261, opened 2026-06-24
feat(contractor)!: migrate ContractorProfile to useContractorDetailsForm

serikjensen:
## Summary

Migrates `ContractorProfile` to a new headless `useContractorDetailsForm` hook (mirroring `useEmployeeDetailsForm`), powering individual/business + admin/self-onboarding flows from a single source. The legacy `useContractorProfile` implementation is removed.

- **New hook `useContractorDetailsForm`**: supports all contractor create/update fields with conditional requiredness, redacted-PII handling (`ssn`/`ein`), the self-onboarding toggle, and a schema-driven `Fields` surface. Base `requiredFieldsConfig` mirrors the API (create-required vs never), with `email` required via a functional predicate when self-onboarding is on.
- **Field exclusion**: inapplicable fields are dropped from both validation and render via a single `getExcludedContractorFields` array helper, forked on individual vs business. Field visibility derives from "not excluded" (email is the lone exception). The resolver builds its per-validation schema from the submitted values, decoupling schema generation from `useWatch`.
- **`ContractorProfile` integration**: consumes the hook, layering SDK-stricter requiredness via a static `optionalFieldsToRequire` superset for backwards compatibility.
- **Tests/docs**: robust schema-level + hook + component tests; updated `useContractorDetailsForm.md` and regenerated the public API report.

## Breaking change

`ContractorProfile`'s `defaultValues` prop shape changed: `contractorType` → `type`, and `startDate` is now an ISO date `string` (was `Date`). See the `BREAKING CHANGE:` footer on commit `d177311` for the migration snippet.

## Commits

1. `feat(contractor)!: migrate ContractorProfile to useContractorDetailsForm`
2. `refactor(contractor): simplify field exclusion to a single array helper`

## Test plan

- [ ] `npm run test -- --run src/components/Contractor/Profile` (61 passing locally)
- [ ] `npx tsc --noEmit` clean
- [ ] `npm run lint:check` (no new errors)
- [ ] Manually inspect the diff for the `defaultValues` breaking change + migration guidance

Made with [Cursor](https://cursor.com)

PR #2272, opened 2026-06-25
feat(forms): support a value-aware function for buildFormSchema excludeFields

serikjensen:
## Summary

Widens the internal `buildFormSchema` `excludeFields` option to accept **either**:

- an `Array<keyof T>` — existing behavior, removes the field from the schema at build time, **or**
- a function `(data, mode) => Array<keyof T>` — evaluated during the required-check `superRefine`, so a schema can skip requiredness checks for fields that don't currently apply (e.g. business-only fields while an individual contractor is selected).

The function form keeps fields in the schema shape (validated as optional) and keeps requiredness **static and promotable** via `optionalFieldsToRequire`. This is the foundation the upcoming `useContractorDetailsForm` hook builds on for individual/business applicability gating.

## Why

Contractor details need fields to appear/disappear based on `type`/`wageType`/self-onboarding while remaining promotable to required via the public `optionalFieldsToRequire` API. Doing this purely with functional `requiredFieldsConfig` predicates would lose promotability; a value-aware `excludeFields` keeps both.

## Test plan

- `npm run test -- --run src/partner-hook-utils/form/buildFormSchema.test.ts` (62 passing; adds a `excludeFields as a value-aware function` block)
- `npx tsc --noEmit` clean
- Array-form `excludeFields` behavior unchanged (regression covered by existing tests)

First of a 4-PR stack decomposing #2261.

Made with Cursor

### DIFF
--- a/src/partner-hook-utils/form/buildFormSchema.test.ts
+++ b/src/partner-hook-utils/form/buildFormSchema.test.ts
@@ -301,6 +301,100 @@ describe('buildFormSchema', () => {
     })
   })
 
+  describe('excludeFields as a value-aware function', () => {
+    const fieldValidators = {
+      kind: z.enum(['individual', 'business']),
+      firstName: z.string(),
+      businessName: z.string(),
+    }
+
+    const requiredFieldsConfig = {
+      firstName: 'create',
+      businessName: 'create',
+    } as const
+
+    // firstName applies only to individuals; businessName only to businesses.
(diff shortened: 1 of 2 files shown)


serikjensen:
@cursoragent is there a way to avoid the following cast: new Set(excludeFieldsFn(data as { [K in keyof T]: z.infer<T[K]> }, mode).map(String))

mariechatfield (approved):

PR #2273, opened 2026-06-25
test(contractor): add ContractorProfile behavior characterization tests

serikjensen:
## Summary

Adds a component-behavior regression net for `ContractorProfile` that locks in the durable contract **before** the migration to `useContractorDetailsForm`:

- conditional field rendering (business/individual swap, hourly rate, self-onboarding revealing email and hiding SSN/EIN)
- submit-time validation blocking the create/update mutation when required fields are empty
- redacted-PII handling
- emitted `onEvent` payloads

These tests pass against the **current (legacy) implementation as-is** — they required no calibration — and are intended to stay **green and unedited** through the migration PR, proving the migration preserves behavior. Any discrepancy that surfaces later is resolved by adjusting the hook/component to match this contract.

The exhaustive schema-level validation matrix lives alongside the hook in a later PR (`contractorDetailsSchema.test.ts`).

## Test plan

- `npm run test -- --run src/components/Contractor/Profile/ContractorProfile.test.tsx` (9 passing against legacy)
- `npx tsc --noEmit` clean

Second of a 4-PR stack decomposing #2261. Stacked on #2272.

Made with Cursor


serikjensen:
@cursoragent this "'Layer B — component behavior'" description for this test seems not great or descriptive. can we make sure the tests make sense? i think the agent that last worked on this was indexing on the work context instead of making this readable for future developers

mariechatfield (approved):

PR #2275, opened 2026-06-25
feat(contractor): add useContractorDetailsForm hook

serikjensen:
## Summary

Adds a headless `useContractorDetailsForm` hook (mirroring `useEmployeeDetailsForm`) that powers contractor create/update for both **individual** and **business** types from a single source. It is exported publicly but **not yet consumed** by `ContractorProfile` (that integration is the next PR).

- All contractor create/update fields with conditional requiredness (base `requiredFieldsConfig` mirrors the API; `email` required via a functional predicate when self-onboarding is on).
- Redacted-PII handling for `ssn`/`ein`.
- Self-onboarding toggle management.
- Schema-driven `Fields` surface with applicability gating via the value-aware `excludeFields` function from #2272 — inapplicable fields are skipped during validation while staying promotable through the public `optionalFieldsToRequire`.
- Widens the public `ContractorSelfOnboardingStatuses` set type to `Set<ContractorOnboardingStatus1>` for membership checks, and drops a now-unused `@ts-expect-error` in the legacy hook (the legacy hook is removed entirely in the integration PR).
- Docs (`useContractorDetailsForm.md`, hooks index) and regenerated public API report.

## Test plan

- `npm run test -- --run src/components/Contractor/Profile/shared/useContractorDetailsForm` (52 passing: schema matrix + hook behavior)
- `npx tsc --noEmit` clean
- `npm run lint` clean on changed files
- API report regenerated via `npm run build && npm run api-report:derive`

Third of a 4-PR stack decomposing #2261. Stacked on #2273.

Made with Cursor


dmortal (approved):

.reports/embedded-react-sdk.api.md
```diff
@@ -1108,6 +1110,119 @@ interface ConfirmWireDetailsProps extends BaseComponentInterface<'Payroll.Confir
...
+    Type: typeof ContractorTypeField;
+    WageType: typeof ContractorWageTypeField;
+    WorkState: typeof ContractorWorkStateField | undefined;
+}
+
+// @public
+export type ContractorDetailsFieldsMetadata = UseContractorDetailsFormReady['form']['fieldsMetadata'];
+
+// @public
+export type ContractorDetailsFormData = { type: "Business" | "Individual"; wageType: "Fixed" | "Hourly"; startDate: string; hourlyRate: number; selfOnboarding: boolean; fileNewHireReport: boolean; email: string; firstName: string; lastName: string; middleInitial: string; businessName: string; workState: string; ssn: string; ein: string; };
```
mariechatfield, line 1175: nice, love to see that the rewriting step is working 🚀

.reports/embedded-react-sdk.api.md
```diff
@@ -1270,6 +1433,33 @@ adjustForMinimumWage: z.ZodBoolean;
 minimumWageId: z.ZodString;
 }>;
 
+// Warning: (ae-forgotten-export) The symbol "ContractorDetailsSchemaOptions" needs to be exported by the entry point index.d.ts
+// Warning: (ae-internal-missing-underscore) The name "createContractorDetailsSchema" should be prefixed with an underscore because the declaration is marked as @internal
+//
+// @internal (undocumented)
+export function createContractorDetailsSchema(options?: ContractorDetailsSchemaOptions): BuildFormSchemaResult<    {
+type: z.ZodEnum<{
```
mariechatfield: Did you mean to export this?

docs/hooks/useContractorDetailsForm.md
mariechatfield, line 1: I don't think we want to create new hand-authored hooks files, was this on purpose?

src/index.ts
```diff
@@ -281,6 +281,63 @@ export type {
   EmployeeDetailsFields,
 } from '@/components/Employee/Profile/shared/useEmployeeDetailsForm'
 
+export {
+  useContractorDetailsForm,
+  ContractorType,
+  WageType,
+  ContractorDetailsErrorCodes,
+  createContractorDetailsSchema,
```
mariechatfield: I think this is the line we need to delete because it's internal only

mariechatfield (changes requested):
Can you remove the internal only export for the schema? otherwise LGTM!

docs/hooks/useContractorDetailsForm.md
mariechatfield: Oh I should say: if there's prose you want to add to the hook that doesn't belong in the tsdoc remarks, you can add it as `GUIDE.md` inside the hook directly and it'll slot into the generated reference

here's an example (the comments tell the autogenerator whether the prose goes at the top or bottom of the page. so `## heading <!-- slot: overview -->` will be put near the top and `## heading <!-- slot: appendix -->` at the bottom) https://github.com/Gusto/embedded-react-sdk/blob/main/src/components/Employee/StateTaxes/shared/useEmployeeStateTaxesForm/GUIDE.md

serikjensen:
@cursoragent could you integrate the feedback above from @mariechatfield? we want to remove the export for the createContractorDetailsSchema. we also want to remove the hand documented hooks file that was added for this

docs/hooks/useContractorDetailsForm.md
serikjensen: claude skill holdover from before, will remove!

src/index.ts
serikjensen: yup, will delete!

.reports/embedded-react-sdk.api.md
serikjensen: likely unintentional, will remove and then update the skill to not export

mariechatfield (approved):

PR #2280, opened 2026-06-25
feat(Contractor.Address): add useContractorAddressForm hook

serikjensen:
## Summary

Second PR in the `Contractor.Address` hook migration stack. Introduces the public **`useContractorAddressForm`** hook following the SDK hook architecture, with no changes to the existing component yet (that's PR3).

- Fetches the contractor and its single address; operates in **update-only** mode (a contractor always has exactly one address)
- Exposes the standard `data` / `status` / `actions` / `errorHandling` / `form` surface, plus `contractorType` so consumers can label the address as home (Individual) or business (Business)
- State options localize via `getOptionLabel`
- **Required-field config mirrors the API contract**: every address field is optional by default (only `version` is required, and it is not a form field). Consumers needing stricter requiredness opt in via `optionalFieldsToRequire`
- `zip` validates format (`INVALID_ZIP`) whenever a value is present, independent of requiredness
- Exported from the SDK root barrel as a partner-facing hook

## Stack

1. PR1 — test coverage
2. **PR2 (this)** — the hook → stacked on PR1
3. PR3 — integrate the hook into the component

## Test plan

- [x] `npm run test -- --run src/components/Contractor/Address/shared/useContractorAddressForm/useContractorAddressForm.test.tsx`
- [x] `npm run tsc`


jeffredodd (approved):

serikjensen:
@cursoragent could you resolve conflicts and bring this up to date?

PR #2302, opened 2026-06-29
feat(contractor): add useContractorPaymentMethodForm hook

serikjensen:
## Summary

- Adds `useContractorPaymentMethodForm`, a headless React Hook Form hook for a contractor's payment method type (Direct Deposit vs Check).
- On submit it `PUT`s the payment method. The Direct Deposit path defers actual bank-account creation to `useContractorBankAccountForm`, so this hook owns only the type selection and the Check write.
- Exposes the `TypeField` subcomponent and types via the public surface.

## Notes

- This is part 2 of a 3-PR split. PR1 (#2301) adds `useContractorBankAccountForm`; PR3 integrates both into the `PaymentMethod` component (and will supersede #2293).

## Test plan

- [x] `useContractorPaymentMethodForm` unit + schema tests pass (6/6)
- [x] `npm run derive` regenerates API report + docs

Made with [Cursor](https://cursor.com)

PR #2301, opened 2026-06-29
feat(contractor): add useContractorBankAccountForm hook

serikjensen:
## Summary

- Adds `useContractorBankAccountForm`, a headless React Hook Form hook for creating/replacing a contractor's bank account (name, routing number, account number, account type).
- The `POST /v1/contractors/:uuid/bank_accounts` endpoint is a full **replace** — `account_number` is required on every write. The API treats the exact masked token (e.g. `XXXX1207`) as a "keep existing account" sentinel. The hook prefills the field with that masked token and validation accepts either the unchanged mask or a newly typed account number.
- Creating a bank account also flips the contractor to Direct Deposit as a server-side side effect.
- Exposes Field subcomponents and types via the public surface.

## Notes

- This is part 1 of a 3-PR split. PR2 adds `useContractorPaymentMethodForm`; PR3 integrates both into the `PaymentMethod` component (and will supersede #2293).

## Test plan

- [x] `useContractorBankAccountForm` unit + schema tests pass (10/10)
- [x] `npm run derive` regenerates API report + docs

Made with [Cursor](https://cursor.com)

PR #2293, opened 2026-06-26
feat(contractor): migrate PaymentMethod to useContractorPaymentMethodForm hook

serikjensen:
## Summary

Migrates the Contractor PaymentMethod component from its monolithic inline-form pattern to the hook-based architecture (following `/migrate-sdk-component-to-hooks`). Single screen, single hook.

- Adds `useContractorPaymentMethodForm` — one headless hook for the combined payment-type + bank-account form. Uses a value-aware `excludeFields` function so the bank fields stay in `requiredFieldsConfig` (promotable) but are dropped from required validation when `type === 'Check'`, mirroring the `useContractorDetailsForm` precedent.
- Rewrites `PaymentMethod.tsx` to `BaseBoundaries` + thin `Root` + `SDKFormProvider`, preserving the event surface (`contractor/bankAccount/created`, `contractor/paymentMethod/updated`, `contractor/paymentMethod/done`) and the masked-account-number behavior.
- Hardens pre-migration test coverage (`PaymentMethod.test.tsx`) and adds hook/schema unit tests.
- Exports the hook, schema, error codes, and fields from `src/index.ts`.
- Removes the now-unused `PaymentTypeForm` / `BankAccountForm` and the inline schema.

## Notes

- Faithfully preserves current behavior, including the original "Direct Deposit always POSTs a bank account on submit" sequence. A follow-up is open for discussion: aligning the Direct Deposit path with `gws-flows` (drop the redundant `PUT /payment_method` since the bank-account POST already updates the payment method) and optionally gating the POST on actual field changes.

## Test plan

- [x] `npm run test -- --run` (full suite green)
- [x] `npm run lint:check` (0 errors)
- [ ] Manual verification in the SDK dev app / Storybook

Made with [Cursor](https://cursor.com)

PR #2303, opened 2026-06-29
feat(contractor): integrate payment-method and bank-account hooks into PaymentMethod

serikjensen:
## Summary

- Refactors the `PaymentMethod` component to compose `useContractorPaymentMethodForm` and `useContractorBankAccountForm` via `composeSubmitHandler`, replacing the old `PaymentTypeForm` / `BankAccountForm` internals.
- Direct Deposit submits through the bank-account form (which creates/replaces the bank account and flips the contractor to Direct Deposit server-side); Check submits through the payment-method form. The masked account number is prefilled and sent as the keep-existing sentinel so an unchanged Direct Deposit preserves the account.
- Adds a dedicated SCSS module to give the section full width.

## Notes

- **Supersedes #2293**, which will be closed.
- This PR is **stacked on top of #2301 (`useContractorBankAccountForm`) and #2302 (`useContractorPaymentMethodForm`)**. Because it targets `main`, it currently shows those hooks' code as well; once #2301 and #2302 merge, rebase this branch on `main` and the diff will narrow to the integration changes.

## Test plan

- [x] `PaymentMethod` component + both hook test suites pass (26/26)
- [x] `npm run derive` regenerates API report + docs

https://github.com/user-attachments/assets/47282261-0320-4c20-a525-c673670dbbb6


serikjensen:
@cursoragent you should be able to bring this one up to date with main. The PRs that is based on for the useContractorBankAccount and payment method hooks have been merged so you should be able to base the integration off of that

src/components/Contractor/PaymentMethod/PaymentMethod.tsx
```diff
@@ -72,142 +52,138 @@ type PaymentMethodSchemaInputs = z.input<typeof PaymentMethodSchema>
...
  * ```
  */
-export function PaymentMethod(props: PaymentMethodProps) {
+export function PaymentMethod({ dictionary, ...props }: PaymentMethodProps) {
+  useComponentDictionary('Contractor.PaymentMethod', dictionary)
   return (
-    <BaseComponent {...props}>
-      <Root {...props}>{props.children}</Root>
-    </BaseComponent>
+    <BaseBoundaries componentName="Contractor.PaymentMethod">
```
mariechatfield, line 58: Does this mean that we won't handle `LoaderComponent` if you provide it? I thought that was in base boundaries

mariechatfield (approved):
Nice, I was able to use the payment method section in contractor onboarding in the SDK app and it looks the same as it was before!

src/components/Contractor/PaymentMethod/PaymentMethod.tsx
serikjensen: Good callout! it still pulls it from the top level gusto provider which can get set universally. per component overrides of loading have been lost with this setup which is an oversight

Created this ticket to address it https://gustohq.atlassian.net/browse/SDK-1074

Basically we just need to update BaseLayout which is already being used in the hook components to take a loading indicator. We then need to update each of those components to pass the loading indicator through from the base props to the base layout

PR #2336, opened 2026-07-02
feat!: promotable conditional requiredness for form schemas

serikjensen:
## Summary

- Broadens `buildFormSchema`'s `optionalFieldsToRequire` to accept predicate ("sometimes") fields, letting consumers promote them from conditionally required to always required — mirroring how `'never'` fields already promote. The runtime already supported the OR semantics (`predicate: isPartnerRequired ? () => true : effectiveRule`); this unlocks it at the type level (`OptionalOnCreate` / `OptionalOnUpdate` now include a `PredicateRule`).
- Converts employee `email` to the predicate `data => data.selfOnboarding`, so its self-onboarding requirement is an API-aligned baseline that drives both validation and `fieldsMetadata.isRequired` (keeping the rendered label and validation in sync). `AdminProfile` and `ProfileEditForm` continue to promote `email` so it stays always-required in those surfaces.
- Removes the now-redundant `superRefine` block and the `EMAIL_REQUIRED_FOR_SELF_ONBOARDING` error code; regenerated derived docs/API report and updated the hand-written `docs/hooks/*.md`.

## Breaking change

`EMAIL_REQUIRED_FOR_SELF_ONBOARDING` is removed from `EmployeeDetailsErrorCodes`, and the `Fields.Email` `validationMessages` type is narrowed to `REQUIRED | INVALID_EMAIL`. An empty email while self-onboarding is enabled now emits `REQUIRED` instead of `EMAIL_REQUIRED_FOR_SELF_ONBOARDING`.

## Test plan

- [x] `npx tsc --noEmit`
- [x] `npm run test -- --run src/partner-hook-utils/form src/components/Employee/Profile src/components/Contractor/Profile` (258 passing)
- [x] `npm run lint:check` (0 errors)
- [x] `npm run format:check` (changed files clean)
- [x] New tests: predicate-field promotion (framework) + self-onboarding email baseline (hook), promoted vs. unpromoted
- [x] CI green

## Demo

To verify functionality, here is a version of profile with email removed from the optional fields to require. Observe that email is optional until self onboarding is toggled

https://github.com/user-attachments/assets/91c87878-c46a-48e8-b2f4-94fdce60ad88

Now restoring existing behavior promoting email to always be required

https://github.com/user-attachments/assets/16b0d950-7287-4a41-badf-870fcb43d77f

### DIFF
--- a/.reports/embedded-react-sdk.api.md
+++ b/.reports/embedded-react-sdk.api.md
@@ -1127,7 +1127,7 @@ export interface CompensationFormFields {
 }
 
 // @public
-export type CompensationOptionalFieldsToRequire = { create?: "title"[] | undefined; update?: ("title" | "flsaStatus" | "paymentUnit" | "rate" | "effectiveDate")[] | undefined; };
+export type CompensationOptionalFieldsToRequire = { create?: ("title" | "minimumWageId")[] | undefined; update?: ("title" | "flsaStatus" | "paymentUnit" | "rate" | "effectiveDate" | "minimumWageId")[] | undefined; };
 
 // @public
 interface CompensationProps extends BaseComponentInterface<'Employee.Compensation'> {
@@ -1673,7 +1673,7 @@ export interface ContractorDetailsFormFields {
 export type ContractorDetailsNameValidation = (typeof ContractorDetailsErrorCodes)['REQUIRED' | 'INVALID_NAME'];
 
 // @public
-export type ContractorDetailsOptionalFieldsToRequire = { create?: ("ssn" | "ein" | "middleInitial")[] | undefined; update?: ("hourlyRate" | "businessName" | "ssn" | "ein" | "startDate" | "firstName" | "lastName" | "middleInitial" | "workState")[] | undefined; };
+export type ContractorDetailsOptionalFieldsToRequire = { create?: ("ssn" | "ein" | "email" | "middleInitial")[] | undefined; update?: ("hourlyRate" | "businessName" | "ssn" | "ein" | "startDate" | "email" | "firstName" | "lastName" | "middleInitial" | "workState")[] | undefined; };
 
 // @public
 export type ContractorDetailsRequiredValidation = typeof ContractorDetailsErrorCodes.REQUIRED;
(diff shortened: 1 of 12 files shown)


docs/hooks/useEmployeeDetailsForm.md
mariechatfield, line 1: I thought I deleted all the docs/hooks directory. Is this an older branch or did I just actually forget to clean them up?

src/components/Employee/Profile/shared/useEmployeeDetailsForm/employeeDetailsSchema.ts
```diff
@@ -80,13 +74,25 @@ export type EmployeeDetailsFormOutputs = EmployeeDetailsFormData
...
+// and validation always agree. Consumers that need email required regardless of
+// self-onboarding promote it via `optionalFieldsToRequire`.
 const requiredFieldsConfig = {
   firstName: 'create',
   lastName: 'create',
   middleInitial: 'never',
-  email: 'never',
   dateOfBirth: 'never',
   ssn: 'never',
+  email: data => data.selfOnboarding,
```
mariechatfield, line 95: fancy

.reports/embedded-react-sdk.api.md
```diff
@@ -1127,7 +1127,7 @@ export interface CompensationFormFields {
 }
 
 // @public
-export type CompensationOptionalFieldsToRequire = { create?: "title"[] | undefined; update?: ("title" | "flsaStatus" | "paymentUnit" | "rate" | "effectiveDate")[] | undefined; };
+export type CompensationOptionalFieldsToRequire = { create?: ("title" | "minimumWageId")[] | undefined; update?: ("title" | "flsaStatus" | "paymentUnit" | "rate" | "effectiveDate" | "minimumWageId")[] | undefined; };
```
mariechatfield, line 1131: Did you expect for these changes to shake out of your type changes? It looks like `stateWcClassCode` is now also a valid field you can promote to required for `JobOptionalFieldsToRequire` and I thought the idea for that one is the user can never directly promote it, they can only ever promise `stateWcCovered` and _that_ will keep the class code in sync?

serikjensen: i think this is actually an expected result in this case! optional fields to require only lets you move in one direction and that is towards requiredness. so they can require things above and beyond the api requiredness but they can't require less. These are conditionally required fields that are now surfacing as optional to require.

I don't necessarily think it makes sense in a case where those fields get hidden (ex. compensation) so there is likely still work to be done here. We could think about including some sort of exclude, or think about having some indication of if it is a conditional requiredness because of conditional rendering.

In the meantime this at least fixes the broken state on profile where email couldn't be promoted to required always when it was conditionally required

docs/hooks/useEmployeeDetailsForm.md
serikjensen: sigh, i don't know where these keep resurfacing. i'll look to clean up and keep a better eye out for them

serikjensen: actually looks like all of these got restored at some point? https://github.com/Gusto/embedded-react-sdk/tree/main/docs/hooks lmk if you'd like me to put up a follow up to delete that dir

mariechatfield (approved):
Cool, as long as the other type changes make sense to me it looks good to me!

PR #2350, opened 2026-07-06
test(Contractor): add individual self-onboarding e2e canary

serikjensen:
## Summary

Adds a canary-style end-to-end test that drives a freshly seeded **individual** contractor through the full self-onboarding flow (Landing → Profile → Address → Payment → Documents → Summary) to the **"You're all set!"** screen. The driver is type-parameterized (`individual` now; `business` is a trivial fast-follow if an observed issue warrants it).

- **Harness** (`e2e/main.tsx`): wire the `contractor-self-onboarding` flow into the e2e app (`FlowType`, `contractorId` param, render case, flow selector option).
- **Scenario** (`e2e/scenarios/shared/onboarded.json`): seed a bare `selfindividual` contractor — no address, SSN, or onboarding status — so the flow has real state left to fill; expose `contractorIds.selfindividual` in `expectedContext`.
- **Driver** (`e2e/utils/contractorFlowDrivers.ts`): add `runContractorSelfOnboarding` with demo-robustness mitigations discovered while running against the live demo backend:
  - **Idempotent address entry** — skips filling when the address is already populated (retry / pre-seeded), which otherwise hung on the React Aria state picker.
  - **Documents step gates on the Continue button** (`hasSignedAllDocuments`) rather than "Sign document" affordances, and races "signable form appears" vs "bounced back to list" — the demo **auto-signs the W-9 asynchronously** after contractor creation, so eagerly opening the doc could land on an already-signed form.
- **`.gitignore`**: ignore `e2e/.e2e-scenarios.json`, a runtime provisioning artifact (contains demo flow tokens), alongside its sibling `e2e/.e2e-state.json`.

## Test plan

Verified locally against the demo backend with fresh provisioning:

```bash
E2E_DISABLE_CACHE=true npm run test:e2e:demo -- tests/contractor/94-self-onboarding-individual-submit.spec.ts
```

- [x] Passes cleanly on a fresh contractor — first attempt, no retry, no flaky (`1 passed (3.1m)`, flow body `26.4s`)
- [x] Self-skips under MSW mode (no scenario provisioning)
- [x] Lint + Prettier clean

> Note: because the demo backend auto-signs the W-9, this canary validates that the flow *reaches* completion; the manual W-9 signature path is a tolerant fallback rather than the guaranteed path. Deterministic W-9 signing UX is better asserted in Storybook/unit tests.


e2e/utils/contractorFlowDrivers.ts
```diff
@@ -224,6 +224,201 @@ export async function runBusinessContractorOnboarding(
...
+  const submitButton = page
+    .getByRole('button', { name: /^sign$/i })
+    .or(page.getByRole('button', { name: /^acknowledge$/i }))
+    .first()
+  const documentsHeading = page.getByRole('heading', { name: /^documents$/i })
+
+  await expect(submitButton.or(documentsHeading).first()).toBeVisible({ timeout: LONG_WAIT })
+
+  // Already bounced back to the list (auto-signed) — nothing to do here.
+  if (!(await submitButton.isVisible({ timeout: 1_000 }).catch(() => false))) {
```
mariechatfield: Why do some timeouts get `LONG_WAIT` but this one gets `1_000`?

mariechatfield (approved):
🚀  let's go e2e

e2e/utils/contractorFlowDrivers.ts
serikjensen: Good catch — that number was actually meaningless: Playwright ignores the `timeout` option on `locator.isVisible()` (it's a deprecated no-op that returns an immediate snapshot), so `1_000` vs `LONG_WAIT` made no difference here either way.

Fixed in 5b105624: `signOpenedDocument` now waits for the signer form via `locator.waitFor` (whose timeout *is* honored) — its absence means the demo already auto-signed the doc, so we bail and let the caller re-check the Continue gate. I also dropped the ignored `{ timeout }` args from the remaining point-in-time `isVisible` snapshots, which run after a landmark assertion has already gated rendering. Re-ran the demo canary green.

— Claude (auto-generated response)

PR #2359, opened 2026-07-07
feat: add FormBox UI component and adapter slot

aaronlee777:
## Summary

- Adds a new `FormBox` SDK UI primitive alongside `Box`, registered on the component adapter so partners can override the container used for form-section grouping independently of the general-purpose `Box`.
- `FormBox` mirrors `Box` (supports `header`, `withPadding`, `className`) but omits the `footer` prop — form sections don't use a footer slot, so dropping it keeps the partner API surface tighter.
- Wires the new slot through `ComponentsContextType`, the default adapter, the Storybook `PlainComponentAdapter`, and sdk-app's `nativeAdapter`. Includes stories, tests, and an SCSS module.

## Demo

Screenshot 2026-07-08 at 11 43 40 AM
Screenshot 2026-07-08 at 11 43 32 AM
Screenshot 2026-07-08 at 11 43 26 AM

## Test plan

- [x] `npm run test -- --run src/components/Common/UI/FormBox/FormBox.test.tsx` — new tests (incl. axe accessibility checks) pass
- [ ] `npm run storybook` — visit **UI/Components/FormBox** and verify Default, WithHeader, WithCustomHeader, WithHeaderAndContent, WithCustomClassName, FlushContent, and WithEmbeddedTable stories render as expected
- [ ] Confirm `Box` usage is unaffected (rendering identical, no partner-facing API change on `Box`)
- [ ] Run `npm run derive` to regenerate `docs/reference/component-inventory.md` with `FormBoxProps` and the new components-map row


mariechatfield:
@aaronlee777 would you mind adding some screenshots to this PR description? I find them helpful when reviewing as well for our future selves if we ever need a refresher on what we added when and what the original intent was

src/components/Common/UI/FormBox/FormBoxTypes.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+export interface FormBoxProps {
+  /**
+   * Content rendered inside the form box body.
+   */
+  children: ReactNode
+  /**
+   * Optional content rendered above the body in the form box header section.
+   */
+  header?: ReactNode
```
mariechatfield, line 18: All of your examples use `BoxHeader` when building the header, should that be documented as the expected usage? e.g. the header shouldn't wrap any content in its own box header because it should expect that the header prop _is already wrapped_ in the styled component?

src/components/Common/UI/FormBox/FormBoxTypes.ts
```diff
@@ -0,0 +1,27 @@
+import type { ReactNode } from 'react'
+
+/**
+ * Props your `FormBox` implementation must accept from the component adapter.
+ * Renders a sectioned container used to group related form fields, with an optional header slot.
```
mariechatfield, line 5: What is the difference between this and the standard box? in the `Common/UI` implementation, I don't see anything like a [`<fieldset>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/fieldset) or other markup specific to forms

mariechatfield: Ah I see in the description now

> FormBox mirrors Box (supports header, withPadding, className) but omits the footer prop — form sections don't use a footer slot, so dropping it keeps the partner API surface tighter.

We should make sure that this kind of "when to use FormBox vs Box" and the footer omission is clearly outlined in the tsdoc since that's what partners will have access to (in intellisense or in the docs site)

mariechatfield (commented):
Image

nice, looks like a box! and I do see in the stories that the standard box has a footer example, whereas this one doesn't

Image

src/components/Common/UI/FormBox/FormBoxTypes.ts
serikjensen: Yes this was actually a rec from me, basically for adapters i wanted to make sure that a partner had more specific control over boxes in form usages in case they had differing patterns.

src/components/Common/UI/FormBox/FormBoxTypes.ts
serikjensen: I think this is worth discussing, @aaronlee777 are we sure that we want BoxHeader usage for form headers? i think just using header might be sufficient? BoxHeader was an adapter for a very specific use case, so we may not want that to apply to form usages. Ex. you may want a form specific header abstraction

one way to think of all of this is just imagining being a partner trying to change these things universally and then reasoning about what side effects would be unintentional. If a partner was updating the BoxHeader for the box instances, would they also want the forms to change?

aaronlee777: Thank you both. I agree that this component deserves its own specific header component. I've just pushed it.

mariechatfield (approved):
Thanks for taking the time to address the feedback! This looks great, let's ship it 🚀

PR #2364, opened 2026-07-07
chore(SDK-1054): deterministic union sort + @internal typeof const expansion in expandDtsTypeof

mariechatfield:
First of a 4-PR stack splitting the SDK-1054 doc-coverage work for reviewability. Stack: **PR1 (this)** → PR2 (shrink surface) → PR3 (grow surface) → PR4 (enforce).

## What

Two independent enhancements to the `.d.ts` post-processor, both prerequisites for turning API Extractor release-tag enforcement on (PR4):

- **Deterministic union sort** — sort all-literal union members alphabetically at every nesting depth. The checker's `typeToString` emits union order non-deterministically, churning the tracked API report; member order is semantically irrelevant. Sorting runs in an isolated in-memory project so it never perturbs the build's type checker. Keyword unions (`string | null`, `T[] | undefined`) keep their given order.
- **`@internal typeof <const>` expansion** — resolve `typeof` references to `@internal` consts (exported-local and imported) to their concrete types, leaving declarations in place and pruning now-orphaned import specifiers, so `@public` aliases no longer trip `ae-incompatible-release-tags`.

## Scope

Tooling only (`build/expandDtsTypeof.ts` + tests). No change to the published API surface. Safe to land first.

### DIFF
--- a/.reports/embedded-react-sdk.api.md
+++ b/.reports/embedded-react-sdk.api.md
@@ -678,7 +678,7 @@ export const BankFormErrorCodes: {
 };
 
 // @public
-export type BankFormField = "name" | "accountNumber" | "routingNumber" | "accountType";
+export type BankFormField = "accountNumber" | "accountType" | "name" | "routingNumber";
 
 // @public
 export interface BankFormFields {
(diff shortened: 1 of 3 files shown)


.reports/embedded-react-sdk.api.md
```diff
@@ -3674,10 +3674,8 @@ interface OnboardingExecutionFlowProps {
     withEmployeeI9?: boolean;
 }
 
-// Warning: (ae-forgotten-export) The symbol "INITIAL_COMPONENT_MAP" needs to be exported by the entry point index.d.ts
-//
 // @public
-type OnboardingExecutionInitialState = keyof typeof INITIAL_COMPONENT_MAP;
```
mariechatfield, line 3680: Example of translating even more inline constant references. This one was for a constant that _was_ exported but marked as `@internal`.

serikjensen (approved):

PR #2365, opened 2026-07-07
refactor(SDK-1054): shrink the public API surface

mariechatfield:
Second of the SDK-1054 stack. Base is **PR1** (#2364); merge order 1→2→3→4.
Stack: PR1 (tooling) → **PR2 (this: shrink surface)** → PR3 (grow surface) → PR4 (enforce).

## What

Pull internal-only members off the public surface so partners see a clean API and API Extractor has nothing to flag. Behavior-neutral.

- **Flow/payment entry points** — `PayrollExecutionInternalFlow`, `PaymentSummaryInternal`, `PaymentsListInternal` carry flow-injected props (`prefixBreadcrumbs`, alerts) on `@internal` extension types; parent flows render the internal variants.
- **Compound-attachment removal** — `InformationRequestForm.Footer` and `Compensation.JobsList`/`EditCompensation` drop their compound-component attachments.
- **Field element registry** — the internal `_fieldElementRegistry` **key** is kept entirely off `HookFormInternals`/`FormHookResult`. A non-exported `@internal HookFormInternalsWithRegistry` carries it; `useHookFormInternals` builds that shape and the three internal readers (`composeSubmitHandler`, `useHookFieldResolution`, `SDKFormProvider`) cast back to it. No public key, no `_FieldElementRegistry` export — it never reaches the entry point, so no underscore rule applies.
- **Internal-type-ref removal from public props** — `Address` `FallbackComponent` and `GustoProvider` `LoaderComponent` now use explicit public types; `FlowBreadcrumb` retagged `@internal`.
- Drop the `EmployeeStateTaxesMetadataConfig` export.

## Organizing principle

This PR is the "shrink" half of the surface work; new documented exports live in PR3 ("grow"). Existing tests updated to the new entry points.


src/components/Contractor/Payments/PaymentsList/PaymentsList.tsx
```diff
@@ -17,10 +17,16 @@ import { usePagination } from '@/hooks/usePagination/usePagination'
...
+
+/**
+ * Props for the flow-internal {@link PaymentsListInternal}, which layers
+ * flow-injected alerts on top of the public {@link PaymentsListProps}.
+ *
+ * @internal
+ */
+export interface PaymentsListInternalProps extends PaymentsListProps {
+  /** Flow-injected alerts (e.g. wire-transfer confirmation, payment cancellation). */
   alerts?: InternalAlert[]
```
mariechatfield, line 30: Instead of marking individual keys as `internal`, rework the types so the public export doesn't have the key, and our internal usages can use the `Internal` version that has it

src/components/Payroll/PayrollExecutionFlow/PayrollExecutionFlow.tsx
```diff
@@ -53,18 +53,27 @@ export interface PayrollExecutionFlowProps {
...
+ * prefix breadcrumbs on top of the public {@link PayrollExecutionFlowProps}.
+ *
+ * @internal
+ */
+export interface PayrollExecutionInternalFlowProps extends PayrollExecutionFlowProps {
+  /**
+   * Breadcrumbs prepended to the flow's own breadcrumb trail. Set by a parent flow (e.g. an
+   * off-cycle creation step) so the breadcrumb history remains coherent across the handoff.
+   */
+  prefixBreadcrumbs?: FlowBreadcrumb[]
```
mariechatfield, line 74: I don't think we ever intended the flow `prefixBreadcrumbs` to be used by partners, so moved this to an internal interface that can be reused by our other flows that wrap this one

src/partner-hook-utils/form/composeSubmitHandler.ts
```diff
@@ -98,7 +97,11 @@ function focusFirstInvalidAcrossForms(results: FormValidationResult[]): void {
   const candidates: Candidate[] = []
   for (const { input, formMethods, errors } of results) {
     if (isRawUseForm(input)) continue
-    const registry = input.form.hookFormInternals._fieldElementRegistry
+    // `_fieldElementRegistry` rides on the runtime object but is kept off the
+    // public form-hook types (see HookFormInternalsWithRegistry); read it back here.
+    const registry = (
+      input.form.hookFormInternals as { _fieldElementRegistry?: FieldElementRegistry }
+    )._fieldElementRegistry
```
mariechatfield: This is the one bit of trickery where we cast back to saying the `_fieldElementRegistry` must exist. Considering that we set it up in the first place, I'm not especially worried about it

src/partner-hook-utils/form/composeSubmitHandler.ts
```diff
@@ -98,7 +99,8 @@ function focusFirstInvalidAcrossForms(results: FormValidationResult[]): void {
   const candidates: Candidate[] = []
   for (const { input, formMethods, errors } of results) {
     if (isRawUseForm(input)) continue
-    const registry = input.form.hookFormInternals._fieldElementRegistry
+    const registry = input.form.hookFormInternals._fieldElementRegistry as
```
mariechatfield, line 102: Update the `_fieldElementRegistry` _public type_ as `unknown` so we don't have to export the details of the `FieldElementRegistry` type. When we're ready to use it internally, we cast back to the concrete type

serikjensen (approved):

PR #2370, opened 2026-07-08
chore(docs-site): emit the standard PageViewed analytics event

mariechatfield:
## What

Records docs-site page views under the current `PageViewed` analytics event so they land in analytics as `PageViewed` rather than the legacy page-view event.

- Bump the analytics CDN bundle `15.11.4 → 15.11.6` (latest available).
- Opt the docs site into server-side provider delivery for page views (`serverSideEligibleProviders: ['amplitude']`), set before the bundle loads.
- Emit the standard `Page`/`Viewed` event via `track` instead of the legacy auto page-view call.
- Introduce a small `AnalyticsEvent` type; the event owns its identity (`eventCategory`/`eventName`/`data`) and the emitter reads it off the event, so adding future events is a one-line union.

## Why

The legacy page-view path is recorded under the old event name. Emitting the standard `Page`/`Viewed` event records page views under the current `PageViewed` name.

## Notes

A server-side allowlist entry is also required for these events to be delivered end-to-end; that change lives outside this repo and is tracked separately.

## Testing

- `tsc --noEmit` clean for the changed files.
- Verify after deploy: page views emit `Page`/`Viewed` with a non-empty `server_side_eligible_providers`, and appear downstream as `PageViewed`.

The page view event includes this context (decoded):

```json
{
    "referrer": "http://localhost:3000/docs/getting-started/proxy-security-partner-guidance?cache_reset=1783539989654",
    "path": "/",
    "url": "http://localhost:3000/",
    "title": "Gusto Embedded"
}
```

### DIFF
--- a/docs-site/src/analytics/PageViewedEvent.ts
+++ b/docs-site/src/analytics/PageViewedEvent.ts
@@ -1,14 +1,9 @@
 import type { AnalyticsEvent } from '@site/types'
 
 /**
- * Tracks a page view. `name` should be the page title.
+ * Tracks a page view.
  */
-export class PageViewedEvent implements AnalyticsEvent<{ name: string }> {
+export class PageViewedEvent implements AnalyticsEvent {
   readonly eventCategory = 'Page'
   readonly eventName = 'Viewed'
-  readonly data: { name: string }
-
-  constructor(name: string) {
-    this.data = { name }
-  }
 }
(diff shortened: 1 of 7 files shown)


serikjensen (approved):

mariechatfield:
heads up: just changed our tracker namespace from `gusto_sdk_docs` to `embedded-sdk-docs`. When making the backend config changes I realized all the other apps use kebab-case instead of snake_case and none of them prefix with gusto anyway; embedded is more specific to the sdk.

Since none of these docs have been making it to amplitude anyway, I think a breaking change right now is fine

PR #2380, opened 2026-07-10
ci: add Lokalise translation workflows for i18n strings

dmortal:
## Summary

Adds GitHub Actions workflows to route this SDK's `src/i18n/en/*.json` strings through Gusto Embedded's shared Lokalise translation pipeline, replicating the pattern already used by gws-flows and zenpayroll via the reusable `Gusto/embedded-i18n-actions` actions.

## Changes

- `push-sources.yaml` — pushes `src/i18n/en/*.json` to Lokalise on `main` pushes touching those files, or manual dispatch
- `pull-translations.yaml` — manually pulls reviewed `es_US` translations from Lokalise and opens a PR (`file_format: json`, since the action defaults to Rails-style `ruby_yml`)
- `check-manual-translations.yaml` — blocks PRs that hand-edit `src/i18n/es_US/*.json` outside the Lokalise pull-PR flow

## Related

- Reference: [Gusto/embedded-i18n-actions](https://github.com/Gusto/embedded-i18n-actions)

## Testing
TBD

### DIFF
--- a/.github/workflows/check-manual-translations.yaml
+++ b/.github/workflows/check-manual-translations.yaml
@@ -1,0 +1,19 @@
+name: 'Check for manual translations'
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled]
+    paths:
+      - "src/i18n/es_US/**.json"
+jobs:
+  check-for-manual-translations:
+    runs-on:
+      group: gusto-ubuntu-default
+    permissions:
+      contents: read
+      pull-requests: write
+    steps:
+      - name: Run manual translations check
+        uses: Gusto/embedded-i18n-actions/manual_translations_check@main
+        with:
+          file_patterns: |
+            **/i18n/es_US/**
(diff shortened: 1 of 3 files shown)


.github/workflows/pull-translations.yaml
```diff
@@ -0,0 +1,18 @@
+name: Pull Translations
+on:
+  workflow_dispatch:
```
mariechatfield, line 3: So this one only runs when we trigger it manually right?

.github/workflows/check-manual-translations.yaml
mariechatfield, line 1: Can you add a temporary commit that would cause this job to fail the PR just to verify it runs when expected? Otherwise this LGTM!

krisxcrash (approved):

.github/workflows/pull-translations.yaml
dmortal: for now, yes, will automate once we have it working end-to-end

PR #2392, opened 2026-07-13
feat: enable W-9 download on Contractor Submit

aaronlee777:
## Summary

- **Personalized copy** — `documentRequirements.description` now addresses the contractor by name: individual contractors get their `firstName`, business contractors get their `businessName`.
- **Better BoxHeader instructions** — reworked the "collect signed copies" line so it reads naturally with the interpolated name.
- **Working download button** — the "Download document" button is wired to the real PDF URL via `useContractorDocumentsGetPdf`, mirroring the pattern used in the contractor `SignatureForm`. Items in `documentRequirements.items` now carry a `documentName` key that maps to the API document (`taxpayer_identification_form_w_9` for W-9), and the button hides when no matching document or URL is available.
- **New `Skeleton` primitive + adapter** — added `src/components/Common/UI/Skeleton/` with `width`/`height` props, registered it in the `ComponentsContext` type, `defaultComponentAdapter`, and the Storybook `PlainComponentAdapter`. The Submit screen renders it as the button label while the PDF URL is loading, so the button doesn't pop in after the rest of the surface paints.

## Demo

Screenshot 2026-07-13 at 11 48 45 AM

## Test plan

- [ ] Storybook: open the `Contractor.Submit` story with a mock contractor whose documents include a W-9 template. Confirm the button shows the shimmering skeleton briefly, then renders "Download document" and opens the PDF in a new tab on click.
- [ ] Storybook: override the mock so no W-9 document exists (or the PDF URL is missing). Confirm the button never appears and the row still renders title + description cleanly.
- [ ] Individual contractor: description reads "…collect signed copies of the following documents from `<firstName>`."
- [ ] Business contractor: description reads "…collect signed copies of the following documents from `<businessName>`."
- [ ] `npm run test -- --run` passes.


src/components/Common/UI/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,27 @@
...
+      {...otherProps}
+      className={classnames(styles.skeleton, className)}
+      style={{ width: toDimension(width), height: toDimension(height) }}
+      role="status"
+      aria-label={otherProps['aria-label'] || 'Loading'}
+      aria-busy
+      aria-live="polite"
+    />
+  )
+}
```
serikjensen: I'm confused on this one and i think we need to get organized on our loading indicators prior to introducing this.

Currently we have `LoaderComponent` which is set via GustoProvider. This is the fallback skeleton that renders with each SDK component currently. It is defined in `src/components/common/Loading`

We also have a `LoadingSpinner` which seems to be used for situations like payroll calculation, or for information request footer loading state.

Now we are introducing an additional loading indictor via skeleton which shares a lot of UI with the existing loader component.

My immediate recommendation for this PR would be to just use the LoadingSpinner from component context for the button instead of defining this component. (Which honestly, shouldn't you get that for free by just setting the Button to isLoading true?)

As a follow up effort outside of this PR, I think we need to inventory and clearly define our loading indicators. This would include
* A design effort to determine exactly which loading indicators to support
* An engineering update where we migrate to these loading indicators
* A code update to remove the LoaderComponent from the GustoProvider and just run all loading indicators through the component adapter.

I don't want this work to get blocked by that effort, so my recommendation for this PR would be to remove the Skeleton so we can work on SDK loading states holistically and use the default button isLoading prop for now

src/components/Contractor/Submit/Submit.tsx
```diff
@@ -110,17 +123,23 @@ const Root = ({ contractorId, selfOnboarding, dictionary }: ContractorSubmitProp
...
+            return (
+              <DocumentRequirementItem
+                key={item.documentName}
+                title={item.title}
+                description={item.description}
+                document={document}
+                downloadLabel={t('documentRequirements.downloadCta')}
+              />
+            )
+          })}
```
serikjensen: This was likely an oversight from my initial implementation, but I don't think it's great that the translations are driving the items rendered here. Basically the API should be the source of truth for determining what is rendered, and then the translations should be creating the copy for the rendered items

I would prompt claude as follows

```
This is brittle as implemented because it is duplicating the source of truth between the api response and the hard coded translations array. Ideally we would iterate over the documents and then have translations available. You can see that pattern implemented elsewhere for rendered items in the SDK where we will get a key from the api response like entity type and then use those as translation keys. I think optimally that is the better approach for dealing with rendered data like this.

Without knowing exactly the shape of the documents returned from the API, we need to add some more safety here

We should create a simple DocuemntRequirementItem that would work with the returned api shape. Ex. things like document name. This one would be very minimal and would just render the docs we got back from the API with non translated data that could look at least presentable and display to the user.

We should then have a branch to check on if the document type is W9. If it is W9, we should have specific W9 translations set that we can reference.

In other words, the end result here would be looping over the direct documents response from the API, rendering the minimal amount needed for non w9 documents (ex. name) and then for W9 do the rich presentation we are seeing with specific w9 translation keys. It makes no sense to iterate over the translations and then reverse index back into the documents returned from the API. This update should make the API the source of truth
```

src/components/Contractor/Submit/Submit.tsx
```diff
@@ -138,6 +157,50 @@ const Root = ({ contractorId, selfOnboarding, dictionary }: ContractorSubmitProp
...
+          onClick={() => {
+            if (pdfUrl) window.open(pdfUrl, '_blank', 'noopener,noreferrer')
+          }}
+        >
+          {isLoading ? (
+            <Skeleton width={120} height={16} aria-label={downloadLabel} />
+          ) : (
+            downloadLabel
+          )}
+        </Button>
```
serikjensen: See feedback above, my rec on this PR would be to just use the isLoading prop on button for now and do a separate effort to align our loading states outside of this PR

serikjensen, line 206: I don't think it's great that we're showing a disabled document when pdf url is not defined. My rec would be to just not render the button at all if there's no pdf.

src/components/Contractor/Submit/Submit.tsx
```diff
@@ -138,6 +157,50 @@ const Root = ({ contractorId, selfOnboarding, dictionary }: ContractorSubmitProp
...
+      <Flex flexDirection="column" gap={4}>
+        <Text weight="medium">{title}</Text>
+        <Text variant="supporting">{description}</Text>
+      </Flex>
+      {(isLoading || pdfUrl) && (
+        <Button
+          variant="secondary"
+          isDisabled={isLoading || !pdfUrl}
+          onClick={() => {
+            if (pdfUrl) window.open(pdfUrl, '_blank', 'noopener,noreferrer')
```
serikjensen, line 202: Optimally this goes away if we avoid rendering the button altogether if there is no PDF

serikjensen (changes requested):
Thanks for getting this one started! biggest required changes from my end

- [x] Skeleton adds a lot of confusion given the loading indicators we already have in the SDK. Recommend to remove that from this PR and just use the Button isLoading for now. We should do a more holistic loading state inventory/revamp in the SDK as a follow up effort to keep this unblocked
- [x] Recommendation to not drive the rendered documents from the translations array. I left a prompt you can give to claude to adjust that

src/components/Contractor/Submit/Submit.tsx
aaronlee777: @serikjensen the isDisabled is gone now. I left the check for pdfUrl as part of the onClick even though the condition for rendering the button checks for it as well

serikjensen (approved):
Looks good! def be sure to manually verify prior to merge

PR #2394, opened 2026-07-13
feat: warn admins when editing contractor profile/address after W-9 signing

aaronlee777:
## Summary

- Adds an SDK `Alert` (warning variant) above the **Contractor Profile** (admin edit surface) and **Contractor Address** forms when the contractor already has a signed W-9 on file.
- Alert copy signals to the admin that changing name/SSN/EIN or the mailing address will require re-collecting a new W-9.
- Introduces a small `useContractorHasSignedW9(contractorId)` Suspense hook that reads the documents endpoint and returns `true` iff the contractor has a W-9 with a `signedAt` timestamp. The presence of a signed W-9 naturally covers both the `self_onboarding_review` and `onboarding_completed` cases without threading state through the flow machine.
- Profile: alert only mounts in edit mode (`mode === 'update'` + resolved `uuid`), so create-mode never fetches documents.
- Address: `contractorId` is always required, so no gating needed.
- Self-onboarding contractor branch (`SelfOnboardingContractorProfile`) is intentionally left untouched.

## Test plan

- [x] `npm run test -- --run src/components/Contractor/Profile src/components/Contractor/Address` (111 passing, including 5 new W-9 warning cases)
- [x] Manual admin review path: pull SDK dev app, load a contractor in `self_onboarding_review` state with a signed W-9, verify warning renders above both edit surfaces

JIRA: SDK-1029


serikjensen (changes requested):
Code is looking good! blocking since we still need confirmation from legal that this is the correct way forward

serikjensen (approved):

serikjensen:
@cursoragent could you take a look at test failures on this one and resolve?

PR #2413, opened 2026-07-15
ci: fix patch snapshot refresh to apply structural changes

mariechatfield:
## Summary

- The patch refresh logic only updated files already in the versioned snapshot, so renames and additions in a patch release were silently skipped. Updated content linking to new paths would break the downstream Docusaurus build.
- Refresh now also copies files present in live `docs/` but absent from the snapshot. Orphans (in snapshot but not in live docs) are left as-is so existing external links keep resolving.
- Extracted all snapshot decision + file logic into `.github/scripts/sync-snapshot.sh` so it can be run and tested locally: `.github/scripts/sync-snapshot.sh --downstream ../embedded-sdk-docs --trigger workflow_run --dry-run`
- Prints a warning when structural changes occur (additions or orphans) so someone can verify the snapshot looks right.

## Test plan

- [x] Dry-run locally against embedded-sdk-docs: `--trigger workflow_run --dry-run` shows the 17 expected adds and 2 orphans for version-0.52
- [ ] Merge and trigger workflow — embedded-sdk-docs build should pass with version-0.52 snapshot complete

### DIFF
--- a/.github/scripts/sync-snapshot.sh
+++ b/.github/scripts/sync-snapshot.sh
@@ -1,0 +1,162 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Decides whether to create or refresh the versioned docs snapshot in the
+# downstream embedded-sdk-docs repo, then performs the file operations.
+#
+# Extracted from publish-docs.yaml so the logic can be exercised locally
+# without triggering a full CI sync. In CI the workflow handles cloning,
+# wiping, copying fresh source, staging, and pushing; this script only owns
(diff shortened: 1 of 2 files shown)


.github/scripts/sync-snapshot.sh
mariechatfield, line 1: Most of this logic is extracted from `publish-docs.yaml`, so that I could run this locally and manually reason about what will happen

krisxcrash (approved):

serikjensen (approved):

PR #2415, opened 2026-07-15
fix(forms): clear API-derived field errors on user input

serikjensen:
## Summary

Field-level errors synced from API 422 responses previously persisted on the field until the next submit, even after the user corrected the input — forcing a resubmit just to see whether the fix was accepted. `SDKFormProvider` now clears a field's server-derived error the moment its value changes, so every form built on the provider gets this behavior automatically.

- **`useSyncFieldErrors`** applies API field errors, then subscribes via react-hook-form's imperative `watch(callback)` to clear each field's error on the first change to that field. The subscription does **not** trigger re-renders; clearing is scoped to the one field (via a closure `Set`) and hands the field back to RHF's own validation afterward.
- **`useApplicableFieldErrors`** is a render-phase stabilizer: it filters/normalizes the errors and keys a referentially-stable array on a primitive content signature (sorted `JSON.stringify`). This prevents the fresh-array-every-render churn (`collectErrors` + `flatMap`) from re-applying an error the user just cleared on an unrelated re-render.
- The top-level error alert is intentionally left untouched — only inline field errors clear.

Scope is deliberately limited to the SDK form infrastructure plus its tests; no consumer/prototype churn.

## Test plan

- [x] `npm run test -- --run src/partner-hook-utils/form/SDKFormProvider.test.tsx` — 4 new tests: errors apply to fields, changed field clears while banner persists, only the changed field clears, and a cleared error is not resurrected by an unrelated re-render.
- [x] Regression suites green: `src/components/Company/PaySchedule/`, `src/components/Employee/Profile/`, `src/partner-hook-utils/form/` (244 tests).
- [x] No new type or lint errors in the touched files.

### Pay Schedule

https://github.com/user-attachments/assets/8a066956-079f-4761-9435-d8edd9c5af47

### Employee onboarding

https://github.com/user-attachments/assets/1f306567-6a68-4df3-860b-4dd72811af29

Also smoke tested all forms in contractor and employee onboarding to verify no regressions

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1159,6 +1159,37 @@
         "url": "https://ko-fi.com/dangreen"
       }
     },
+    "node_modules/@commitlint/read/node_modules/conventional-commits-filter": {
+      "version": "6.0.1",
+      "resolved": "https://registry.npmjs.org/conventional-commits-filter/-/conventional-commits-filter-6.0.1.tgz",
+      "integrity": "sha512-cs+LadpH7Kpw0M3k8wurk+sOVVDAENA0iK4OBOrkL94j5lEVYRJ4j3zd2bhY9qgzyrPqthdcYT3axzRN7AliMg==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "peer": true,
+      "engines": {
+        "node": ">=22"
+      }
+    },
+    "node_modules/@commitlint/read/node_modules/conventional-commits-parser": {
+      "version": "7.1.0",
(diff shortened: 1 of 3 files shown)


serikjensen:
@cursoragent looks like some prettier checks to fix here? could you take care of those?

krisxcrash (approved):

dmortal (approved):

serikjensen:
@cursoragent could you run a fresh npm install on the latest and include the lockfile fix

krisxcrash (approved):

PR #2420, opened 2026-07-16
feat: inline bank form UX for PaymentMethod onboarding

aaronlee777:
## Summary

- **FormBox footer slot**: adds `footer` prop to `FormBox` (mirrors `Box`), enabling actions below flush content
- **Inline initial bank form**: selecting Direct Deposit immediately reveals bank account fields inline — no intermediate "Add bank account" button required
- **Inline add-another form**: clicking "+ Add another bank account" (in the FormBox footer) reveals the form below the DataView in the same view rather than navigating away; Cancel dismisses it, Save creates the account
- **Check hides all deposit UI**: switching to Check hides the DataView, Add another, and Split paycheck
- **Header and title**: wraps the view in `FormBox` with header "Payment method" (`withPadding={false}`)
- **BankFormBody `onCancel` optional**: Cancel button only renders when the prop is provided (supports inline forms with no cancel action)

## Test plan

- [ ] Select Direct Deposit with no accounts — bank form fields appear immediately, no Cancel button, Continue hidden
- [ ] Fill in bank form and Save — DataView appears with the new account, Continue visible, footer shows Add another
- [ ] Click Add another — inline form appears below DataView with top border, DataView still visible, Cancel and Save present
- [ ] Cancel inline form — form hides, footer restores
- [ ] Select Check — DataView, Add another, and Split paycheck all hide; Continue appears
- [ ] 2+ accounts — Split paycheck appears in footer alongside Add another
- [ ] `FormBox` `WithFooter` story renders correctly in Storybook


serikjensen (changes requested):
I'm getting turned around a bit with these changes. I think we may need to do more to integrate the bank form with the state machine to keep separation of concerns here

let's set up some time to talk about this one in more detail. I think we need to better outline exactly what the state transitions should be and then we can design a state machine to properly account for the behavior

serikjensen (approved):

serikjensen (approved):

PR #2434, opened 2026-07-20
refactor(SDK-1124): extract usePaymentAmountsEditor from CreatePayment

mariechatfield:
## Summary

Extracts the amounts-editing slice of `CreatePayment` into a new internal shared hook `usePaymentAmountsEditor` at `Contractor/Payments/shared/`. `CreatePayment`'s public API and all six events are unchanged.

## Changes

- **New hook** `usePaymentAmountsEditor` owns: per-contractor payment state (`virtualContractorPayments`), totals calculation, and the edit-modal state machine, returned as a single `editModal: { isOpen, formMethods, open, close, submit }` object. It's parameterized by `allowedPaymentMethods` so `CreatePayment` passes `['Check', 'Direct Deposit']` and a future caller (e.g. historical payments) can restrict it further.
- **`CreatePayment` Root** is reduced to the preview/submission flow, payment date, and alerts — the amounts-editing slice is replaced by a single `usePaymentAmountsEditor(...)` call. Alert construction and event firing happen via `onEditOpen`/`onEditSave` callbacks so the hook itself stays free of i18n/event coupling.
- **`SUPPORTED_PAYMENT_METHODS`/`allowedPaymentMethods` widened** to include `'Historical Payment'`, in prep for the historical-payments flow reusing this hook. `allowedPaymentMethods` is now typed off the SDK's generated `PostV1CompaniesCompanyIdContractorPaymentGroupsPaymentMethod` enum instead of a hand-rolled literal union, which also removes the `as 'Check' | 'Direct Deposit' | 'Historical Payment'` casts that were needed before.
- **Grouped the edit-modal return values** under a single `editModal` object (`isOpen`/`formMethods`/`open`/`close`/`submit`) instead of five flat sibling properties with inconsistent verb naming (`isModalOpen`, `onCloseModal`, `onEditContractor`, `onEditContractorSubmit`). Matches the shape sketched in the Historical Payments tech spec. No behavior change.
- **Pre-migration tests** (commit de2c5e28) cover the extracted behaviors: `isTouched` flag, totals (hourly vs. fixed), edit modal initial values, `allowedPaymentMethods` gating, continue-to-preview filtering, and all six events. These were nearly all untested before this PR.

## Demo

Pure refactor, no changes.

https://github.com/user-attachments/assets/1d2cb5d8-3bd4-498c-90ce-fae6d96efe27

## Related

- Jira: [SDK-1124](https://gustohq.atlassian.net/browse/SDK-1124)

## Testing

- `npm run test -- --run src/components/Contractor/Payments/`
- Review commit-by-commit: de2c5e28 (tests only, no logic) → 4872ece6 (the extraction) → afd2d0be (formatting) → 768c9281 (presentation-concerns cleanup + Historical Payment prep) → 925e5596 (regression fixes + cast cleanup) → 6820a049 (editModal grouping). Starting with the tests gives you the behavioral contract before reading the implementation move.

[SDK-1124]: https://gustohq.atlassian.net/browse/SDK-1124


src/components/Contractor/Payments/CreatePayment/CreatePayment.tsx
```diff
@@ -145,62 +137,46 @@ const Root = ({ companyId, dictionary, onEvent }: CreatePaymentProps) => {
...
+  const { virtualContractorPayments, totals, editModal } = usePaymentAmountsEditor({
+    contractors,
+    allowedPaymentMethods: ['Check', 'Direct Deposit'],
+    onEditOpen: () => {
+      clearAlerts()
+      onEvent(componentEvents.CONTRACTOR_PAYMENT_EDIT)
+    },
+    onEditSave: data => {
+      const displayContractor = contractors.find(c => c.uuid === data.contractorUuid)
+      const displayName = DOMPurify.sanitize(
```
dmortal, line 157: Not sure why this is here - the name does not come from the user input here, right? it's from the API response?

dmortal (approved):

src/components/Contractor/Payments/CreatePayment/CreatePayment.tsx
mariechatfield: I can't say I understand it either, but this is just copied from the file's previous version and moved around (you can see this same logic on line 320 in the side-by-side diff).

If I had to guess, this is a bit of over-the-top extra security that's not needed, but I didn't want to remove it without being certain

serikjensen (approved):

PR #2441, opened 2026-07-21
fix(payroll): apply and surface default date range in payroll filters

aaronlee777:
## Summary

- Apply a default date range in `PayrollList` (today → 3 months out) and `PayrollHistory` (3 months back → today), routed through the `useDateRangeFilter` hook so both API queries and the filter UI show the same range on load.
- Display the effective default range in the `DateRangeFilter` trigger and popover instead of the "Select dates" placeholder, so users see what they're seeing before opening the filter.
- Differentiate the empty state in both lists: when the user has actively filtered by date, show "No payrolls in this date range / Try adjusting the date filter to see more payrolls." Otherwise keep the original "no payrolls yet" copy.
- All date serialization goes through the timezone-safe `formatDateToStringDate` / `normalizeToDate` helpers per the SDK date-handling contract.

## Test plan

- [ ] Storybook: `Payroll/PayrollList` and `Payroll/PayrollHistory` — verify the default range appears in the filter trigger on load and that the popover opens pre-populated with the same range.
- [ ] Storybook: verify the empty-state variant renders both "no payrolls" (default range) and "no payrolls in this date range" (after picking a custom range with no matches).
- [ ] Confirm no regressions in existing PayrollList / PayrollHistory tests and stories.

### DIFF
--- a/src/components/Common/DateRangeFilter/DateRangeFilter.tsx
+++ b/src/components/Common/DateRangeFilter/DateRangeFilter.tsx
@@ -23,7 +23,6 @@ interface DateRangeFilterProps {
   resetLabel: string
   selectDatesLabel: string
   triggerLabel: string
-  isFilterActive: boolean
   maxEndDate?: Date
   minStartDate?: Date
 }
@@ -42,7 +41,6 @@ export const DateRangeFilter = ({
   resetLabel,
   selectDatesLabel,
   triggerLabel,
-  isFilterActive,
   maxEndDate,
   minStartDate,
 }: DateRangeFilterProps) => {
--- a/src/components/Common/DateRangeFilter/DateRangeFilter.tsx
+++ b/src/components/Common/DateRangeFilter/DateRangeFilter.tsx
@@ -82,9 +82,9 @@ export const DateRangeFilter = ({
   }, [])
 
   const filterDateLabel = useMemo(() => {
-    if (!isFilterActive || !startDate || !endDate) return null
+    if (!startDate || !endDate) return null
     return `${formatFilterDate(startDate)} – ${formatFilterDate(endDate)}`
-  }, [isFilterActive, startDate, endDate])
+  }, [startDate, endDate])
 
   const triggerButton = (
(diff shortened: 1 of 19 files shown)


src/components/Common/DateRangeFilter/DateRangeFilter.tsx
```diff
@@ -82,9 +82,9 @@ export const DateRangeFilter = ({
   }, [])
 
   const filterDateLabel = useMemo(() => {
-    if (!isFilterActive || !startDate || !endDate) return null
```
serikjensen, line 85: I think we may need to do some additional work here. I'm worried that the isFilterActive prop is getting out of sync with the start/end dates being supplied. Basically i'm not sure what the source of truth is for the state on this component. I think i'd feel better about it if startDate/endDate were driving the component behavior rather than the isFilterActive. Claude may be able to respond to this feedback

src/components/Payroll/PayrollHistory/PayrollHistory.tsx
```diff
@@ -65,6 +75,8 @@ const Root = ({ onEvent, companyId, dictionary }: PayrollHistoryProps) => {
   const { currentPage, itemsPerPage, getPaginationProps, resetPage } = usePagination()
 
   const dateRangeFilter = useDateRangeFilter({
+    defaultStartDate: getDefaultStartDate(),
+    defaultEndDate: getDefaultEndDate(),
```
serikjensen: Yeah to my point above there are some react risks taking place here where we now have a default start date/end date along with start date/end date. We should have start date and end date be a single source of truth for the component rather than reconciling with the default. This is likely because the current date range filter is trying to be uncontrolled when we need it to be controlled only. I would recommend passing this to claude to see if it can rectify. but you also may consider just writing the acceptance criteria into the ticket

serikjensen (changes requested):

serikjensen (approved):

serikjensen (approved):

PR #2455, opened 2026-07-22
feat(employee): rehire form with active/dismissed status badges

serikjensen:
## Summary

Implements the rehire flow for terminated employees (SDK-1147) plus the supporting status badges and a tab-classification fix.

- **RehireEmployee** form (container + presentation) wired to `useEmployeeEmploymentsCreateRehireMutation`, added as a `rehire` state in the `EmployeeListFlow` state machine and exported from the `EmployeeManagement` namespace.
- **Dismissed tab actions**: surface **Rehire** and **Edit** on dismissed employees' menu.
- **Status badges**:
  - `RehireStatusBadge` — badge on the Dismissed tab for employees with an upcoming rehire.
  - `LastDayStatusBadge` — badge on the Active tab for employees with a scheduled (today-or-future) termination.
  - Shared badge date formatting (`badgeDate.ts`, conditional-year) between both.
- **Tab filter fix**: the Dismissed tab now queries `terminatedToday` instead of `terminated`. `terminated` includes *scheduled-future* terminations, which wrongly pulled soon-to-be-dismissed employees into Dismissed; `terminatedToday` keeps them on Active (matching the gws-flows reference implementation).

## Testing

- Unit tests for `RehireEmployee` (submit/cancel/validation), `RehireStatusBadge`, `LastDayStatusBadge`, and extended `EmployeeListFlow` + `useEmployeeList` coverage.
- `npm run test` green for touched suites (35 tests); `tsc --noEmit` clean.

## Notes

- Last-day badge uses `status="warning"` to distinguish an upcoming departure from the `info` rehire badge — easy to change if a neutral pill is preferred.
- Not yet done (flagged for follow-up): the upcoming-termination badge on the **Onboarding** tab (flows renders it there too).


serikjensen:
## Open question: intermittent rehire 422 (`employment_status` inclusion) — backend-side, unresolved

Flagging an investigation we couldn't fully close, so it isn't lost. It is **not** believed to be a defect in this PR's code.

### Symptom
Rehiring some employees returns a 422 from ZenPayroll:

```
Employment status's value is not included in the list
```

…while other employees rehire fine — with the **identical request payload**.

### What we confirmed
- Our rehire body **omits** `employmentStatus` on purpose. This matches the gws-flows reference implementation, which also omits `employment_status`.
- Both the SDK and gws-flows pin API version **2026-06-15**.
- The error is a backend **inclusion validation** on `employment_status`, i.e. it's validating the employee's *existing/stored* employment record, not the (absent) field in our request.
- The SDK's request is structurally correct — it **succeeds (201)** against a local ZenPayroll with the same omitted-field payload.

### Environment / employee matrix (same payload, different outcomes)
| Consumer | Backend | Employee | Result |
|---|---|---|---|
| gws-flows UI | demo env | its demo employee | ✅ 201 |
| SDK app | demo env | demo-company employees | ❌ 422 |
| SDK app → local gws-flows | local dev ZenPayroll | Todd Packer | ✅ 201 |

### Where it's murky
We could not isolate *why* certain employees fail. Leading theory: the employee's **stored** `employment_status` is a legacy / `not_set` value that the rehire path re-validates against a stricter enum. Alternatives not ruled out: demo-data seeding differences, or version/behavior drift between the demo ZenPayroll and local dev ZenPayroll. Tracing the exact ZenPayroll validation was interrupted, so this stays open.

### Suggested next step
Trace the ZenPayroll inclusion validation on `employment_status` in the rehire path and compare the stored value for a failing demo employee vs. a passing one (local Todd Packer). If ZP requires a valid `employment_status` on rehire for certain records, we may need to collect/send it rather than relying on the backend default — but that should be confirmed against ZP behavior, not assumed.

serikjensen:
Closing in favor of #2478 while the rehire form and status badges await prioritization. Keeping this draft as the reference for future spike work.\n\n*Posted by Codex on behalf of Steve*

PR #2461, opened 2026-07-23
feat(SDK-1125): add useSelectContractors hook

mariechatfield:
## Summary

Adds an internal `useSelectContractors` hook that fetches every contractor page up front, filters to active + onboarded contractors, and owns client-side pagination, search, and cross-page selection state. Not exported from the SDK barrel — first piece of the SDK-1126 `SelectContractors` block.

## Changes

- New `useSelectContractors` hook: parallel per-page fetch (non-suspense), active+onboarded eligibility filter, `useClientPagination` composition
- Exposes `actions.onSelect` / `onSelectAll` / `onSearchChange` / `onSearchClear`, matching the `actions: { onX }` convention used by sibling list hooks (`useContractorList`, `useEmployeeList`)
- `selectedIds` is derived each render to exclude uuids for contractors that become ineligible or drop off the list
- Unit + behavioral tests: eligibility filter, multi-page fetch/merge, client pagination, and cross-page select-all scoping (with and without an active search)

## Demo

_None — internal hook only, no UI._

## Related

- Jira: [SDK-1125](https://gustohq.atlassian.net/browse/SDK-1125)

## Testing

- `npm run test -- --run src/components/Contractor/shared/SelectContractors/` — 17/17 passing
- `npm run test -- --run` — full suite, 293 files / 3501 tests passing
- `npx tsc --noEmit -p .` — clean
- `npx eslint src/components/Contractor/shared/SelectContractors/` — clean

[SDK-1125]: https://gustohq.atlassian.net/browse/SDK-1125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
```diff
@@ -0,0 +1,194 @@
...
+    contractor.isActive &&
+    contractor.onboardingStatus === ContractorOnboardingStatus.ONBOARDING_COMPLETED
+  )
+}
+
+function matchesContractorSearch(contractor: Contractor, query: string): boolean {
+  const name =
+    contractor.businessName ?? `${contractor.firstName ?? ''} ${contractor.lastName ?? ''}`
+  return name.toLowerCase().includes(query.toLowerCase())
+}
```
mariechatfield, line 28: We should normalize query once instead of re-normalizing it for each possible match

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
```diff
@@ -0,0 +1,194 @@
...
+   * Unaffected by the current search — see `pagination.totalCount` for the
+   * eligible-and-search-matching count instead.
+   */
+  eligibleCount: number
+  /**
+   * Currently selected contractor uuids. Always a subset of the current
+   * eligible contractors: uuids for contractors that become ineligible or
+   * disappear from the list on a refetch are dropped automatically.
+   */
+  selectedIds: Set<string>
```
mariechatfield, line 50: Should these be in `data`? Or is that really only for data loaded from the server

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
```diff
@@ -0,0 +1,194 @@
...
+  const restPageResults = useQueries({
+    queries: Array.from({ length: Math.max(0, totalServerPages - 1) }, (_, i) => ({
+      ...buildContractorsListQuery(gustoClient, {
+        companyUuid: companyId,
+        page: i + 2,
+        per: SERVER_MAX_PER_PAGE,
+      }),
+      enabled: !!firstPage,
+    })),
+  })
```
mariechatfield, line 111: Is any of this worth sharing with the `useSelectEmployeeData` hook that we adapted or would that be an abstraction too early?

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
```diff
@@ -0,0 +1,194 @@
...
+  const errorHandling = composeErrorHandler([firstPageQuery, ...restPageResults])
+
+  const eligibleContractors = useMemo<Contractor[]>(() => {
+    if (!firstPage) return []
+    const contractors = [
+      ...(firstPage.contractors ?? []),
+      ...restPageResults.flatMap(r => r.data?.contractors ?? []),
+    ]
+    return contractors.filter(isEligibleContractor)
+  }, [firstPage, restPageResults])
```
mariechatfield, line 125: This feels like the wrong way to create this memoized data. Why not save `eligibleContractors` as regular state, and manually update it whenever we get new results? We're in control of whenever those results changes; observer-based patterns are easy to mess up.

I think the counter-argument is that derived state should always be derived not persisted into its own state, but I'm not sure that this is a complex enough use case that "make sure we always update the eligible list whenever we get back new contractors" is harder than "make sure the memoization helper runs exactly as often as it should and we're not accidentally calling it too frequently or too infrequently based on something subtle like a function that always returns a new array object identity"

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
```diff
@@ -0,0 +1,194 @@
...
+    searchPredicate: matchesContractorSearch,
+    defaultItemsPerPage: 25,
+  })
+
+  const isFetching = isFirstPageFetching || restPageResults.some(r => r.isFetching)
+
+  const selectedIds = useMemo(() => {
+    const validUuids = new Set(eligibleContractors.map(contractor => contractor.uuid))
+    return new Set([...rawSelectedIds].filter(uuid => validUuids.has(uuid)))
+  }, [rawSelectedIds, eligibleContractors])
```
mariechatfield: Why memoize this too? It is so easy to mess up logic with memoizing early, what is this to guard against?

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
mariechatfield: Disagree — this matches existing convention. `useContractorList`'s `UseContractorListReady` puts `pagination` alongside `data` rather than inside it, and derived/computed fields (`eligibleCount`, `selectedIds`) follow the same pattern here — `data` holds the primary entity payload, everything computed sits as a sibling. Keeping as-is.

— Claude, reviewing on Marie's behalf

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
mariechatfield: Agree this is worth pursuing, but as a follow-up rather than in this PR. `useSelectEmployeesData` is a near-structural clone: same first-page-then-fan-out pagination pattern, the same connection-limit comment verbatim, and the same select-all-scoped-to-search-not-visible-slice logic. Not extracting here because the Suspense vs. non-Suspense split (this hook needs `HookLoadingResult`, that one uses Suspense) means the query fan-out itself probably won't extract cleanly — but the select-all/selection-pruning logic looks genuinely shareable.

— Claude, reviewing on Marie's behalf

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
mariechatfield: Disagree — keeping the `useMemo`. `eligibleContractors` is a pure derivation of `firstPage`/`restPageResults`, and TanStack Query already gives structural sharing on both (`replaceEqualDeep` in `QueriesObserver`), so they're referentially stable when unchanged — the 'array identity flips even when nothing changed' risk this comment raises doesn't apply here. More fundamentally: this hook doesn't own *when* those values change — they're subscriptions into the query cache, not something we set ourselves — so there's no imperative moment to hang a manual `setState` off of, unlike `rawSelectedIds`, which the hook does own via `onSelect`/`onSelectAll`. Manually syncing state would mean re-implementing `useMemo`'s own change-detection by hand, with more surface area for the exact class of bug we're trying to avoid. It's also load-bearing today, not hypothetical: `useClientPagination` and `onSelectAll` both consume this value, so an unstable reference would cascade into real extra recomputation.

— Claude, reviewing on Marie's behalf

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
mariechatfield: On reflection, reverting this — the fix cost more than it saved. `matchesContractorSearch` re-lowercases the query on every call, but the query is a short string and this list is documented (a few lines up) as typically <100 contractors, so the actual saved work is unmeasurable. The clean fix (caching the normalized query per search pass) needed a `useRef` plus a cross-file invariant — '`useClientPagination` happens to call us with the same query for every item within one filter pass' — that isn't enforced by any type or contract, just documented behavior in a different file. That's more moving parts and a subtler failure mode than the redundant `.toLowerCase()` calls it was meant to avoid. Leaving as originally written.

— Claude, reviewing on Marie's behalf

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
```diff
@@ -0,0 +1,192 @@
...
+/** @internal */
+export function isEligibleContractor(contractor: Contractor): boolean {
+  return (
+    contractor.isActive &&
+    contractor.onboardingStatus === ContractorOnboardingStatus.ONBOARDING_COMPLETED
+  )
+}
+
+function matchesContractorSearch(contractor: Contractor, query: string): boolean {
+  const name =
```
dmortal: I believe we have some helper for contractor name in place already

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
dmortal: I would argue that real world application in this case does not need useMemo  at all - we are talking about sub ~10 items arrays in this case with users not using this component extensively. shooting a fly with a bazooka type of deal

dmortal (approved):

src/components/Contractor/shared/SelectContractors/useSelectContractors.ts
mariechatfield: oh thanks for the callout! updating to use the `getContractorDisplayName` now (and moving it out of `CreatePayment/helpers` into a shared utility file for all contractor domain)

PR #2462, opened 2026-07-23
feat: add "Manage tax rates" history/scheduling view to StateTaxes

dmortal:
## Summary

Adds a "Manage tax rates" view for Company.StateTaxes that shows effective-dated rate history and lets partners schedule a new future-dated rate via the API's `scheduling` query param. Also fixes two real bugs surfaced while testing against live data.

## Changes

- New `TaxRateManagement` component (+ `AddRateAction`, `AddTaxRateDialog`, `TaxRateHistorySection`) showing per-key rate history (Current/Scheduled/Historical) with an "Add tax rate" flow that schedules a rate for an available future effective date, plus a success confirmation after saving.
- `StateTaxesList` row actions become a hamburger menu ("Edit tax settings" / "Manage tax rates"); new state machine transition + event wiring.
- Extracted shared dynamic-schema/field-rendering logic so `StateTaxesForm` and the new dialog share one implementation instead of duplicating it.
- Fixed: `TaxRequirement`/`TaxRequirementSet` keys aren't guaranteed unique within one API response — duplicate keys previously caused two visually separate fields/sections to silently share one react-hook-form field. `getUniqueRhfKey` now disambiguates.
- Fixed: `PercentageField` silently clamped out-of-range values via react-aria's hard min/max bounds with no error shown — now shows a real validation error and keeps the typed value visible.
- Fixed: `Dialog` overlays (Select/ComboBox/DatePicker/Menu) portaled outside the native `<dialog>`'s own subtree, which the browser makes inert while open, making nested popovers unclickable — `Dialog` now scopes the portal container to its own subtree.

https://github.com/user-attachments/assets/2dcc767b-21e2-4495-a530-5a5d546fded2

## Testing

- `npm run test -- --run` (full suite) — 297 files / 3563+ tests pass
- `npx tsc --noEmit -p .` — clean
- `npx eslint` on touched areas — clean
- Manually verified live against real API data via Playwright (demo company, CA state): scheduling a new rate, duplicate-key fields no longer share state, Select/DatePicker popovers work inside the dialog, out-of-range percentage shows a validation error, success banner appears after save.


src/components/Company/StateTaxes/shared/RequirementFields.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const { control } = useFormContext()
+  const watchedValues = useWatch({ control }) as StateTaxesFormValues
+
+  return (
+    <>
+      {requirements?.flatMap((requirement, index) => {
+        if (requirement.editable === false) return []
+        if (!isRequirementApplicable(requirement, setKey, watchedValues)) return []
+        return [
+          <Fragment key={`${requirement.key ?? 'requirement'}-${index}`}>
```
mariechatfield: Will the requirement ever _not_ have a key?

src/components/Company/StateTaxes/StateTaxesForm/Form.tsx
```diff
@@ -28,21 +31,8 @@ export function Form() {
...
-              requirement={{
-                ...requirement,
-                key: `${key}.${toRhfKey(requirement.key as string)}`,
-              }}
-              questionType={requirement.metadata?.type ?? 'Text'}
-              key={requirement.key}
-            />,
-          ]
-        })}
+        <RequirementFields requirements={requirements} setKey={setPath} />
```
mariechatfield: Just checking: `setKey` and `setPath` here all refer to `the key for this _set of requirements_` as opposed to `update the key` right? That's what I'm guessing from context, I just have to keep double checking since `setNoun` is so often a state setter function

mariechatfield (approved):
Code LGTM and it works as expected when I run it locally!

src/components/Company/StateTaxes/shared/RequirementFields.tsx
dmortal: Good catch — per the generated SDK types, `TaxRequirement.key` is `string | undefined`, so yes, it's technically possible. In every real API response we've seen (GA/WA/CA fixtures) it's always present, but I tightened this up anyway: `RequirementFields` now skips any requirement with no `key` (matching `buildRequirementSchema`'s existing guard) instead of rendering a field under a mangled, effectively-anonymous form path. Pushed in d247a7d.

src/components/Company/StateTaxes/StateTaxesForm/Form.tsx
dmortal: Confirmed — yes, it's "the key for this set of requirements", not a setter. Fair point that `setX` reads like a state setter at a glance; renamed to `requirementSetKey`/`requirementSetPath` everywhere (this file, RequirementFields, applicableIf, StateTaxesForm) in d247a7d.

dmortal: renamed for clarity

PR #2463, opened 2026-07-23
refactor(SDK-1127): extract CreatePaymentFlow and ViewPaymentFlow

mariechatfield:
## Summary

`ContractorManagement.PaymentFlow` remains exactly the same; `ContractorManagement.CreatePaymentFlow` and `ContractorManagement.ViewPaymentFlow` are new `@alpha` flows that represent the spokes / internal flows.

https://github.com/user-attachments/assets/f9852c7c-7713-4f6e-ad69-53f385ce8ee7

## Changes

- Test-first redo of the SDK-1127 hub+spoke split. Phase 1 (`fb9173d6`, `2f3a2910`) adds a flow-integration test layer against the current flat `PaymentFlow` — the only layer that renders the real component tree and can catch a wiring bug between hub and spoke — plus fixes two weak/no-op spots in the e2e harness that a prior attempt at this refactor needed but never landed.
- Phase 2 (`66f953ea`) re-extracts `CreatePaymentFlow`/`ViewPaymentFlow` as spokes, mirroring the `PayrollFlow`/`PayrollExecutionFlow` pattern. Fixes two real regressions a previous attempt at this same split introduced: each spoke now builds its machine via a lazy `useState` initializer (not `useMemo`, which doesn't guarantee identity survives a re-render triggered by a bubbled `onEvent` call) and passes `prefixBreadcrumbs` straight through without stripping `onNavigate` (which had broken "click breadcrumb to return to the hub").
- `CreatePaymentFlow`/`ViewPaymentFlow` are marked `@alpha`, not `@public` — they exist to prep a 4th spoke slot for the upcoming `HistoricalPaymentFlow`, not because a partner asked for standalone entry points. `PaymentFlow`'s public docs still describe the underlying blocks (`CreatePayment`, `PaymentSummary`, `PaymentHistory`, `PaymentStatement`) as before.
- Renamed the view spoke from `ViewHistoryFlow` to `ViewPaymentFlow` (`9cee5ead`) so it follows the same `<Verb>Payment` pattern as its sibling `CreatePaymentFlow`.
- Derived files (`.reports/`, `docs/reference/**`, `docs/guides/endpoint-*`, `sdk-app/generated-registry-data.ts`) intentionally excluded — CI regenerates and commits those separately.

## Test plan
- [x] `npm run test -- --run` — 3559 passed, 1 expected fail
- [x] `npx eslint` clean (only pre-existing unrelated warnings)
- [x] `npm run build` clean
- [x] `npx knip` clean on new files
- [x] Phase 1's `PaymentFlow.test.tsx`, written against the old flat implementation, passes unmodified against the new hub+spoke code


src/components/Contractor/Payments/PaymentFlow/PaymentFlow.test.tsx
mariechatfield, line 1: This is the important new test. My first attempt at refactoring accidentally broke some navigation. The _state machine unit tests_ passed because the logic for navigating on events remained; what was missing was _firing the expects we expected and when_.

These tests let us walk through more of the variations of navigation in a way that's cheaper/faster than a full e2e test but more thorough than purely a state machine test.

src/components/Contractor/Payments/PaymentFlow/PaymentFlow.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+        }}
+      >
+        force parent re-render
+      </button>
+      <PaymentFlow companyId={COMPANY_ID} onEvent={onEvent} />
+    </div>
+  )
+}
+
+describe('PaymentFlow', () => {
```
mariechatfield, line 129: Claude: I think we're missing the create payment happy-path flow, I know it's in e2e but we should include it here too for ease of local testing

mariechatfield: Claude: we also need to add some tests for the information request flow, while we're here

src/components/Contractor/Payments/ViewHistoryFlow/GUIDE.md
```diff
@@ -0,0 +1,16 @@
+<!-- Partner-facing guide content, published to the SDK docs site. -->
+
+# ViewHistoryFlow
+
+## Step flow <!-- slot: appendix -->
+
+`ViewHistoryFlow` centers on `PaymentHistory` as its hub: it shows a payment group's details and can either drill into an individual contractor's statement (`contractor/payments/view/details` → `PaymentStatement`) or cancel the group outright (`contractor/payments/cancel`), which exits the flow.
```
mariechatfield: claude: is this really a hub-and-spoke if it has a terminal step?

src/components/InformationRequests/InformationRequestsComponents.tsx
```diff
@@ -31,3 +31,4 @@ export function InformationRequestFormContextual() {
     />
   )
 }
+InformationRequestFormContextual.Footer = Footer
```
mariechatfield, line 34: Latent bug that we found while adding information requests to the flow test 🙃

dmortal (approved):

PR #2469, opened 2026-07-24
feat(SDK-1119): add contractor profile management

dmortal:
## Summary

Adds a Contractor Profile management surface (mirroring the existing Employee Profile management pattern) for viewing and editing a contractor's basic details after onboarding: `Profile`, `ProfileCard`, and `ProfileEditForm`, orchestrated via a `robot3` state machine.

## Changes

- `Profile`: orchestrator driving the read-view card and edit form via an internal state machine
- `ProfileCard`: standalone read-only card fetching its own data via a new `useContractorProfileSummary` hook, with an Edit action
- `ProfileEditForm`: standalone edit form for name/business name, start date, tax ID (SSN or EIN depending on contractor type), and email; an SSN/EIN already on file renders as a locked, masked value with a "Change" action
- Fix: `ProfileCard` and `ProfileEditForm` now show the real (partially-masked) EIN value returned by the API instead of a hardcoded `XX-XXXXXXX` placeholder; SSN keeps its hardcoded mask since the `Contractor` model has no equivalent real SSN value field
- New i18n namespace `Contractor.Management.Profile`, new component events (`contractor/management/profile/editRequested|updated|editCancelled`), docs regenerated

https://github.com/user-attachments/assets/ceba30a6-c083-41d7-b918-d7464f257bac

## Related

- Jira ticket: [SDK-1119](https://gustohq.atlassian.net/browse/SDK-1119)

## Testing

- `npm run test -- --run src/components/Contractor` — 27 test files, 305 tests passing
- `npx tsc --noEmit -p .` and `npx eslint src/components/Contractor/Profile` — clean

[SDK-1119]: https://gustohq.atlassian.net/browse/SDK-1119?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/Contractor/Profile/management/Profile.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * | ----- | ----------- | ---- |
+ * | `contractor/management/profile/editRequested` | Fired when the user clicks Edit on the read-view card | `{ contractorId: string }` |
+ * | `contractor/management/profile/updated` | Fired after the profile is successfully saved | {@link APIModels.Contractor} |
+ * | `contractor/management/profile/editCancelled` | Fired when the user cancels editing | — |
+ *
+ * @param props - See {@link ProfileProps}.
+ * @returns The contractor profile management surface.
+ * @public
+ */
+export function Profile({ dictionary, FallbackComponent, ...props }: ProfileProps) {
```
mariechatfield, line 56: Should this be released as `ProfileFlow`? or is it not really intended to be used as a standalone flow by end-consumers?

src/components/Contractor/Profile/shared/useContractorProfileSummary/useContractorProfileSummary.tsx
```diff
@@ -0,0 +1,79 @@
...
+ * Read-only data hook for the basic-details contractor Profile card.
+ *
+ * @remarks
+ * Fetches a single contractor record scoped to the fields the basic-details
+ * surface displays. Pair with {@link useContractorDetailsForm} to render an
+ * edit form against the same record.
+ *
+ * @param input - See {@link UseContractorProfileSummaryParams}.
+ * @returns A {@link HookLoadingResult} while loading, or the ready result with the loaded contractor once available.
+ * @public
```
mariechatfield: This looks like a great hook and it's marked `@public` but I don't think it's exported from the top-level `src/index` so it won't actually be available for use. Can we either export it, or downgrade it to `internal` or `alpha`?

src/components/Contractor/Profile/management/Profile.tsx
mariechatfield: ahh never mind, when I try using it locally I get it -- it's really _not_ a flow, it's just a block with some internal state management to show view vs edit

mariechatfield (approved):

krisxcrash (approved):

src/components/Contractor/Profile/shared/useContractorProfileSummary/useContractorProfileSummary.tsx
dmortal: Downgraded to `@internal` in 8e8c4624 — it's only ever used internally by `ProfileCard`, and there's no product need yet to expose it standalone (matching the existing `@internal` `useStateTaxesSummary`, which has the same read-only-summary-hook shape). Also removed it from the `Profile/shared` barrel so it's not sitting next to the genuinely public `useContractorDetailsForm` export.

PR #2471, opened 2026-07-24
fix(Employee): cache company locations queries to avoid refetch on every mount

krisxcrash:
## Summary
- `useLocationsGet` (in `useWorkAddressForm`) and `useLocationsGetMinimumWages` (in `useCompensationForm`) had no `staleTime` override, unlike sibling reference-data hooks in the same areas (`useHomeAddressSummary`, `useCompensationManagement`, etc.) which set `staleTime: Infinity`.
- Without it, remounting the WorkAddress or Compensation step re-fires the `/v1/companies/{id}/locations` (and minimum-wages) request on every mount/refocus instead of reusing cached data.
- Investigated as part of a root-cause look into elevated `/v1/companies/{uuid}/locations` traffic observed in demo for `gws-flows`-partner companies. SDK-side query gating is otherwise correct (no unconditional/top-of-tree calls found); this fix removes a refetch multiplier for any harness that mounts/remounts these steps repeatedly across many companies.

## Test plan
- [x] `npm run test -- --run src/components/Employee/Profile/shared/useWorkAddressForm src/components/Employee/Compensation/shared/useCompensationForm` — 50 tests pass
- [ ] Confirm in Storybook/dev app that remounting the WorkAddress or Compensation step no longer re-fires the locations request


src/components/Employee/Compensation/shared/useCompensationForm/useCompensationForm.tsx
```diff
@@ -464,7 +464,7 @@ export function useCompensationForm({
 
   const minWagesQuery = useLocationsGetMinimumWages(
     { locationUuid: locationUuid ?? '' },
-    { enabled: !!locationUuid },
+    { enabled: !!locationUuid, staleTime: Infinity },
```
mariechatfield, line 467: Do we need to always use `staleTime: Infinity` for these _kinds_ of hooks? How do we prevent this from happening again in the future?

mariechatfield (approved):

PR #2472, opened 2026-07-24
feat(contractor): add "Record a historical payment" CTA to payments list

aaronlee777:
## Summary
- Adds a "Record a historical payment" Box CTA below the contractor payments DataView on `PaymentsListPresentation`, mirroring the off-cycle CTA pattern on `PayrollList`.
- Uses a Flex column of Text components (title + supporting description) as the Box header, with a secondary Button footer.
- Button `onClick` is a no-op — UI only for now, action to be wired up in a follow-up.

## DEMO

Screenshot 2026-07-24 at 12 34 34 PM

## Test plan
- [ ] Verify the CTA renders below the DataView in Storybook / sdk-app
- [ ] Verify the CTA renders in both populated and empty payments states
- [ ] Verify the title, description, and button label read correctly
- [ ] Verify layout matches the off-cycle CTA on the payroll list at the same 640px max width

### DIFF
--- a/docs/reference/Translations/index.md
+++ b/docs/reference/Translations/index.md
@@ -1689,6 +1689,10 @@ Translation keys for the `Contractor.Payments.PaymentsList` i18n namespace.
 | `dateRanges.last12Months` | `"Last 12 months"` |
 | `dateRanges.last3Months` | `"Last 3 months"` |
 | `dateRanges.last6Months` | `"Last 6 months"` |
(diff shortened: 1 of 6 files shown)


serikjensen (changes requested):
Request to move this one to the design directory so we don't end up with a broken entrypoint for historical payments

serikjensen:
Closed since this is a minor lift to implement. We will implement once historical payments are built so that the link properly navigates

PR #2500, opened 2026-07-31
feat(SDK-1126): historical contractor payment - select contractors

mariechatfield:
## Summary
- Adds the shared `SelectContractors` block (multi-select contractor table with pagination, search, wage column, type subtitle) for reuse across contractor payment flows.
- Adds `getHistoricalPaymentCheckDateBounds`, implementing the Jan 1–8 grace-window cutover so the check date is bounded to the current allowed tax year(s).
- Adds `HistoricalPaymentContractors`: the first step of the historical contractor payment flow — pick a check date and the contractors being paid, then emit `contractor/historicalPayments/contractorsSelected`.

## Demo

Screenshot 2026-07-31 at 10 01 06 AM

Continue disabled if no contractors selected
Screenshot 2026-07-31 at 10 01 11 AM

Previous tax year is invalid
Screenshot 2026-07-31 at 10 00 44 AM

Future date is invalid
Screenshot 2026-07-31 at 10 00 56 AM

## Test plan
- [x] `npm run test -- --run` passes for new/changed files
- [x] Manual check in Storybook / sdk-app


src/i18n/en/Contractor.Payments.HistoricalPaymentContractors.json
```diff
@@ -0,0 +1,8 @@
+{
+  "heading": "Record a historical payment",
+  "subtitle": "Log a contractor payment that already happened outside Gusto. Pick a paid date and the contractors you paid.",
+  "dateLabel": "Payment date",
+  "dateInFutureError": "You cannot issue historical payments for the future. Please choose a date in the past.",
+  "dateTooEarlyError": "You cannot create a payment in {{year}}. Please select a {{allowedYear}} date.",
```
mariechatfield, line 6: I don't think these are the right error messages, will need some product/design guidance on copy here. I _think_ the goal would be to explain more like "you can't do a previous tax year bc that would mess up your reporting" and "if you're doing a future date, you need to switch to the _make a payment_ flow, this is just the _record a payment that already happened_ flow"

src/components/Contractor/shared/SelectContractors/SelectContractorsPresentation.stories.tsx
mariechatfield, line 1: Stories added for the select contractors internal block only

Image
Image

src/components/Contractor/shared/SelectContractors/SelectContractors.test.tsx
```diff
@@ -0,0 +1,182 @@
...
+          <SelectContractors
+            companyId="company-123"
+            onSelectionChange={onSelectionChange}
+            {...overrides}
+          />
+        </ComponentsProvider>
+      </ThemeProvider>
+    </LocaleProvider>,
+  )
+}
```
serikjensen, line 71: I believe we have a GustoTestProvider which should already be setting these rather than needing to manually declare all the providers? (same feedback for the others in this file)

serikjensen (approved):
Looks good! left a suggestion to update to use the existing test provider

PR #2501, opened 2026-07-31
fix(SDK-1018): stop PayrollOverview throwing on stale calculatedAt after calculate

serikjensen:
## Summary

Looks to potentially fix an intermittent `Error: Payroll is not calculated` that trips the error boundary when landing on `PayrollOverview`. This has been a challenging bug to reproduce and only happens occasionally. The best guess around it is that it has something to do with the query cache, since calculate payroll does not invalidate. This adds that invalidation to ensure we load the payroll again when we get to the submit screen. It also does some additional guards to ensure the payroll is done fetching prior to checking on the calculated at.

This is a best guess fix since we could not repro consistently. I ran this through various manual tests on payrolls locally, including regular, off cycle, and termination. We can watch after this to see if we can have other instances of payroll throwing the error after this.

## Related

- Jira: [SDK-1018](https://gustohq.atlassian.net/browse/SDK-1018)

## Verification

https://github.com/user-attachments/assets/ae62fd31-8e2e-450e-aa3c-48e5aa5cc358

---

[SDK-1018]: https://gustohq.atlassian.net/browse/SDK-1018?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


serikjensen:
@claude running a test to verify if this works. Could you add a little additional description to the end of the PR description that's already there as a test?

mariechatfield (approved):

serikjensen:
@claude running a test to verify if this works. Could you add a little additional description to the end of the PR description that's already there as a test?

serikjensen:
@claude could you remove the comments that were added to the new tests in the test file in this PR?

PR #2505, opened 2026-07-31
feat(SDK-484): enforce accessibility testing in CI

mariechatfield:
## Summary

Closes out SDK-484's implementation checklist (add axe tests to CI, zero-violation threshold, PR gate, contributor docs) by making accessibility testing automatic instead of opt-in, in both test layers.

Originally stacked on #2521 (the a11y fixes), since turning the gate on required those violations fixed first. #2521 is now merged into main, so this diff is just the testing infrastructure:

- **vitest**: every test that renders a component now gets an automatic axe check (`src/test/setup.ts`), instead of relying on individual tests to opt in.
- **Playwright e2e**: every spec using `localTestFixture` now gets automatic axe checks against the flow's own rendered output, with zero per-test boilerplate.

## What gets checked, and when

**vitest** — a global `afterEach` hook (`src/test/setup.ts`) runs axe against the rendered DOM of *every* test that renders a component, right before RTL's cleanup (`RTL_SKIP_AUTO_CLEANUP` disables RTL's own auto-cleanup so the DOM is still mounted when the check runs). It uses a relaxed rule set (`INTEGRATION_TEST_DISABLED_RULES` in `src/test/accessibility.ts`, pre-existing) that ignores page-level rules like `region`/`heading-order`, since an isolated component rendered outside a real page has no surrounding landmarks to check against. Components with dedicated a11y test cases can still opt in to the stricter `expectNoAxeViolations(container)` check.

**Playwright e2e** — every spec using `localTestFixture`'s `page` fixture gets checked at three points, all funneling through a single `A11yViolationTracker` (`e2e/utils/a11yViolationTracker.ts`) reset per test:

1. **Initial page load** — the fixture's `page.goto` override checks once the app has mounted and settled.
2. **Every `waitForLoadingComplete()` call** — this runs at nearly every point a test expects the screen to have changed, so it's the main source of coverage on intermediate screens (dialogs, drawers, validation states) a test passes through but never ends on. Each finding is labeled with the calling function/file:line (`e2e/utils/callerInfo.ts`) so a report says which step in the flow produced it.
3. **End of test** — a final `expectNoAxeViolations(page)` mirrors the vitest gate's final check.

Checkpoints 1 and 2 dedupe by rule + target (normalized to ignore React's per-render auto ids), so one broken shared component surfaces once with a `recurredAt` list instead of once per screen it appears on. Any that survive to a passing test are attached as `a11y-intermediate-violations.txt` and fail it — same zero-tolerance bar as the final check. One manual mid-test call is retained in `04-cancel-mid-onboarding.spec.ts`, for a screen that isn't reached via `waitForLoadingComplete` at all.

Both layers scope to the component/flow's own container (`main` in e2e, since `e2e/main.tsx` wraps every flow in one) rather than the whole document — the SDK doesn't own the host page's landmarks, heading hierarchy, `lang`, or title, so checking against the harness's or jsdom's page chrome would just test the harness, not the SDK.

## Changes

**vitest gate**
- `src/test/setup.ts` / `vite.config.ts`: global `afterEach` axe check on every render, with RTL auto-cleanup disabled so it runs before unmount.

**Playwright e2e gate**
- `e2e/utils/a11y.ts` (new): `expectNoAxeViolations(page)`, wrapping `@axe-core/playwright`, scoped to `<main>`.
- `e2e/utils/a11yViolationTracker.ts` / `callerInfo.ts` (new, with tests): the intermediate-checkpoint collector and its caller-labeling, described above.
- `e2e/utils/localTestFixture.ts`: wires the `page.goto` and end-of-test checkpoints, plus the tracker reset.
- `e2e/utils/helpers.ts`: wires the `waitForLoadingComplete()` checkpoint. Also fixes that helper's selector, which looked for `role="region"` and never matched the SDK's `role="status"` Loading indicator — so it never actually waited for anything before this fix.
- One manual mid-test call retained in `04-cancel-mid-onboarding.spec.ts` (see above).

**Docs**
- `CONTRIBUTING.md`, `e2e/CLAUDE.md`: document both gates for contributors, including the three e2e checkpoints and how they dedupe.

## Known follow-ups (not blocking this PR)

- **`color-contrast` is temporarily disabled by default** in the e2e gate (`TEMPORARILY_DISABLED_RULES` in `e2e/utils/a11y.ts`). A real contrast bug on a shared `Text` component (`xs`/`weight-semibold` variant) fails nearly every spec that renders it; re-enable once that's fixed.

## Related

[SDK-484](https://gustohq.atlassian.net/browse/SDK-484)

## Testing

- `npm run test -- --run` — full suite green post-rebase (334 files / 3740 tests; one unrelated flake in `HistoricalPaymentAmounts.test.tsx` that passes in isolation).
- `npx tsc --noEmit` clean.
- e2e: CI running post-rebase, will confirm green before requesting review.

[SDK-484]: https://gustohq.atlassian.net/browse/SDK-484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mariechatfield:
@SocketSecurity ignore npm/@axe-core/playwright@4.12.1

This is a `devDependency` / test-only library which is allowed per our OSM

serikjensen (approved):

PR #2542, opened 2026-08-07
fix: require frequency option for twice-per-month pay schedules

serikjensen:
## Summary

For a "Twice per month" pay schedule the user must choose how the two pay days are defined — the **15th/last** preset or **custom** days — but the "Frequency Options" radio (`customTwicePerMonth`) was marked optional with no default. This let an incomplete twice-per-month schedule pass client-side validation (the server would then reject it), and a prior refactor's fallthrough could also surface confusing **"Day 1 is required" / "Day 2 is required"** errors pointing at day inputs that aren't even visible.

This makes the radio required when frequency is "Twice per month", so validation lands on the **visible** field with an actionable message instead of on hidden day fields.

## Changes

- **`payScheduleSchema`**: `customTwicePerMonth` is now required when frequency is "Twice per month" (predicate rule instead of `'never'`). `day1`/`day2` requiredness is unchanged — still gated on the `'custom'` choice.
- **`usePayScheduleForm`**: `deriveCustomTwicePerMonth` only pre-selects `'custom'` when real day values exist (e.g. editing an existing custom schedule); with no day basis it leaves the radio unselected so the required validation prompts an explicit choice rather than silently forcing empty day fields into a required state.
- **`fields` / `PayScheduleForm`**: wired a `REQUIRED` validation message ("Please select the pay days for the month"). The "(optional)" suffix clears automatically since the field forwards the metadata-driven `isRequired`.
- Regenerated i18n types and the auto-generated reference docs.

## Behavior after the fix

- **Twice per month + no option selected** → clear required error on the radio; no phantom day1/day2 errors; submission blocked.
- **"15th and Last day"** → auto-fills 15/31, day fields stay hidden, valid.
- **"Custom"** → reveals editable Day 1 / Day 2 inputs (required when empty).
- Other frequencies (weekly/monthly) unaffected.

## Note

The screenshot in the ticket reflected older behavior. An earlier form-hooks refactor had already stopped the phantom day errors, but left the opposite gap: a twice-per-month schedule could be submitted with no pay days at all. This PR closes that gap.

## Testing

Verified live in the SDK dev app (`/companyonboarding/PaySchedule`): submitting twice-per-month with no option shows the radio required error and no day1/day2 errors; the "(optional)" suffix is gone; "15th and Last day" and "Custom" both resolve correctly.

https://github.com/user-attachments/assets/1db49006-eefe-4bd3-8b50-f3ab0e0360cb


src/components/Company/PaySchedule/shared/usePayScheduleForm/payScheduleSchema.ts
```diff
@@ -70,6 +70,10 @@ export type PayScheduleFormOutputs = PayScheduleFormData
 
 // ── Required fields config ─────────────────────────────────────────────
 
+function needsCustomTwicePerMonth(data: PayScheduleFormData): boolean {
+  return data.frequency === 'Twice per month'
```
dmortal, line 74: nit: this should probably be a const of some sort

dmortal (approved):

krisxcrash (approved):

PR #2552, opened 2026-08-11
feat(SDK-1129): add HistoricalPaymentsFlow

mariechatfield:
## Summary
- Adds `HistoricalPaymentFlow`, chaining `CreateHistoricalPayment` -> `HistoricalPaymentSummary` (mirrors `CreatePaymentFlow`'s shape: two-state machine, lazy-init so in-flight state survives a host re-render, non-final terminal state so the summary screen stays interactive if not unmounted).
- Wires it in as a fourth spoke off `PaymentFlow`: new `contractor/historicalPayments/create` event, a "Record a historical payment" CTA on the payments list, and a `historicalPayment` hub state that returns to `landing` on exit.
- Exports `HistoricalPaymentFlow` publicly (`@alpha`) alongside its breadcrumb labels, and regenerates the derived API report / docs / i18n types.

https://github.com/user-attachments/assets/212baea8-5dae-4a95-afd6-8e16d057f956

## Test plan
- [x] `npm run tsc`
- [x] `npm run test -- --run src/components/Contractor/Payments` (81 tests passing)
- [x] New `HistoricalPaymentFlow.test.tsx` covering the create -> summary chain and exit
- [x] Extended `PaymentFlow.test.tsx` covering CTA entry, the full create -> summary -> exit chain through the hub, and re-render resilience mid-flow
- [ ] Manual click-through in Storybook / sdk-app

### DIFF
--- a/src/components/Contractor/Payments/HistoricalPaymentFlow/HistoricalPaymentFlow.tsx
+++ b/src/components/Contractor/Payments/HistoricalPaymentFlow/HistoricalPaymentFlow.tsx
@@ -1,0 +1,126 @@
+import { createMachine } from 'robot3'
+import { useState } from 'react'
+import {
+  historicalPaymentBreadcrumbsNodes,
+  historicalPaymentMachine,
+} from './historicalPaymentMachine'
+import {
+  CreateHistoricalPaymentContextual,
+  type HistoricalPaymentFlowContextInterface,
+  type HistoricalPaymentFlowProps,
+} from './HistoricalPaymentFlowComponents'
+import { Flow } from '@/components/Flow/Flow'
+import type { FlowBreadcrumb } from '@/components/Common/FlowBreadcrumbs/FlowBreadcrumbsTypes'
+import { buildBreadcrumbs, updateBreadcrumbs } from '@/helpers/breadcrumbHelpers'
+
+const EMPTY_BREADCRUMBS: FlowBreadcrumb[] = []
+
+/**
+ * Props for the flow-internal {@link HistoricalPaymentInternalFlow}, which layers a parent flow's
+ * prefix breadcrumbs on top of the public {@link HistoricalPaymentFlowProps}.
+ *
+ * @internal
+ */
+export interface HistoricalPaymentInternalFlowProps extends HistoricalPaymentFlowProps {
+  /**
+   * Breadcrumbs prepended to the flow's own breadcrumb trail. Set by a parent flow (e.g.
+   * `PaymentFlow`) so the breadcrumb history remains coherent across the handoff.
+   */
+  prefixBreadcrumbs?: FlowBreadcrumb[]
+}
+
+/**
+ * Guided flow to record a historical contractor payment and review the resulting summary.
+ *
+ * @remarks
+ * This is the inner flow that powers the historical-payment spoke of `ContractorManagement.PaymentFlow`.
+ * Render it directly when you have built your own payments landing page and want to hand the user
+ * off to the standard historical-payment experience without re-implementing it. A historical payment
+ * already happened outside Gusto and does not move money, so unlike `CreatePaymentFlow` there is no
+ * Fast ACH blocker or wire-transfer step.
+ *
+ * @events
+ * | Event | Description | Data |
+ * | ----- | ----------- | ---- |
+ * | `contractor/historicalPayments/edit` | The edit modal was opened for a contractor | — |
+ * | `contractor/historicalPayments/update` | A contractor's payment values were updated locally | The updated form values (hours, wage, bonus, reimbursement, payment method, etc.) |
+ * | `contractor/historicalPayments/preview` | The preview API call succeeded | The contractor payment group preview response |
+ * | `contractor/historicalPayments/backToEdit` | The user returned from preview to continue editing | — |
+ * | `contractor/historicalPayments/created` | The payment group was successfully created | The created `ContractorPaymentGroup` |
+ * | `contractor/historicalPayments/exit` | User is done reviewing the summary | — |
+ * | `breadcrumb/navigate` | Fired when the user clicks a breadcrumb to navigate back | `{ key: string, onNavigate: (ctx) => ctx }` |
+ *
+ * @components
+ * - {@link CreateHistoricalPayment}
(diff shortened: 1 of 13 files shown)


src/components/Contractor/Payments/PaymentFlow/PaymentFlow.tsx
mariechatfield, line 1: This is the one thing that's a bit weird. The components for `HistoricalPaymentsFlow` are marked `alpha` but this is live-updating the `public` `PaymentFlow` which feels... a bit weird. Should there be a way to mark an alpha feature as opt-in?

serikjensen: For a similar situation with i9 in employee onboarding, I ended up just going with a prop that was available at the top level of the flow component. LIke `withEmployeeI9` defaulted to false on the initial rollout. If you're concerned about releasing pre test fest, that might be a good way to keep it hidden while the test fest and fixes happen. Then when we feel ready we can remove the flag

You could also go more descriptive with the prop like `withUnstableInProgressHistoricalPaymentsUseWithCaution` or similar

src/components/Contractor/Payments/HistoricalPaymentFlow/historicalPaymentMachine.ts
```diff
@@ -0,0 +1,73 @@
...
+  },
+  historicalPaymentSummary: {
+    parent: null,
+    item: {
+      id: 'historicalPaymentSummary',
+      label: 'breadcrumbLabel',
+      namespace: 'Contractor.Payments.HistoricalPaymentSummary',
+    },
+  },
+}
```
serikjensen: It's worth touching base with @aaronlee777 on the breadcrumb use here. I think breadcrumbs are generally something we're getting away from. He may also have an alternative for nav recommendation for the flow

src/components/Contractor/Payments/HistoricalPaymentFlow/HistoricalPaymentFlow.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    return createMachine(
+      'createHistoricalPayment',
+      historicalPaymentMachine,
+      (initialContext: HistoricalPaymentFlowContextInterface) => ({
+        ...initialContext,
+        ...initialBreadcrumbContext,
+        component: CreateHistoricalPaymentContextual,
+        companyId,
+      }),
```
serikjensen, line 116: You might consider placing the machine creation in a useMemo here instead. Ex. if we got a new company id, we'd want to tear down the machine and create a new one

serikjensen (commented):
Looking good! couple of things

* Would recommend adding the boolean flag prop and defaulting it to false while we're rolling this out pre test fest (as described in slack)
* We should align with Aaron on breadcrumb use, i think we're getting away from that so that could simplify some things in the implementation
* Recommendation to update the state machine to properly re-create on company id change

src/components/Contractor/Payments/PaymentFlow/PaymentFlow.tsx
mariechatfield: from slack discussion: removing the `PaymentFlow` changes and only adding new alpha surface area in this PR

src/components/Contractor/Payments/HistoricalPaymentFlow/HistoricalPaymentFlow.tsx
mariechatfield: I think if the company id ever changes, we want to completely tear down the entire component and re-create it from scratch which fits more with the pattern of passing `key={companyId}` to the component when we render it. I can do that from the parent flow when I update it

src/components/Contractor/Payments/HistoricalPaymentFlow/historicalPaymentMachine.ts
mariechatfield: Aaron is out today -- I can go ahead and remove these entirely for now, or are you okay with merging as is?

src/components/Contractor/Payments/HistoricalPaymentFlow/HistoricalPaymentFlow.tsx
serikjensen: @mariechatfield i'd be hesitant to not account for it in the component itself since partners can struggle with configurations like that (see my recent history 😭) so my rec would be to account for it in the component itself if we are able

src/components/Contractor/Payments/HistoricalPaymentFlow/historicalPaymentMachine.ts
serikjensen: Yup! insofar as we're able to follow up with aaron

serikjensen (changes requested):
This looks good! request to make sure that we account for changing essential flow dependencies inside our component consistent with other SDK state machines rather than requiring the partner to key those and track them on their end

src/components/Contractor/Payments/HistoricalPaymentFlow/HistoricalPaymentFlow.tsx
mariechatfield: I went ahead and updated all of the new flows to use `key={companyId}` or `key={paymentId}` on the internal components around their state machines because for any of them, it's not just about re-creating the state machine. Any internal state anywhere in the components (like contractors, payments, selected dates, etc) needs to be entirely cleared and refetched, to avoid updating the state machine but leaving transient state in place.

This doesn't require a partner to change how they use the components, but adds a more thorough guard around changing the prop on the fly

mariechatfield:
@serikjensen this should be ready for re-review! Got rid of the breadcrumbs in favor of the back buttons that the design prototype _already used_ 🤦🏻 and updated each of the three components to use a `key={companyId}` or `key={paymentId}` so that not just the state machine but the entire internal state is completely reset if the main identity ever changes

Screenshot 2026-08-18 at 9 26 55 AM
Screenshot 2026-08-18 at 9 27 03 AM
Screenshot 2026-08-18 at 9 27 09 AM

serikjensen (approved):
Looks good!

PR #2572, opened 2026-08-14
feat(SDK-1161): add print payroll checks alert and modal

krisxcrash:
## Summary

Employees paid by check need a way to generate printable checks from the payroll overview screen. This redesigns the existing check-payment alert (`status="warning"` → `status="info"`, same copy) and adds a "View and print checks" action, shown once the payroll is processed, that opens a new modal for choosing check stock and generating the check PDF via the existing printable-payroll-checks API.

## Changes

- New `PrintChecksModal` (hook + presentation + container) handling the check-stock choice, the conditional starting-check-number field, and the generate → poll → PDF flow via `usePayrollsGeneratePrintableChecksMutation` / `useGeneratedDocumentsGet` (already available at the current pinned API version, no bump needed)
- `PayrollOverviewPresentation`'s check-payment alert switched to `status="info"`, with the button placed in its own row below the description, matching the established alert+CTA pattern elsewhere in the codebase
- New `Payroll.PrintChecksModal` i18n namespace; three new `runPayroll/printChecks/*` events
- New MSW handlers for the two endpoints, used in a new integration test suite

## Related

- Jira ticket: [SDK-1161](https://gustohq.atlassian.net/browse/SDK-1161)

## Testing

- `npm run test -- --run` — full suite passes (336 files / 3765 tests)
- `npm run build` and `npm run lint:check` — clean
- Storybook: `Domain/Payroll/PrintChecksModal` (Default, BlankCheckStockSelected, Generating, Succeeded, Failed) and `Domain/Payroll/PayrollOverview` (WithCheckPaymentEmployeeUnprocessed, WithCheckPaymentEmployeeProcessed)
- Manually verified in the sdk-app dev build against a real check-payment employee, both pre- and post-submission

[SDK-1161]: https://gustohq.atlassian.net/browse/SDK-1161?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Screenshots
Screenshot 2026-08-14 at 2 32 40 PM
Screenshot 2026-08-14 at 2 39 58 PM
Screenshot 2026-08-14 at 2 41 33 PM


src/components/Payroll/PayrollOverview/PayrollOverviewPresentation.tsx
```diff
@@ -782,10 +784,19 @@ export const PayrollOverviewPresentation = ({
...
               >
-                {t('alerts.checkPaymentWarningDescription')}
+                <Flex flexDirection="column" gap={12}>
+                  <Text>{t('alerts.checkPaymentWarningDescription')}</Text>
+                  {isProcessed && (
+                    <div>
+                      <Button variant="secondary" onClick={onPrintChecksOpen}>
+                        {t('alerts.printChecksCta')}
+                      </Button>
+                    </div>
```
serikjensen, line 797: i think you can possibly set this via the `action` prop available on Alert? that way partners can manage this with their component adapter setup.

it'd look like

```
<Alert action={isProcessed && <Button ...>....</Button>} ...>...</Alert>
```

src/components/Payroll/PrintChecksModal/PrintChecksModal.stories.tsx
```diff
@@ -0,0 +1,75 @@
...
+  decorators: [
+    (Story: React.ComponentType) => (
+      <Suspense fallback={<div>Loading translations...</div>}>
+        <I18nLoader>
+          <GustoTestProvider>
+            <Story />
+          </GustoTestProvider>
+        </I18nLoader>
+      </Suspense>
+    ),
```
serikjensen, line 26: Nit: it's been a bit since i've added stories, isn't there some sort of global configuration for this so we don't have to duplicate these for each story?

src/components/Payroll/PrintChecksModal/PrintChecksModal.test.tsx
```diff
@@ -0,0 +1,269 @@
...
+    renderWithProviders(<PrintChecksModal {...defaultProps} />)
+
+    await user.click(await screen.findByRole('button', { name: 'View checks' }))
+
+    await waitFor(() => {
+      expect(getDocumentResolver).toHaveBeenCalled()
+    })
+    expect(generateResolver.mock.invocationCallOrder[0]!).toBeLessThan(
+      getDocumentResolver.mock.invocationCallOrder[0]!,
+    )
```
serikjensen, line 146: Nit: I would push claude on this one since it is likely doing something hacky if it has to assert like this

src/components/Payroll/PrintChecksModal/PrintChecksModalPresentation.tsx
```diff
@@ -0,0 +1,132 @@
...
+                  errorMessage={startingCheckNumberErrorMessage}
+                  isDisabled={isGenerating}
+                  min={0}
+                />
+              )}
+            </Flex>
+          </Form>
+        </FormProvider>
+      )}
+    </Modal>
```
serikjensen, line 130: A few things on this one that we need to think about

* We end up tracking the progress through the modal and transitioning screens by phase, i would recommend navigating that via state machine as the de facto way of managing screen transitions
* We end up needing to do manual tracking of retries and form reset because the modal opens and closes but the content never unmounts
* We also are unable to export the individual pieces of this print checks experience for usage in case a partner wants to utilize those components

I think the best reference implementation to claude to work with for this is the ConfirmWireDetails or RecoveryCases components inside payroll

Those ones will show precedent for
* Creating a modal that will just open close
* Placing a BaseComponent as the root component of the modal content
* Having api calls occur within the modal so that we can catch errors and display them inside the modal and use the existing BaseComponent error handling consistent with other components
* Managing the transitions between screens with a lightweight state machine
* Allow for exporting the individual component pieces in isolation which are then connected by the state machine

I would recommend just having those components manage their api calls internally rather than centralizing the functionality in a hook. If we do provide hooks at this point, we'd want them to align with the existing codebase hooks patterns which is likely more than you'd want to take on right now

So in the end you'd end up with
* PrintChecks <- Creates state machine, renders triggering banner with CTA and modal
* PrintChecksBanner <- Banner component renders the button and the banner launch
* PrintChecksForm
* PrintChecksFailure <- Failure state when the polling indicates a failure status
* PrintChecksSummary <- Success, link to other checks

Lmk if you'd like to pair on this!

src/components/Payroll/PrintChecksModal/usePrintChecksModal.ts
```diff
@@ -0,0 +1,178 @@
...
+
+const buildRequestBody = (data: PrintChecksFormValues): PrintablePayrollChecksBody => ({
+  printingFormat: data.printingFormat,
+  ...(data.printingFormat === PrintingFormat.Bottom && data.startingCheckNumber !== undefined
+    ? { startingCheckNumber: data.startingCheckNumber }
+    : {}),
+})
+
+/** @internal */
+export function usePrintChecksModal({
```
serikjensen, line 66: See my longer comment, my rec would be to get away from a hook for this implementation as it will end up increasing the scope here. We likely only want a hook if this is being used in multiple locations and this is an internal helper to centralize some duplicated logic. Or, alternatively, if we want to make the hook available to partners, in which case we'd want to make sure this aligns with the hook guidelines. I don't think that's necessarily worth worrying about this pass

src/components/Payroll/PrintChecksModal/usePrintChecksModal.ts
```diff
@@ -0,0 +1,178 @@
...
+      formMethods.reset({ printingFormat: PrintingFormat.Top, startingCheckNumber: 0 })
+      setPhase('form')
+      setRequestUuid(null)
+      setDocumentUrl(null)
+      setErrorMessage(null)
+      setIsPolling(false)
+      printWindowRef.current?.close()
+      printWindowRef.current = null
+    }
+  }, [isOpen, formMethods.reset])
```
serikjensen, line 107: This goes away with my recommendation to create dedicated standalone components that only get rendered when the modal mounts. Modal is rendering always in this case which means the component is always mounted for this which is why we need the manual reset.

serikjensen (commented):
Looking good! Left some feedback here to update this to be consistent with the other Payroll modal components!

src/components/Payroll/PayrollOverview/PayrollOverviewPresentation.tsx
krisxcrash: great catch!

src/components/Payroll/PrintChecksModal/PrintChecksModal.stories.tsx
krisxcrash: Let me look into that!

krisxcrash:
Splitting this into a stack of smaller, more reviewable PRs per feedback on this PR's size:

1. #2584 — terminal screens (PrintChecksFailure/Summary) + event constants
2. #2585 — PrintChecksBanner
3. #2586 — PrintChecksForm (generate/poll/download logic)
4. #2587 — state machine + orchestrator, public export
5. #2588 — wires PrintChecks into PayrollOverview

Each PR is stacked on the previous and was verified to build/lint/test independently. Closing this one in favor of that stack.

